PR #237, opened 2015-01-01
Static directory

Ian-Stewart-Binks:
@david-yz-liu This should make next week's installation process much easier. Let me know if this works on Windows.

### DIFF
--- a/hs/server.hs
+++ b/hs/server.hs
@@ -29,21 +29,20 @@ about = "about"
 static :: String
 static = "static"
 
-staticDir :: String
---staticDir = "C:\\Users\\David\\Documents\\courseography"
-staticDir = "/home/cynic/4/courseography"
-
 course :: String
 course = "course"
 
 main :: IO ()
-main = simpleHTTP nullConf $
-  msum [ dir grid $ gridResponse,
-         dir graph $ graphResponse,
-         dir about $ aboutResponse,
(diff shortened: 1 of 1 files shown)


hs/server.hs
```diff
@@ -29,21 +32,20 @@ about = "about"
...
-main = simpleHTTP nullConf $
-  msum [ dir grid $ gridResponse,
-         dir graph $ graphResponse,
-         dir about $ aboutResponse,
-         dir static $ serveDirectory EnableBrowsing [] staticDir,
-         dir course $ path (\s -> liftIO $ queryCourse s)
-       ]
+main = do
+    cwd <- getCurrentDirectory
+    let staticDir = encodeString $ parent $ fromText (T.pack cwd)
```
david-yz-liu: use decodeString

Ian-Stewart-Binks: I used encodeString because the result of `parent` is a `Filesystem.Path.CurrentOS.FilePath`, which is unrecognised by Happstack. It either accepts a `String` or a `Prelude.FilePath`. From what I tried, decodeString will not work.

david-yz-liu: Ah, I meant decodeString instead of `fromText` and `T.pack`; I think `cwd` is a String, no?

Ian-Stewart-Binks: You are correct! Changes have been made.

PR #246, opened 2015-01-05
dummy

david-yz-liu:


planner.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 
 <html>
-
+hahahah
```
david-yz-liu: This is not funny at all. :(

pkukulak: Hilarious! +1

Issue #233, opened 2014-12-29
Graph: New Graph Missing Prereq


Ian-Stewart-Binks:
The new graph is missing the prerequisite 324 -> 488.

david-yz-liu:
It is missing a few more than this. It appears that when multiple courses appeared in a hybrid, only the one course is appearing.

Ian-Stewart-Binks:
This is going to take a bit of work, I forgot that SVG text wrapping/newlines in text elements is a tricky task. It shouldn't be that bad, though.

PR #248, opened 2015-01-06
Issue 233: Missing prereqs

Ian-Stewart-Binks:
This pull request fixes issue #233, where some hybrid nodes would not show the entire prerequisite list for a course. This also adds in the ECE prefix for ECE courses.

The problem was that the parser was discarding any text after the first `,` or `/` it saw, and it would also only account for one text element. In the input graph, text elements spanning multiple lines are actually multiple text elements. This is fixed within the `Rect` object by converting `rect.text` to a list. The alternative was to store an extra variable (which I had initially called `extra_text` in some of the previous commits) that stores the second text element text. The original approach was bad for extensibility, it would be cumbersome to account for more than 2 text elements.

The text elements are also ordered based on their y position in the original graph, so a text element that appears above another in the original graph will appear above the other in SVGGen.hs (the output graph).

Note: Now that the ECE prefix is contained within the node, and the element IDs account for all node text, I have removed the ECE prefix addition on line 39 of `rect.py`.

### DIFF
--- a/utilities/svg_parsing/rect.py
+++ b/utilities/svg_parsing/rect.py
@@ -34,6 +34,11 @@ class Rect:
         self.colour = '#fff'
         self.class_ = 'hybrid' if self.hybrid else 'node'
 
+        self.input_text_y = 0 # The y value of the first text value seen by the parser.
+                              # When a second text value for a node is seen, we want
+                              # to be able to identify which text element goes above,
+                              # and which goes below.
+
     def output_haskell(self):
         if self.hybrid:
             self.colour = "#bbb"
@@ -53,8 +58,10 @@ class Rect:
                 self.area = area
                 break
 
-        
-
+        # Certain nodes have multiple lines of text. SVG text elements do not support
+        # line wrapping or newlines, so two text elements need to be created.
+        # Since there are now two text elements fitting into the node, the y positions
+        # of the text elements need to be recalculated to account for this.
         if len(self.extra_text) == 0:
             text = ("             S.text_ " +
                    " ! A.x \"" + str(self.text_x) +
--- a/utilities/svg_parsing/rect.py
+++ b/utilities/svg_parsing/rect.py
@@ -80,7 +80,6 @@ class Rect:
         print("S.g ! A.class_ \"" + self.class_ + "\" " +
               " ! A.id_ \"" + code + "\""
               " ! S.dataAttribute \"group\" \"" + self.area + "\""
-              " ! S.customAttribute \"text-rendering\" \"geometricPrecision\"" +
               " ! A.style \"" + "\" $ do \n"  +
               "             S.rect ! A.width \"" + self.width +
               "\" ! A.height \"" + self.height +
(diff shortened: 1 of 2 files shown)


Ian-Stewart-Binks:
It wasn't meant to work for more than two lines initially. It was just meant to facilitate future changes. However, I doubt that we will come across many multi-line text elements any time soon.

In any case, I have added the new code in. If you have any suggestions on how to lighten up on the dictionary work (the longer lines) I would love to hear it, the complexity of working with a dictionary is one of the downsides to this change.

Also, I have not yet been able to properly test the dynamic offset of the text elements for more than two text elements. Would you like me to, or should we tackle it when it comes up?

david-yz-liu:
The other reason I suggested using a dictionary is that it should actually be *easier* to write the code. Rather than using a loop, I recommend extracting the keys and values (`items`), sorting by keys, and going from there.

Ian-Stewart-Binks:
@david-yz-liu I made the changes we talked about today, but I have not yet fully reviewed my own work.

Using the new approach with the native y position of the text element has caused a problem that I am finding difficult to track down. On line 121 of `rect.py`, you will see that I adjust the translation by `4`. This is a magic number that I found using trial and error. The native y position is not correctly positioning the text elements, but the difference is very small. I have checked as many of the translation values that I have found, and they are all the native translation values from `graph_regions.svg`. Do you have any thoughts as to why this is happening?

To see the full problem, check out this branch, change line 121 in `rect.py`, regenerate `SVGGen.hs`, and run the server. 

Thanks.

utilities/svg_parsing/rect.py
```diff
@@ -36,23 +43,42 @@ def output_haskell(self):
         if self.hybrid:
             self.colour = "#bbb"
         prefix = ""
-        if self.text == '385' or self.text == '489':
-            prefix = 'ECE'
-        elif not self.text[0].isalpha():
+        sorted_text = sorted(self.text.items()) 
+
+        if not sorted_text[0][1][0].isalpha():
```
david-yz-liu: Please put the "self.hybrid" check first.

utilities/svg_parsing/rect.py
```diff
@@ -36,23 +43,42 @@ def output_haskell(self):
...
             prefix = "CSC"
         if self.hybrid:
             prefix = "hCSC"
 
-        #Figure out the research area
-        code = (prefix + self.text)[:6]
+
+        # Figure out the research area
+        code = (prefix + sorted_text[0][1] +
+               (sorted_text[1][1] if len(self.text) > 1 else ""))[:6]
```
david-yz-liu: This needs to be indented by one more space.

david-yz-liu:
@Ian-Stewart-Binks I'm not sure why this this happening, but if can be fixed by a manual offset then you can leave it like that for now.

Ian-Stewart-Binks:
It's mind breaking. I've checked all of the filtered out CSS styles (like `text-rendering`), font sizes, translations and y coordinates. They all seem correct. Nothing else seems to fix it. I'll leave it as is for now and leave a comment as well. Perhaps the act off actually applying the translation does something funnier? That's a pretty big stretch, though.

Ian-Stewart-Binks:
There was an issue with node id naming. hCSC20 defaulted to CSC207, so 320 would appear to have 207 as a prerequisite.

Issue #214, opened 2014-12-18
Graph: 465 Is Displayed With The Wrong Colour


Ian-Stewart-Binks:
In the new graph, CSC465 is grey. It should be red.

PR #250, opened 2015-01-06
Issue 214: 465 Displays Wrong Colour Fix

christinem:
This fixes Issue #214. Turns out CSC465 was not in the 'se' array in rect.py, so it was being interpreted as 'core'  when outputted to SVGGen.hs.


Ian-Stewart-Binks:
Nice!

Issue #241, opened 2015-01-03
Graph: CSS Error 


david-yz-liu:
When you load the graph, an error message "Unexpected value strokewidth parsing markerUnits attribute." is displayed in the console.

Ian-Stewart-Binks:
There is a chance that the bug may exist in the file `courseography/utilities/svg_parsing/svg_parser.py` around line 58. I am not certain of this, however.

Ian-Stewart-Binks:
PR merged. Is this still good to close?

In any case, it was mentioned that there are still potential jQuery errors happening. I'll make an issue for that.

david-yz-liu:
Thanks, @Ian-Stewart-Binks

PR #256, opened 2015-01-11
Issue-241: CSS Error

pkukulak:
Resolved issue #241 : Unexpected value strokewidth parsing markerUnits attribute." It is unexpected because markerUnits expects "strokeWidth", not "strokewidth." Thanks for pointing me in the right direction, @Ian-Stewart-Binks.

setup.py should be run once again to reflect the changes in svg_parser.py.


Ian-Stewart-Binks:
No problem, Phil! Good work!

Issue #243, opened 2015-01-03
Bool ellipses stroke should not be full opacity when inactive


david-yz-liu:
They should match the stroke style of the inactive nodes.

david-yz-liu:
@jmatholia

PR #257, opened 2015-01-11
Issue 243: Style, bool node opacity when inactive

jmatholia:
Fixed Issue #243: The boolean nodes had high opacity when they were not satisfied. To fix it I added the faded characteristic to the inactive bool nodes in cssGen.hs. setup.py should be run again in order to regenerate the css files.


Ian-Stewart-Binks:
Good Job, Jahnavi!

PR #259, opened 2015-01-13
Issue 233: Hybrid ID fix

Ian-Stewart-Binks:
Hybrid IDs would contain things like `/` and `,`. The hybrid ids are now guaranteed to be unique and contain HTML5 acceptable characters. The IDs are now of the format `h12`.


david-yz-liu:
Nice job, @Ian-Stewart-Binks. I noticed that in your previous PR, you accidentally overrode Phil's change. I fixed this, though.

PR #269, opened 2015-01-14
Add fb login button

Ian-Stewart-Binks:
This pull request adds in the Facebook 'login' button. The button is added in `MasterTemplate.hs`, so it can be accessed in both the Grid and in the Graph.

When a user is logged in, there name is retrieved from Facebook and added to the navigation bar. This is just meant to be an exercise in using the API through JavaScript.

When a user is logged in, they then are given the option to 'log out'.

I have also been playing around with the `fb` Haskell bindings. I can't find a good way to implement the 'log in' button with `fb` yet, there might not be one. I am still playing around with it, though.

Please try this out on your own machine, and let me know if it works.

Note that the App ID is fine to publish on GitHub. The App Secret is what we need to lock up.

### DIFF
--- a/hs/MasterTemplate.hs
+++ b/hs/MasterTemplate.hs
@@ -32,6 +32,7 @@ header page = createTag H.nav "" "row header" $ do
     H.li ! A.class_ "fb-login-button" 
          ! H.customAttribute "data-max-rows" "1" 
          ! H.customAttribute "data-size" "medium"
+         ! H.customAttribute "autologoutlink" "true"
          ! H.customAttribute "data-show-faces" "false"
          ! H.customAttribute "data-auto-logout-link" "false" $ ""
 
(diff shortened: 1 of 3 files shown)


Ian-Stewart-Binks:
One thing I'm not too pleased with is the lag in the setup of the login button, and in removing the username from the nav bar. In future pull requests I will try to make this faster.

david-yz-liu:
Got this message: 

```
App Not Setup: The developers of this app have not set up this app properly for Facebook Login.
```

BTW, please add me as a developer of this app.

Ian-Stewart-Binks:
I have added you as an administrator.

Do you still receive the message? I'm using a test app.

david-yz-liu:
Yep, it works! How exciting!

Ian-Stewart-Binks:
I know, right! Thanks!

PR #273, opened 2015-01-17
Er diagram

Ian-Stewart-Binks:
This pull request adds in a file that can be run through an ER generator to produce an ER diagram for Courseography.

I have decided to generate the ER diagram using a library called [erd](https://github.com/BurntSushi/erd).

It is a Haskell library that can be used to generate ER digram PDF files.

One of the dependencies of erd is `base-7.*`, however it appears that this dependency can be replaced with `base-6.*`.

This is not quite done yet. This PR is meant to demonstrate progress on the matter. I have included the PDF file so that anyone checking out this pull request can view the result without using erd.

I also attempted to create my own ER generator using GraphViz, however it became clear that doing that would take a lot longer than is necessary for this task. I have gotten my feet wet with Haskell GraphViz, however.

Note that this is much easier to update than a yEd graphml file.

The multiplicity of some of the relationships may be incorrect. For instance, I am uncertain whether *every* course has a breadth/depth/lecture.

The Wiki would need to be updated to include a page about this matter.


david-yz-liu:
@Ian-Stewart-Binks While I really like the concept, I'm not going to merge this until we do not specify the schema in two different places.

In the mean time, what would be really valuable is exporting what you have to an image and putting it on the wiki.

Ian-Stewart-Binks:
@david-yz-liu Understood, and will do.

PR #274, opened 2015-01-19
Add fb login button demo

Ian-Stewart-Binks:
This pull request explores what user data we can get from Facebook, and how to do it through the server (Happstack) and on the client side (JavaScript).

The Scripts.hs links have been changed to be absolute, however there is most likely a better way to achieve this. This is only meant as a temporary fix, I will create an issue for this.

Note that the secret must be inserted to successfully connect to Facebook. Only David and I have access to the secret right now.

This is built on top of #269.


hs/Scripts.hs
```diff
@@ -11,39 +11,39 @@ import MasterTemplate
...
-                                           "static/js/common/utilities/util.js",
-                                           "static/js/graph/utilities/structs.js",
-                                           "static/js/graph/utilities/util.js",
-                                           "static/js/graph/create_data.js",
-                                           "static/js/graph/parse_graph.js",
-                                           "static/js/graph/mouse_events.js",
-                                           "static/js/graph/setup.js"])
+                                            "http://code.jquery.com/ui/1.10.4/jquery-ui.js",
+                                            "http://localhost:8000/static/js/graph/modal.js",
+                                            "http://localhost:8000/static/js/graph/objects/edge.js",
```
david-yz-liu: Why did you change these?

hs/server.hs
```diff
@@ -32,9 +34,24 @@ about = "about"
 static :: String
 static = "static"
 
+app :: FB.Credentials
+app = FB.Credentials "localhost" "442286309258193" "INSERT_SECRET"
```
david-yz-liu: It would be helpful to add a comment here.

hs/Scripts.hs
Ian-Stewart-Binks: A temporary fix for #277.

hs/server.hs
Ian-Stewart-Binks: I've now added a helpful comment.

hs/Scripts.hs
david-yz-liu: Hmmm okay... I'll ask you about this later.

Ian-Stewart-Binks: I can give you a demonstration tomorrow, it shouldn't take more than a minute.

Ian-Stewart-Binks: Actually, even easier: Run the server on your happstack branch, and go to `http://localhost:8000/graph/`- there should be a surprise waiting there.

PR #276, opened 2015-01-19
Initial layout

christinem:
This PR is an initial layout for the Check My Post section. Code might be a little messy and probably needs some cleaning up and editing. Just an initial idea of how I think the layout could look like, and up for discussion.  

As well, a question (that I'll probably bring up during the meeting) - I know that all the html and css files are now generated through Haskell - I wonder if you want all html and css files to be generated this way (since I've coded the html and css straight).

### DIFF
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -1,26 +1,30 @@
 {-# LANGUAGE OverloadedStrings #-}
 
+module CssGen where
+
 import Clay
 import Prelude hiding ((**))
 import Data.Monoid
 import Data.Text.Lazy
 import System.Directory
 
+
 styleFiles :: [(String, Css)]
 styleFiles = [
-    ("style/common/common.css", common),
-    ("style/graph/graph_styles.css", graphStyles),
-    ("style/grid/timetable_styles.css", timetableStyles),
-    ("style/common/about.css", aboutStyles)
+    ("../style/common/common.css", common),
+    ("../style/graph/graph_styles.css", graphStyles),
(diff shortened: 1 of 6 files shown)


Ian-Stewart-Binks:
Cool! 

This reminds me, we will need to do some screen resolution checking before April.

I also think that the POSt styling and the Courseography styling should match up, however we haven't actually 'set' a style scheme for Courseography yet.

I am also a firm believer that all CSS and HTML must be generated with Blaze-HTML and Clay. Translating files over to Clay and Blaze-HTML shouldn't be tough, the files/templates that are there serve as pretty good examples, and I can help anyone who gets stuck on anything during this process.

Ian-Stewart-Binks:
Also, I put together some [JS etiquette](https://github.com/Courseography/courseography/wiki/JavaScript-Code-Etiquette) last semester, and hope to have fairly similar JS code across the code base. These are small things, such as the use of `'` instead of `"`. I realize that this code isn't going straight into `happstack`, but just a reminder before that day comes.

christinem:
@Ian-Stewart-Binks ah yes, this layout has a bit of an issue when the browser size is smaller (especially with width), something I plan to work on. 
I agree with you about the styling - that stuff can be easily changed once we agree on a style scheme. 
Yes, thanks for the reminder about that! Yeah, the main aim of this is just to get a visual idea out - I am sure the code itself needs some clean up :)

david-yz-liu:
@christinem Some double-quotes and tabbing issues.

christinem:
@david-yz-liu Made fixes to the tabbing and double quotes.

PR #280, opened 2015-01-19
CSC465 missing from create-data.js fix

christinem:
I noticed, while looking through the javascript files, that - just like in what I fixed for PR #250 - CSC465 was missing in the 'se' array in create-data.js. I've just added it back in as I did with the previously referenced pull request.


david-yz-liu:
@christinem Good catch, although I doubt this file will be around for much longer.

Issue #217, opened 2014-12-19
Graph: Opacity Of Regions Is Too High


Ian-Stewart-Binks:
The opacity of the graph regions is too high. Perhaps something around 0.2 may work better.

david-yz-liu:
It would be good to put a class on the regions during the parsing.

PR #282, opened 2015-01-21
Fix for issue 217: Made region class in region.py and decreased opacity ...

jmatholia:
...in CssGen.hs

Fix for issue #217, added a class for regions in region.py (utilities/svg_parsing/region.py) and added a faded style for the region class in CssGen.hs.

This means that the setup.py must be executed again to regenerate the CSS files.

BEFORE: 
image

AFTER:
image

Due to this change, the colours of the nodes are more apparent and I am not sure if is a pleasant change for some of them (web and databases courses look orange, and numerical computing courses look grey).

### DIFF
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -115,6 +115,7 @@ graphStyles = do
     titleCSS
     modalCSS
     regionCSS
+    regionShapeCSS
 
 alignCenter = textAlign $ alignSide sideCenter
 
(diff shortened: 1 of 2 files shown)


jmatholia:
image

image

Ian-Stewart-Binks:
Thanks for the photos @jmatholia !

I agree that the colours for Numerical Computing and Web/DB could be improved. The regions of these two sub-disciplines contrasts quite a lot with the colours of their nodes relative to the contrast of the other regions.

david-yz-liu:
Thanks, @jmatholia!

Issue #234, opened 2014-12-30
Update CSC Prerequisites


arkon:
CSC309 no longer requires CSC343 (or at least it didn't this past semester).

david-yz-liu:
Okay, I checked with the Undergraduate office. Two changes should be made to the graph:

- [ ] Remove 343 from 309
- [ ] Remove Sta2 from 411

@christinem @pkukulak @jmatholia @shaftoes Working on one or both of these changes should be an easy fix for any of you.

christinem:
@david-yz-liu I can make these changes.

PR #283, opened 2015-01-21
Issue 234: Removed prereqs for CSC309 and CSC411

christinem:
This fixes issue #234. Ofcourse, in the respective files for the courses in res/courses, the prereq strings still contain the removed prerequisites, but if I am correct, there is not much we can change with that yet since it parses right from the FAS Calendar.

### DIFF
--- a/res/graphs/graph_regions.svg
+++ b/res/graphs/graph_regions.svg
@@ -90,10 +90,10 @@
      inkscape:cx="552.36748"
      inkscape:cy="308.88099"
      inkscape:document-units="px"
-     inkscape:current-layer="layer1"
+     inkscape:current-layer="g4134"
      showgrid="false"
-     inkscape:window-width="1280"
-     inkscape:window-height="697"
+     inkscape:window-width="1366"
+     inkscape:window-height="705"
      inkscape:window-x="-8"
      inkscape:window-y="-8"
      inkscape:window-maximized="1" />
@@ -2647,7 +2647,8 @@
            xml:space="preserve"
            y="261.46799"
            clip-path="url(#clipPath2)"
-           id="text4707">Sta1,Sta2</text>
+           id="text4707"
+           transform="translate(10,-1.25)">Sta1</text>
         <rect
            style="fill:none"
            x="526.78949"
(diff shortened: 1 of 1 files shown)


Ian-Stewart-Binks:
@christinem The STA2 requirement change for CSC411 now reduces the number of courses in the hybrid to 1, however the size of the box is still much larger than every other hybrid that has only one course.

I think that we should change the size of the hybrid to match the others in order to maintain consistency and to emphasize repetition (following CRAP). What are your thoughts?

christinem:
@Ian-Stewart-Binks Yeah, I was thinking that the box is a little large now. I agree with adjusting the size. I'll fix that so that it matches the other hybrids with only one course.

Ian-Stewart-Binks:
@christinem It looks like the previous commit broke the SVG parser. I think that the hybrid text now has a new translation applied to it that was added by InkScape. The SVG parser now can't find the text node for the corresponding rect node.

This should ultimately be put into the parser, but for right now are you able to manually change the `x` and `y` coordinates of the text based on its translation? I can help if need be.

christinem:
@Ian-Stewart-Binks yeah I noticed the error as soon as I pushed the commit and then tried to run setup.py again. Seems that when I try change the size of the grey fill, it breaks the parser. 

What I was doing right now was adjusting the size of the box and fill using the "Edit paths by node" option. I tried to just edit the x and y coordinates manually from the "Select and Transform Objects" option as well, and it causes a different error. If there's a better way to do it, please, I would appreciate the help :) I'm not too acquainted with how Inkscape and the SVG parser works yet entirely.

Ian-Stewart-Binks:
The error seems to be : 
```
File "......../courseography/utilities/svg_parsing/rect.py", line 58, in output_haskell
    code = (prefix + sorted_text[0][1] +
IndexError: list index out of range
```

This means that the parser couldn't find a corresponding text element to match with the rect element.

The parser determines if a text element is within a rect element based on the x and y positions - and the translation. The parser is good at getting the x and y positions, but not good at getting the translation. It looks like there may be some extra translation that the parser did not pick up on, and consequently can't match the text element to the rect element.

This is, of course, a flaw of the parser.

To change this manually, you can go into the SVG file and find the offending text element, see the translation that was added, and change the x and y coordinates based on that translation.

If you check out the diffs on your commits, you'll see that your first commit actually adds in some translation that wasn't there before. This isn't your fault, and is completely expected- we want to be able to support any type of InkScape output, even if the file has been changed.

The fix would only be a superficial fix for the graph, and wouldn't change the parser, which is what really requires the fix.

I can take a look at the parser, however. That might be the best choice ultimately.

Ian-Stewart-Binks:
Looks good! I'll create an issue for the parser.

christinem:
@Ian-Stewart-Binks great, and thanks for the suggestion! Turns out changing the svg file manually (without using InkScape) was the key, I'll remember that for the future to make things easier.

Ian-Stewart-Binks:
@christem No problem! And don't be afraid to use InkScape to change the file- the parser should be strong like rock, and finding holes in the parser only means that we can make it stronger.

david-yz-liu:
So... using Inkscape inserted a new translation automatically, eh? Interesting.

PR #287, opened 2015-01-25
Exporting

pkukulak:
Added "Export to PDF" to the navbar, which directs to a new response page. It currently displays a picture of my cat :)

### DIFF
--- a/hs/ExportResponse.hs
+++ b/hs/ExportResponse.hs
@@ -1,0 +1,35 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module ExportResponse where
+import           Text.Blaze ((!))
+import qualified Text.Blaze.Html5 as H
+import qualified Text.Blaze.Html5.Attributes as A
+import Happstack.Server
+import MakeElements
+import MasterTemplate
+import Graphics.PDF
+
+exportResponse :: ServerPart Response
+exportResponse =
+   ok $ toResponse $
+    masterTemplate "Courseography - SVG serving test!"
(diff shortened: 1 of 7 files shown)


david-yz-liu:
@pkukulak Nice work so far! Let me know if you're having trouble generating the PDF.

pkukulak:
@david-yz-liu Generating the PDF is no problem; the library comes with a pretty thorough test file that illustrates what needs to be done. I'm hoping to have a download link on the page shortly.

david-yz-liu:
@pkukulak Great!

Ian-Stewart-Binks:
Nice @pkukulak !

hs/ExportResponse.hs
```diff
@@ -0,0 +1,35 @@
...
+exportResponse =
+   ok $ toResponse $
+    masterTemplate "Courseography - SVG serving test!"
+                [H.meta ! A.name "keywords"
+                        ! A.content "",
+                 aboutLinks
+                ]
+                (do
+                    header "export"
+		    exportHtml
```
Ian-Stewart-Binks: We haven't exactly set any constraints on `do` block style, and quite frankly I think I've made my own up somehow (which is probably a bad thing), but please align these two lines.

pkukulak:
Yeesh. Didn't even notice. Fixing now.

pkukulak:
This is turning out to be quite difficult. I think that Haskell is my primary obstacle at the moment. I will need more time to work this out.

PR #288, opened 2015-01-25
Update timetable to have double the number of rows

foreverwander:
Updated timetable so twice the number of rows should be added as the half hour time slots.

Please pardon the 10 commits, I was trying to test out my localhost and I don't really know how to get rid of these...

Also I understand the "time" setting is not the most efficient and likely changes a lot of things, but that is to be worked on.


setup.py
```diff
@@ -1,2 +1,2 @@
 import os
-os.system("python3 ./utilities/svg_parsing/svg_parser.py > ./hs/SVGGen.hs")
+os.system("python ./utilities/svg_parsing/svg_parser.py > ./hs/SVGGen.hs")
```
david-yz-liu: @foreverwander I think this setup.py file is problematic for this reason, but you shouldn't change it in this pull request. (So you should add another commit to remove it.)

david-yz-liu:
@foreverwander You're definition going in the right direction! One complaint I have is I don't think it's necessary to have time labels for each half-hour. I would much prefer keeping the label field empty for the new half-hour rows.

Ian-Stewart-Binks:
Nice work @foreverwander ! I'm going to make some small coding style comments, but don't be discouraged by them.

js/grid/generate_grid.js
```diff
@@ -91,6 +91,11 @@ function appendTableRows(timetableTableFall, timetableTableSpring) {
         appendTableData(trFall, trSpring, i);
         timetableTableSpring.append(trSpring);
         timetableTableFall.append(trFall);
+        var trFall = $('<tr></tr>');
+        var trSpring = $('<tr></tr>');
```
Ian-Stewart-Binks: `trFall` and `trSpring` were declared above. There is no need to use `var` here.

js/grid/generate_grid.js
```diff
@@ -106,7 +111,12 @@ function appendTableData(trFall, trSpring, time) {
 
     var weekPrefixArray = ["M", "T", "W", "R", "F"];
 
-    var adjustedTime = (time === 12 ? 12 : time % 12) + ':00';
+    if (time % 1 == 0){
```
Ian-Stewart-Binks: Please put a space between the final `)` and the `{`.
Also, we've been using the [`===` over the `==`.] (http://stackoverflow.com/questions/523643/difference-between-and-in-javascript)

js/grid/generate_grid.js
```diff
@@ -106,7 +111,12 @@ function appendTableData(trFall, trSpring, time) {
...
+        var adjustedTime = (time === 12 ? 12 : time % 12) + ':00';}
```
Ian-Stewart-Binks: The final brace, `}` should be located on the next line, beside the `else`. I've added a small example as to how this should look [here](https://github.com/Courseography/courseography/wiki/JavaScript-Code-Etiquette).

js/grid/generate_grid.js
```diff
@@ -106,7 +111,12 @@ function appendTableData(trFall, trSpring, time) {
...
+    else {
+    	time -= 0.5
```
Ian-Stewart-Binks: It looks like time isn't indented exactly 4 spaces.

setup.py
foreverwander: Yeah sorry I didn't know how to create a pull request JUST for the js file (is that even possible?). I'll remove the change right now.

js/grid/generate_grid.js
```diff
@@ -121,6 +131,7 @@ function appendTableData(trFall, trSpring, time) {
             .attr("satisfied", "true")
             .addClass("timetable-cell"));
     }
-
+    
     trSpring.append($("<td></td>").addClass("timetable-time").html(adjustedTime));
+
```
david-yz-liu: Remove this blank line.

js/grid/generate_grid.js
```diff
@@ -121,6 +131,7 @@ function appendTableData(trFall, trSpring, time) {
             .attr("satisfied", "true")
             .addClass("timetable-cell"));
     }
-
+
```
david-yz-liu: This is an indication you have added some trailing whitespace; this should be removed.

david-yz-liu:
@foreverwander I made two small comments; please fix these, and I will merge in these changes.

PR #296, opened 2015-01-26
Fb post button

Ian-Stewart-Binks:
This pull request allows the user to post a message to facebook ("Test post please ignore").

The server is getting a bit clunky, I plan to soon separate the FB code from the server code.

Some of the code that this pull request will be simplified in the future, given recent progress on understanding how to properly use `fb`.

### DIFF
--- a/hs/server.hs
+++ b/hs/server.hs
@@ -9,6 +9,7 @@ import qualified Data.ByteString.Lazy.Char8 as BSL
 import qualified Data.ByteString.Char8 as BS
 import Control.Monad    (msum)
 import Network.HTTP.Conduit (withManager)
+import Data.Conduit
 import Happstack.Server
 import GridResponse
 import GraphResponse
@@ -75,6 +76,7 @@ postFB = "post-fb"
 
 main :: IO ()
 main = do
+    print "Server is ready. . ."
     cwd <- getCurrentDirectory
     let staticDir = encodeString $ parent $ decodeString cwd
     simpleHTTP nullConf $
@@ -96,7 +98,7 @@ args2 :: FB.Argument
 args2 = ("message", "Test post please ignore")
 
 performPost :: String -> IO Response
-performPost code = withManager $ \manager -> FB.runFacebookT app manager $ do
+performPost code = performFBAction $ do
         postToFB code =<< getToken url2 code
         return $ toResponse postFB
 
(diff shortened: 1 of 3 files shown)


hs/server.hs
```diff
@@ -69,17 +79,38 @@ main = do
...
+
+args2 :: FB.Argument
+args2 = ("message", "Test post please ignore")
+
+performPost :: String -> IO Response
+performPost code = withManager $ \manager -> FB.runFacebookT app manager $ do
+        token <- FB.getUserAccessTokenStep2 url2 [args code]
+        u <- FB.getUser "me" [] (Just token)
+        postToFB code token
+        return $ toResponse $ postFB
```
david-yz-liu: You don't need the second `$`

hs/server.hs
```diff
@@ -69,17 +79,38 @@ main = do
...
            ]
 
+postToFacebook :: String -> ServerPart Response
+postToFacebook code = (liftIO $ performPost code) >> graphResponse
+
+args2 :: FB.Argument
+args2 = ("message", "Test post please ignore")
+
+performPost :: String -> IO Response
+performPost code = withManager $ \manager -> FB.runFacebookT app manager $ do
```
david-yz-liu: It would be good to refactor the "Step 2 Authentication" from this and the function below.

Ian-Stewart-Binks: I completely agree.

Ian-Stewart-Binks: @david-yz-liu I had initially thought that you were referring to the line that you addressed, however I refactored based on the content of your comment. Is this what you had in mind, or would you rather refactor `withManager $ \manager -> FB.runFacebookT app manager $ do`? The latter probably makes a lot of sense, however it will require some deeper thought.

hs/server.hs
Ian-Stewart-Binks: Done. Thanks.

hs/server.hs
david-yz-liu: I did mean the "withManager" part too, although your change is also appreciated.

It shouldn't be too hard to define a function that basically does `withManager $ \manager -> FB.runFacebookT app manager` and takes in the rest as an argument.

david-yz-liu:
Oh, and you should be able to use "FlexibleContexts" as part of the header of a file.

Ian-Stewart-Binks:
I'll add that in.

hs/server.hs
Ian-Stewart-Binks: This step was a bit tricky when I tried it yesterday, but perhaps I was thinking about it in the wrong way. The `return` call makes things a bit more difficult. I'll see what I can do.

Ian-Stewart-Binks: This is going to take some more time, I haven't been able to make progress on this just yet.

Ian-Stewart-Binks: I should really stop working when I write something like that.

I believe I have found the solution.

I can't test it yet though, I can't get through to Facebook (even as a regular user). Will test tomorrow.

fberrir

Ian-Stewart-Binks: It works.

PR #297, opened 2015-01-26
Check my POSt (following UGO Handout)

christinem:
This is associated with #165. I've employed the idea that David and Paul suggested during the last meeting - listing all the requirements together as is done on the UGO handout, per POSt. In this commit I've also added a link to the Check My Post! page on the nav bar (once the html is generated through haskell I will create a proper Response page for it, right now it just links to the html page).


david-yz-liu:
@christinem You have multiple whitespace issues; it seems you are still using tabs instead of spaces. Again, you need to fix this, and configure your editor to not let this happen.

christinem:
@david-yz-liu my mistake, I had thought I had configured my editor properly. I think I've fixed the remaining whitespace issues.

post/change_div.js
```diff
@@ -0,0 +1,45 @@
...
+	
+    e.preventDefault();
+    resetAttributes();
+    $('#div_major').show();
+    $('#major').css('background-color', '#9C9C9C');
+});
+
+
+$('#minor').click (function (e) {
+    'user-strict';
```
david-yz-liu: use not user

david-yz-liu:
@christinem I'm not actually sure there's a real conflict; this time only you can do a pull from happstack and update your pull request (along with the fix for my comment). Then if there's still a problem, I can help deal with it.

A layout comment: For the conditions where you can "choose X FCEs from this list", I think the handout from the UGO is much better than what we have currently. I'd want to see a lot more text input fields for next week.

post/change_div.js
```diff
@@ -0,0 +1,45 @@
+$(document).ready(function() {
```
Ian-Stewart-Binks: Missing space between `function` and `()`.

christinem: Right, fixed.

post/change_div.js
christinem: Fixed.

christinem:
@david-yz-liu I made the fix, and I pulled happstack into the branch as well, but there still seems to be a merge conflict. I'm not entirely sure where it is. 

As for the layout comment, yes I see what you're saying. I can definitely have that incorporated in for next week.

david-yz-liu:
@christinem Yeah, that was weird. Oh well, it's done now.

PR #304, opened 2015-01-27
Web request

shaftoes:
Preliminary Arts and Science parser: currently outputs all course codes


david-yz-liu:
@shaftoes I'll take a detailed look in a bit, but for now... please remove your output.txt file. (`git rm path/to/output.txt` should work)

shaftoes:
done

hs/web_parsing/ArtSciParser.hs
```diff
@@ -1,34 +1,72 @@
...
+----------------------------------------------------------------------------------------}
+toDelete :: [String]
+toDelete = ["199299398399Big_Ideas_(Faculty_of_Arts_&_Science_Programs).html",
+            "Joint_Courses.html",
+            "Writing_in_the_Faculty_of_Arts_&_Science.html"]
+
+{----------------------------------------------------------------------------------------
+INPUT: 'processed' main page
+OUTPUT: a list of department html page names
+-----------------------------------------------------------------------------------------}
```
Ian-Stewart-Binks: We should come up with a uniform way of documenting Haskell code.

Ian-Stewart-Binks:
We should make it so that any hs file of Courseography can interact with any other file in Courseography without specifying the directory to the compiler.

@david-yz-liu Is there a way to do this without making Courseography a package? Should we make our own package?

david-yz-liu:
@Ian-Stewart-Binks I don't think so; making a package is an excellent idea, but I'm not going to do it in the middle of the term.

@shaftoes Can you try specifying the path to Tables.hs as a compilation flag, e.g. `runhaskell -i../database ArtSciParser.hs`? This should eliminate the need for a second Tables.hs.

david-yz-liu:
@shaftoes I merged your changes in, but make sure to remove Tables.hs by next week.

PR #305, opened 2015-01-28
Issue 295: Clean up server.hs.

Ian-Stewart-Binks:
This pull request separates database code from response code and server code.

I have also taken the path filters from their definitions and put them directly in the list that is passed into `msum`. I discovered that this was unnecessary while making some documentation.


Ian-Stewart-Binks:
Not certain that I like the file name 'CourseQueries.hs' though. Couldn't think of a better name.

PR #310, opened 2015-01-30
Initial Schema

Ian-Stewart-Binks:
The pull request creates the initial SVG schema. The schema will be extended to be able to manage other data as necessary (for instance: classes).

Any suggestions are welcome, as always.


hs/database/Tables.hs
```diff
@@ -65,6 +65,32 @@ Distribution
...
+    height Rational
+    xPos Rational
+    yPos Rational
+    colour String
+    deriving Show
+
+Texts
+    gId Int
+    rId Int
+    yPos Rational
```
david-yz-liu: Hmmm. Text which belongs to a rect should be centred inside the rect.

Text that doesn't belong to a rect (region labels?) need to have both x and y positions.

All three of these attributes (rId, yPos, xPos) should be Maybes.

hs/database/Tables.hs
```diff
@@ -65,6 +65,32 @@ Distribution
...
+    text String
+
+Paths
+    gId Int
+    d String
```
david-yz-liu: What I'd like to see for Paths is a list of points `[(Rational, Rational)]`.

Ian-Stewart-Binks: Done.

hs/database/Tables.hs
Ian-Stewart-Binks: Text which belongs to a rect should not always be centered inside the rect.

Ian-Stewart-Binks: Change has been made, as well.

david-yz-liu:
Thanks for reminding me about the text positioning; I really hope we come up with something better some day.

If this is the case, it makes sense to make the `xPos` and `yPos` just Rationals, and not Maybes.

Ian-Stewart-Binks:
It may be best to impose some sort of line height on all nodes, or some sort of uniform formatting for nodes with multi-line text (eventually).

david-yz-liu:
We can do this when putting data into the database.

Ian-Stewart-Binks:
Changed `xPos` and `yPos` back to `Rational`s.

david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #311, opened 2015-01-30
Timetable - Hide half hours and non-mainstream hours

foreverwander:
Updated generate_grid so that now all of the full hour time blocks have row span of 2, and all of the half hour blocks are created but set to hidden.

Also took Ian's suggestion to make the time slots after 5pm hidden initially, and also the 8am time slot.

The only problem is that I don't know how to change the "hidden" attribute, setting it to "false" doesn't seem to do anything...

### DIFF
--- a/js/grid/generate_grid.js
+++ b/js/grid/generate_grid.js
@@ -109,7 +109,7 @@ function appendTableRows(timetableTableFall, timetableTableSpring) {
 function appendTableData(trFall, trSpring, time) {
     'use strict';
 
-    var weekPrefixArray = ["M", "T", "W", "R", "F"];
+    var weekPrefixArray = ['M', 'T', 'W', 'R', 'F'];
 
     if (time % 1 === 0) {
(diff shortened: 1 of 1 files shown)


david-yz-liu:
I think you want the display attribute... https://developer.mozilla.org/en-US/docs/Web/CSS/display

foreverwander:
oh okay i see! so I would use that eventually with the mouse interactions?

foreverwander:
I also changed the quotation marks while i was at it, should fix issue#289

david-yz-liu:
You want the cells containing the times also have rowspan 2.

Also, I'm not too big on the 9-5 right now. Please change that back.

foreverwander:
Okay I understand.

But now all of the rows just look like as if they have rowspan of 1, is that normal?

david-yz-liu:
Indeed. The grid should looks exactly as it was, except (eventually) half-hour blocks will take up half the row height.

Issue #290, opened 2015-01-25
Grid - course names


david-yz-liu:
Inside the table, we currently display selected sections with their full codes, "CSC108H1". I think we should do two things

1. Only display the dept and number, but omit the last two letters: "CSC108".
2. Specify either "(L)", "(T)", or "(P)" to distinguish lectures, tutorials, and practicals: "CSC108 (L)".

Ian-Stewart-Binks:
Interesting. I think that this is a good idea, and it shouldn't take too long to implement either one.

christinem:
I like the idea too, but (maybe this is long term) once UTM courses come in, those last two letters might be important (since UTM courses have alot of the same course codes but an "H5" at the end)

david-yz-liu:
I don't know that I care to distinguish between campuses on the timetable, though of course it should be in the course selection pane. It's something that people should have in their heads regardless of what appears in the timetable.

christinem:
@david-yz-liu ah, I misread a little bit, I realize you meant just on the timetable grid. I was thinking of the course selection pane as well. I agree then that it's not needed on the timetable itself.

christinem:
@david-yz-liu I think I'd like to take this issue on.

PR #312, opened 2015-01-30
Issue 290: Change course names

christinem:
This fixes Issue #290. Courses now display on the table without 'H1' at the end, and it displays in brackets the type of section it is - (L), (T), or (P).

### DIFF
--- a/js/common/objects/section.js
+++ b/js/common/objects/section.js
@@ -9,7 +9,7 @@ function Section(times, course, id) {
     'use strict';
 
     this.id = id;
-    this.courseName = this.id.substring(0, 8);
+    this.courseName = this.id.substring(0, 6);
     this.name = this.id.substring(9, 14);
     this.session = this.id.substring(15, 16);
     this.type = this.name.charAt(0);
--- a/js/common/objects/section.js
+++ b/js/common/objects/section.js
@@ -11,7 +11,7 @@ function Section(times, course, id) {
     this.id = id;
     this.name = this.id.substring(9, 14);
     this.type = this.name.charAt(0);
-    this.courseName = this.id.substring(0, 6) + " (" + this.type + ")";
+    this.courseName = this.id.substring(0, 6) + ' (' + this.type + ')';
     this.session = this.id.substring(15, 16);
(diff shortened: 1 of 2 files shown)


js/common/objects/section.js
```diff
@@ -9,10 +9,10 @@ function Section(times, course, id) {
     'use strict';
 
     this.id = id;
-    this.courseName = this.id.substring(0, 8);
     this.name = this.id.substring(9, 14);
-    this.session = this.id.substring(15, 16);
     this.type = this.name.charAt(0);
+    this.courseName = this.id.substring(0, 6) + " (" + this.type + ")";
```
Ian-Stewart-Binks: `"`

christinem: Eek, thanks, fixed. That's a habit I'll have to break.

js/common/objects/section.js
```diff
@@ -9,10 +9,10 @@ function Section(times, course, id) {
     'use strict';
 
     this.id = id;
-    this.courseName = this.id.substring(0, 8);
     this.name = this.id.substring(9, 14);
-    this.session = this.id.substring(15, 16);
     this.type = this.name.charAt(0);
+    this.courseName = this.id.substring(0, 6) + ' (' + this.type + ')';
```
david-yz-liu: You shouldn't change it here; change it when the HTML is rendered.

christinem: Thanks, I've changed it so that now the text is changed when the HTML is rendered.

js/common/objects/section.js
```diff
@@ -101,7 +101,7 @@ Section.prototype.onclick = function () {
 Section.prototype.setTime = function (time) {
     'use strict';
 
-    $(time).html(this.courseName)
+    $(time).html(this.courseName.substring(0,6) + ' (' + this.type + ')')
```
Ian-Stewart-Binks: The above calls to `substring` contain a space before the second parameter. I don't think we've outlined this in the wiki, but can you add this in here?

christinem: Done!

Issue #291, opened 2015-01-25
Grid - Remove colours for section types


david-yz-liu:
Related to #290, we should not use colours to distinguish between different kinds of sections (lecture/tutoral/practical).

To complete this issue, remove all styling which distinguishes the types of sections in the grid, both in the table and in the left sidebar.

Ian-Stewart-Binks:
Should we keep the styling that distinguishes between courses in which all required types of section have been taken (e.g. CSC209 lecture and tutorial)? I believe that the code refers to these states as 'satisfied' and 'unsatisfied'.

david-yz-liu:
@Ian-Stewart-Binks I think a warning icon (the yellow triangle with an exclamation mark) is better for that.

Ian-Stewart-Binks:
@david-yz-liu Yeah, that actually sounds a lot better than having different colours. Good idea.

PR #313, opened 2015-01-30
Issue 291: Remove section colours

christinem:
(Because I was on a roll.)
This fixes Issue #291. All sections now have the same colour (I just chose the colour for Lectures: blue3, unless anyone is opposed to this). 

This PR is built upon #312. That PR should be merged before this is considered.

### DIFF
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -588,8 +588,10 @@ tdColours = ".timetable " ?  do
             backgroundColor red3
             ":hover" & do
                 backgroundColor red4
-    td # ("in-conflict" *= "false") # ("satisfied" *= "true") # ("clicked" *= "true") ? do
-            backgroundColor blue3
+    td # ("in-conflict" *= "false") # ("satisfied" *= "true") ? do
+        backgroundColor blue3
+        ":empty" & do
+            backgroundColor white
             
 
 teal1 = parse "#737A99"
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -590,7 +590,7 @@ tdColours = ".timetable " ?  do
                 backgroundColor red4
     td # ("in-conflict" *= "false") # ("satisfied" *= "true") # ("clicked" *= "true") ? do
             backgroundColor blue3
-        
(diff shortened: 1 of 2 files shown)


hs/CssGen.hs
```diff
@@ -591,9 +591,9 @@ tdColours = ".timetable " ?  do
     td # ("in-conflict" *= "false") # ("satisfied" *= "true") # ("type" *= "L") ? do
             backgroundColor blue3
     td # ("in-conflict" *= "false") # ("satisfied" *= "true") # ("type" *= "T") ? do
-            backgroundColor teal1
```
david-yz-liu: Rather than repeat the selectors, why don't we just remove the `# ("type *= "L")`.

christinem: I actually tried that at first but it just made all the lines on the grid blue :/

christinem: Actually I could probably do it based on whether the clicked attribute is true or false. I'll make that change.

david-yz-liu:
Ah, interesting. This is a result of a lack of css knowledge when the styling was first done (my fault). You should add a new selector under td to match `td:empty` and set the background to be white in this case.

christinem:
@david-yz-liu Ah, alright, I can do that, thanks!

christinem:
@david-yz-liu Hmm, adding a td:empty selector doesn't seem to solve the problem as well. Seems as though the ".timetable  td[in-conflict*='false'][satisfied*='true']" selector is overriding it, since it's more specific.

However, adding a [clicked*='true'] (as I was originally thinking) to the original selector and taking out the type produces the desired effect.

david-yz-liu:
I'm not happy with adding more and more attributes in the selectors. You can nest an `:empty` inside the existing selector on line 591.

You also need to understand how nesting selectors works in Clay - you created a new selector, but this should be nested inside the td selector.

christinem:
@david-yz-liu thanks, I think I've fixed it through using ```:empty```

david-yz-liu:
Thanks!

PR #317, opened 2015-02-01
Moving HTML and CSS generating to Haskell

christinem:
This pull request has all the straight HTML and CSS code that I had written before, be generated through Haskell and run directly from the server.


hs/PostResponse.hs
```diff
@@ -0,0 +1,144 @@
...
+                H.a ! A.href "" $ "Specialist"
+                H.div ! A.id "spec_creds" $ "(0/12.0)"
+            H.li ! A.id "major" $ do
+                H.a ! A.href "" $ "Major"
+                H.div ! A.id "maj_creds" $ "(0/8.0)"
+            H.li ! A.id "minor" $ do
+                H.a ! A.href "" $ "Minor"
+                H.div ! A.id "min_creds" $ "(0/4.0)"
+        H.div ! A.id "div_specialist" $ do
+            H.p $ do
```
Ian-Stewart-Binks: The `$` and `do` blocks aren't necessary here. If you just have one item that you're applying a function to, (in this case, `H.p`), then you can just apply that function directly.

Ian-Stewart-Binks:
Nice work! There are some style issues with the `do`  and `$`.

christinem:
Thanks @Ian-Stewart-Binks, I have removed the redundant '''do'''s and '''$'''

hs/CssGen.hs
```diff
@@ -650,4 +652,79 @@ fceCountCSS = "#FCECountDiv" ? do
     "vertical-align" -: "middle"
     fontSize (em 1.35)
     alignCenter
-
+   
+-- Post Styles
+postStyles = do
+    body ?
+        do backgroundColor $ "#EBE8E4"
```
david-yz-liu: There are `$` here that are redundant; also, please try to use existing colours we've defined in this file, or just add your own. This will be helpful when looking at colours later on.

hs/MakeElements.hs
```diff
@@ -52,5 +52,9 @@ aboutLinks = concatHtml (map stylesheet ["//netdna.bootstrapcdn.com/bootstrap/3.
                                        "static/style/common/about.css",
                                        "static/style/common/common.css"])
 
+postLinks :: H.Html
+postLinks = concatHtml (map stylesheet ["static/style/post/post_styles.css",
+                                       "static/style/common/common.css"])
```
david-yz-liu: Put a space here

hs/Scripts.hs
```diff
@@ -44,4 +44,8 @@ timetableScripts = do jQuery
                                                 "static/js/grid/generate_grid.js",
                                                 "static/js/common/objects/course.js",
                                                 "static/js/common/objects/section.js",
-                                                "static/js/common/utilities/util.js"])
\ No newline at end of file
+                                                "static/js/common/utilities/util.js"])
+postScripts :: H.Html
+postScripts = concatHtml (map makeScript ["http://code.jquery.com/jquery-1.10.2.js",
```
david-yz-liu: @shaftoes changed the URL for the jquery import, which you should follow here. You can find it in #303

david-yz-liu:
@christinem Looking good, I just a few minor comments.

Issue #316, opened 2015-02-01
Adding a visual hover to nav bar


christinem:
I was thinking, it'd be nice to add something to more visually indicate a link you are hovering over on the main nav bar - either a different shade or just a change of colour in the text, or the link be underlined, perhaps.

Ian-Stewart-Binks:
Interesting! I agree.

PR #318, opened 2015-02-01
When links in the navbar are hovered over, they change colour to gray.

pkukulak:
This addresses #316. I've added to CssGen.hs. Now, when links are hovered over, they turn gray. Not particularly exciting, but it certainly looks nice.

Also, "Courseography" in the navbar is no longer bold.

### DIFF
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -77,8 +77,8 @@ headerCSS = do
                 padding 0 (px 10) 0 (px 10)
                 a <? do
                     color white
-                a # hover <? do
-                    color gray
+                    hover & do
+                        color gray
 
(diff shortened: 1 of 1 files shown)


hs/CssGen.hs
```diff
@@ -78,6 +77,8 @@ headerCSS = do
                 padding 0 (px 10) 0 (px 10)
                 a <? do
                     color white
+                a # hover <? do
```
david-yz-liu: You can actually use a nested selector here - something like `hover & do`

If you do a search through this file for `&`, you'll see how it's used.

pkukulak:
It should look a bit nicer now. If anyone has any suggestions on hover animations, let me know.

PR #320, opened 2015-02-02
Facebook: Cleanup & convert SVG to PNG

Ian-Stewart-Binks:
This pull request cleans up some of the Facebook code, and also allows the user to upload a photograph to Facebook (an SVG file that is converted to a PNG file).

### DIFF
--- a/hs/ConvertSVGToPNG.hs
+++ b/hs/ConvertSVGToPNG.hs
@@ -13,7 +13,7 @@ convertSVGToPNG :: String -> String -> IO
                       ProcessHandle)
 
 convertSVGToPNG inName outName = createProcess $ CreateProcess
-                                  (ShellCommand $ "convert ../res/graphs/graph_regions.svg graph.png"
+                                  (ShellCommand $ "convert ../res/graphs/graph_regions.svg INSERT_ID-graph.png"
                                    )
                                   Nothing
                                   Nothing
--- a/hs/ConvertSVGToPNG.hs
+++ b/hs/ConvertSVGToPNG.hs
@@ -1,0 +1,20 @@
+module ConvertSVGToPNG where
+
+import System.Process
+import GHC.IO.Handle.Types
+
+convertSVGToPNG :: IO
+                     (Maybe Handle,
+                      Maybe Handle,
+                      Maybe Handle,
+                      ProcessHandle)
+
+convertSVGToPNG = createProcess $ CreateProcess
+                                  (ShellCommand "inkscape -z -e test.png -w 1024 -h 1024 ../res/graphs/graph_regions.svg")
+                                  Nothing
+                                  Nothing
+                                  Inherit
+                                  Inherit
+                                  Inherit
+                                  False
+                                  False
--- a/hs/ConvertSVGToPNG.hs
+++ b/hs/ConvertSVGToPNG.hs
@@ -23,6 +23,7 @@ convertSVGToPNG inName outName = createProcess $ CreateProcess
                                   CreatePipe
                                   False
                                   False
+                                  False
 -- | Removes a file.
 removePNG :: String -> IO
                      (Maybe Handle,
@@ -37,6 +38,7 @@ removePNG name = createProcess $ CreateProcess
                                   Inherit
                                   CreatePipe
                                   CreatePipe
+                                  False
                                   False
                                   False
 
--- a/hs/ConvertSVGToPNG.hs
+++ b/hs/ConvertSVGToPNG.hs
@@ -3,14 +3,32 @@ module ConvertSVGToPNG where
 import System.Process
 import GHC.IO.Handle.Types
 
-convertSVGToPNG :: IO
+convertSVGToPNG :: String -> String -> IO
                      (Maybe Handle,
                       Maybe Handle,
                       Maybe Handle,
                       ProcessHandle)
 
-convertSVGToPNG = createProcess $ CreateProcess
-                                  (ShellCommand "inkscape -z -e test.png -w 1024 -h 1024 ../res/graphs/graph_regions.svg")
+convertSVGToPNG inName outName = createProcess $ CreateProcess
+                                  (ShellCommand $ "inkscape -z -e " ++
+                                   inName ++
+                                   " -w 1024 -h 1024 " ++
+                                   outName)
+                                  Nothing
+                                  Nothing
+                                  Inherit
+                                  Inherit
+                                  Inherit
+                                  False
+                                  False
+removePNG :: String -> IO
+                     (Maybe Handle,
+                      Maybe Handle,
+                      Maybe Handle,
+                      ProcessHandle)
+
+removePNG name = createProcess $ CreateProcess
+                                  (ShellCommand $ "rm " ++ name)
                                   Nothing
                                   Nothing
                                   Inherit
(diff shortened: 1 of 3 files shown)


hs/ConvertSVGToPNG.hs
```diff
@@ -0,0 +1,48 @@
...
+convertSVGToPNG inName outName = createProcess $ CreateProcess
+                                  (ShellCommand $ "convert ../res/graphs/graph_regions.svg INSERT_ID-graph.png"
+                                   )
+                                  Nothing
+                                  Nothing
+                                  CreatePipe
+                                  CreatePipe
+                                  CreatePipe
+                                  False
+                                  False
```
david-yz-liu: Was there a False missing here?

hs/CourseographyFacebook.hs
```diff
@@ -44,25 +52,29 @@ credentials :: FB.Credentials
 credentials = FB.Credentials (T.pack courseographyUrl) appId appSecret
 
 perms :: [FB.Permission]
-perms = []
+perms = ["publish_actions"]
+
+postPhoto :: FB.UserAccessToken -> FB.UserId -> IO Response
+postPhoto (FB.UserAccessToken _ b _) id_ = withSocketsDo $ withManager $ \m -> do
+    fbpost <- parseUrl $ "https://graph.facebook.com/v2.2/me/photos?access_token=" ++ T.unpack b
```
david-yz-liu: In the spirit of using Text instead of string, use T.++ here? (Does that exist?)

hs/CourseographyFacebook.hs
```diff
@@ -85,23 +97,24 @@ performFBAction action = withManager $ \manager -> FB.runFacebookT credentials m
 -- | Posts a message to the user's Facebook feed, with the code 'code'. GraphResponse
 -- is then sent back to the user.
 postToFacebook :: String -> ServerPart Response
-postToFacebook code = (liftIO $ performPost code) >> graphResponse
+postToFacebook code = (liftIO $ performPost (BS.pack code)) >> graphResponse
```
david-yz-liu: Isn't this the opposite of what we wanted to do?

hs/CourseographyFacebook.hs
```diff
@@ -85,23 +97,24 @@ performFBAction action = withManager $ \manager -> FB.runFacebookT credentials m
...
-performPost :: String -> IO Response
+performPost :: BS.ByteString -> IO Response
 performPost code = 
-	performFBAction $ do
-        postToFB code =<< getToken testPostUrl code
+    performFBAction $ do
+        token <- getToken testPostUrl code
+        user <- FB.getUser "me" [] (Just token)
+        let id_ = FB.userId user
+        liftIO $ createPNGFile ("INSERT_ID" ++ "-graph.png")
```
david-yz-liu: In the above file, you hard-coded this. I'd be happy doing the same here. (For now, of course.)

hs/CourseographyFacebook.hs
```diff
@@ -85,23 +97,24 @@ performFBAction action = withManager $ \manager -> FB.runFacebookT credentials m
...
-postToFB :: (MonadResource m, MonadBaseControl IO m) => String -> FB.UserAccessToken -> FB.FacebookT FB.Auth m FB.Id
-postToFB code token = FB.postObject "me/feed" [args "message" "Test Post Pls Ignore"] token
+getToken :: (MonadResource m, MonadBaseControl IO m) => FB.RedirectUrl -> BS.ByteString -> FB.FacebookT FB.Auth m FB.UserAccessToken
+getToken url code = FB.getUserAccessTokenStep2 url [("code", code)]
 
 -- | Gets a user's Facebook email.
 getEmail :: String -> ServerPart Response
-getEmail code = liftIO $ retrieveFBData code
\ No newline at end of file
+getEmail code = liftIO $ retrieveFBData (BS.pack code)
```
david-yz-liu: Change argument to Bytestring?

hs/ConvertSVGToPNG.hs
Ian-Stewart-Binks: The library has changed since I first installed it. My version only requires 2, the newest version requires 3. Are we keeping up to date with System.Process?

hs/CourseographyFacebook.hs
Ian-Stewart-Binks: `T.++` does not exist. Why would we use a Text here? `parseUrl` takes in a String.

hs/CourseographyFacebook.hs
Ian-Stewart-Binks: How so? `code` needs to be a bytestring. I don't believe that there is a better way to do this considering code as a Text, as it will eventually need to be converted to BS.

There is a function, `lookBS`, but I have not been able to get it to work as it takes in a different ByteString (lazy?).

hs/CourseographyFacebook.hs
Ian-Stewart-Binks: I'm not sure what line you're referring to.

hs/CourseographyFacebook.hs
Ian-Stewart-Binks: Should I just remove the `++`?

hs/ConvertSVGToPNG.hs
david-yz-liu: I think we should generally keep as up to date as possible, until we use cabal to specify certain versions of dependencies.

hs/CourseographyFacebook.hs
david-yz-liu: Hmm okay...

hs/CourseographyFacebook.hs
david-yz-liu: Yes.

hs/CourseographyFacebook.hs
david-yz-liu: That's weird, but never mind.

hs/ConvertSVGToPNG.hs
Ian-Stewart-Binks: Done. I think I now have *both* package versions installed...

hs/CourseographyFacebook.hs
Ian-Stewart-Binks: Done.

hs/ConvertSVGToPNG.hs
david-yz-liu: Ouch, yeah that's annoying. You should probably unregister the other version...

Ian-Stewart-Binks: Unregistering the other version broke *a lot* of packages... Will take a while to recover.

Ian-Stewart-Binks:
```
server.h:1:33:
    Not in scope: `main'
    Perhaps you meant `min' (imported from Prelude)
```
What should I do.

Ian-Stewart-Binks:
I think I've made all necessary changes.

david-yz-liu:
@Ian-Stewart-Binks I realize that for the ByteString issues I was pointing out, the best way to deal with them is to define a helper function that encapsulates the `look "code"` in the happstack side, but also converts it into a ByteString.

Something like ```getCode = look "code" `fmap` BS.pack``` might work?

PR #322, opened 2015-02-02
Post layout

christinem:
This pull requests edits the layout of Check My POSt (based on everyone's suggestions from last week). Now, you can only see the course codes initially, and if you click on the course code, you will get an expanded box under it showing the full names for the courses. I've also split it up into categories like "First Year", "Second Year", etc. It also adds in text input fields for the "Choose X FCEs from this list" categories when you click on the relevant category. Visually it could look better which is what I am still working on. 

This PR is built on #317. That one should be merged before this is considered.


js/post/change_div.js
```diff
@@ -34,6 +35,11 @@ $('#minor').click (function (e) {
     $('#minor').css('background-color', '#9C9C9C');
 });
 
+$(".code").click (function (e) {
```
Ian-Stewart-Binks: `"`

christinem: Fixed.

hs/CssGen.hs
```diff
@@ -650,4 +652,109 @@ fceCountCSS = "#FCECountDiv" ? do
...
+                "transition" -: "all 0.2s"
+                ":hover" & do
+                    "background-color" -: "#9C9C9C !important"
+                    a ? do
+                        "color" -: "white !important" 
+                a ? do
+                    color black
+                    display inlineBlock
+                    lineHeight (px 50)
+                    "padding" -: "0 24px"
```
david-yz-liu: Use the build-in padding function here

david-yz-liu:
@christinem Looking good! One additional comment: it would be nice if one of the three programs (e.g., specialist) appeared by default when the page is loaded.

For your main task this week, you should be able to use javascript to access the cookies to figure out which courses have been taken, and update the checklist appropriately. Looking at js/graph/tabs/post.js will be helpful. 

You may need a checkbox to actually display which courses are selected, or some other mechanism.

christinem:
@david-yz-liu Thanks! That helps me alot. 

I've also made the modifications you mentioned from both pull requests in here (except for the having a program appear on page load which I will add in the next PR).

PR #324, opened 2015-02-02
Drawing

jmatholia:
Moved some style related code from js to css. Created and added functionality to the modification mode, movement of nodes now allowed. Files modified CssGen.hs, DrawResponse.hs and experiment.js.

### DIFF
--- a/js/experiment.js
+++ b/js/experiment.js
@@ -1,14 +1,65 @@
-function drawRect() {
-	var c = document.getElementById("theCanvas");
-	var ctx = c.getContext("2d");
-	ctx.fillStyle = "#FF00FF";
-	ctx.fillRect(0, 0, 80, 100);
+// makes grid
+var nodeId = 0;
+var nodeColour = '#FF8800'
+var nodes = []
+//var mode = 'node'
+
+function drawRect(x, y) {
+	'use-strict';
+
+	var c = document.getElementById('theCanvas');
+	var ctx = c.getContext('2d');
+	ctx.fillStyle = '#FF8800';
+	ctx.fillRect(x, y, 80, 80);
 }
 
-$( "#target" ).click(function() {
-  alert( "Handler for .click() called." );
+var xmlns = 'http://www.w3.org/2000/svg';
+$( '#target' ).click(function() {
+  alert( 'Handler for .click() called.' );
 });
 
-$( "#theCanvas" ).click(function() {
-  drawRect();
-});
+function setupSVGCanvas(){
+	var svg = document.createElementNS(xmlns, 'svg');
+	svg.setAttribute('id', 'mySVG');
+	svg.setAttribute('style', 'border: 2px solid black');
+	svg.setAttribute('width', '800');
+	svg.setAttribute('height', '700');
+	svg.setAttributeNS('http://www.w3.org/2000/xmlns/', 'xmlns:xlink', 'http://www.w3.org/1999/xlink');
+	document.body.appendChild(svg);
+}
+
+setupSVGCanvas();
+document.getElementById('mySVG').addEventListener('click', getClickPosition, false);
+
+function makeNode(x, y){
+	var node = document.createElementNS(xmlns, 'rect');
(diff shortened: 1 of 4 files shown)


js/experiment.js
```diff
@@ -1,27 +1,34 @@
...
 
 function setupSVGCanvas() {
     'use-strict';
 
     var svg = document.createElementNS(xmlns, 'svg');
     svg.setAttribute('id', 'mySVG');
-    svg.setAttribute('style', 'border: 2px solid black');
-    svg.setAttribute('width', '800');
-    svg.setAttribute('height', '500');
+    svg.setAttribute('style', 'border: 2px solid black');   // also should go in CSS?
```
david-yz-liu: Yes, all node styling should be done in the CSS

js/experiment.js
```diff
@@ -49,60 +54,131 @@ function getPosition(elem) {
...
 }
 
+
+function nodeMoved(e) {
+    'use-strict';
+
+    if (mode === 'change-mode' && nodeSelected !== null) {
+        nodeSelected = this;
+        var position = getClickPosition(e);
+        var rectX = e.currentTarget.getAttribute('x')*1;
```
david-yz-liu: Better to use `parseFloat` here

js/experiment.js
```diff
@@ -49,60 +54,131 @@ function getPosition(elem) {
...
+
 function changeMode(id) {
-    //if (mode !== '') {
-    //  oldMode = document.getElementById(mode);
-    //  oldMode.setAttribute('border', 'border: 2px solid #008080');
-    //}
+    'use-strict';
+
+    if (mode !== '') {
+      $('#'+mode).toggleClass('clicked');
```
Ian-Stewart-Binks: Space around `+`, same for below.

js/experiment.js
```diff
@@ -49,60 +54,131 @@ function getPosition(elem) {
...
+    }
     mode = id;
-    //newMode = document.getElementById(mode);
-    //newMode.setAttribute('border', 'border: 2px solid white');
+    newMode = document.getElementById(mode);
```
Ian-Stewart-Binks: I don't see `newMode` as a global variable, nor used anywhere else in this file. What is its purpose?

js/experiment.js
```diff
@@ -49,60 +54,131 @@ function getPosition(elem) {
...
+    'use-strict';
+
+    oldColour = document.getElementById(nodeColourId);      // clicked class
+    oldColour.style.opacity = 1;
+    nodeColourId = id;
     nodeColour = colours[id];
-    //newMode = document.getElementById(mode);
-    //newMode.setAttribute('border', 'border: 2px solid white');
+    newColour = document.getElementById(nodeColourId);
+    newColour.style.opacity = 0.5;
```
Ian-Stewart-Binks: Where is `newColour` declared?

js/experiment.js
```diff
@@ -49,60 +54,131 @@ function getPosition(elem) {
...
-    changeColour('blue');});
-$('#purple').click(function () {
-    changeColour('purple');});
-
 
 setupSVGCanvas();
-document.getElementById('mySVG').addEventListener('click', getClickPosition, false);
\ No newline at end of file
+
+$( '.mode' ).each(function( index ) {
```
Ian-Stewart-Binks: The spacing here looks unfamiliar.

david-yz-liu:
@jmatholia Looking good, although the Firefox problem is likely caused by you using some function for checking for the mouse positions. The styling still needs to be moved over to CSS. The reason the CSS isn't having an impact is because in the actual graph, each rect (and text) is enclosed inside a <g> element, and this element is what actually has class .node, not the rect itself. You should make the same change in the drawing tool.

One thing I forgot in the meeting today: users should somehow be able to add text children to the nodes. The easiest way to do this would be to have a separate form with text input on the left side, and have the text child of a node update to that text value when you click a button or something.

And I'm looking forward to seeing the paths! I would prioritize both of these two over booleans and hybrids.

Issue #319, opened 2015-02-02
The "x" in the left sidebar of the grid does not register as clickable


christinem:
It's small, but I think it'd be nice, whenever you hover over either the "x" or the "i" images on the left sidebar of the grid (where you can either remove a course from the list or see the info modal), for it to register as clickable, as when you hover over an "a" element. Its just a bit disorienting when you hover over it right now - you believe the "x" has no function until you actually click it.

Ian-Stewart-Binks:
I completely agree.

In fact, I don't like that red X.

christinem:
I can actually do this. 

As for whether we want to keep the X or not, that'd be interesting conversation to have.

david-yz-liu:
Sounds good. Adding the mouse pointer should be fine for now. @Ian-Stewart-Binks, you are free to suggest an alternate icon, or something else entirely.

david-yz-liu:
@Ian-Stewart-Binks While I'm closing this issue, if you find another icon to use, great.

Ian-Stewart-Binks:
I will keep my eyes peeled.

PR #326, opened 2015-02-05
Issue 319: Added Cursor Pointer

christinem:
This fixes issue #319. Now when you hover over the delete or about icon on the grid on the left sidebar, it shows a cursor pointer.

### DIFF
--- a/hs/CssGen.hs
+++ b/hs/CssGen.hs
@@ -536,6 +536,9 @@ courseSelectCSS = do
             height (px 20)
             padding (px 1) 0 (px 1) 0
             cursor pointer
+            opacity 0.8
+            ":hover" & do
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@christinem I have an idea I wonder if you could try out: make the icons opacity 90% by default, and make them 100% opacity on hover. This might make them "pop" a bit more. Could you take a look and see if you like it?

christinem:
@david-yz-liu I like it! It was bugging me a bit that you still could not see a difference if you were hovering over one icon or the other, now it definitely shows. Although I think 80% opacity by default might be better because you don't notice it too much when it goes from 90% to 100%.

david-yz-liu:
@christinem Great!

Issue #258, opened 2015-01-12
Text opacity in happstack


Ian-Stewart-Binks:
Unlike in the current version of Courseography, the opacity of the nodes in the graph of the happstack branch do not change when a course's state changes (e.g. `takeable -> inactive`).

Should we change this back to the behaviour on the current (master) version of Courseography?

christinem:
Did you guys still want this changed back to how it was on master? If so, I can take this on.

david-yz-liu:
Sure, go for it!

PR #328, opened 2015-02-08
Issue 258 Text Opacity Fix

christinem:
This fixes Issue #258. Now the opacity of text on nodes matches the live (master) version of courseography.


Ian-Stewart-Binks:
@christinem Looks nice!

PR #330, opened 2015-02-08
SVG Parser: Lots of progress

Ian-Stewart-Binks:
This pull request adds more parsing functionality to the parser. In general, `rect`, `text` and `path` elements are now parsed. 

The parser builds up elements recursively, looking at the current node and then recursing on the children. It builds up things such as `transform` in a way that mimics CSS inheritance. For instance, at each recursive step it accumulates transform to pass onto the children nodes, and to apply at any elements. Similarly, it currently does this with `fill`. This is how the current parser can match colours to `rect`s, even though `g` elements have the colours in the native graph.

The parser inserts this information into the database, and then retrieves it to rebuild the SVG.

There are many things that still need to be worked on, such as absolute paths, regions, more style, etc.


hs/database/Tables.hs
```diff
@@ -77,25 +77,26 @@ Graphs
 
 Rects
     gId Int
-    rId Int
+    rId String
```
Ian-Stewart-Binks: This has been changed to a String for the time being. In the future, I hope to use the Persistent database primary keys.

hs/parser.hs
```diff
@@ -0,0 +1,218 @@
...
+           let rects = parseContent (tag "rect") content
+           let texts = parseContent (tag "text") content
+           let paths = parseContent (tag "path") content
+           let children = getChildren content
+           let transform = getAttribute "transform" content
+           let style = getAttribute "style" content
+           let fill = getFill style
+           let fillx = if null fill then parentFill else fill
+           let filly = if fillx == "none" then parentFill else fillx
+           let fillz = if fillx == "#000000" then "none" else filly
```
Ian-Stewart-Binks: This section should really be improved.

PR #337, opened 2015-02-09
Drawing

jmatholia:
In experiment.js: Allowed selction of nodes and addition of text to nodes through an input field. Functionality for creating and deleting paths. Explicit relationships between nodes and paths. Movement of text. 
In CssGen.hs and DrawResponse.hs: Showing selction of nodes and added the input field and buttons(Add and Path) and style for paths. All css for paths and nodes now connected to the graph_styles.css, some additions in that section as well.

### DIFF
--- a/js/experiment.js
+++ b/js/experiment.js
@@ -1,14 +1,65 @@
-function drawRect() {
-	var c = document.getElementById("theCanvas");
-	var ctx = c.getContext("2d");
-	ctx.fillStyle = "#FF00FF";
-	ctx.fillRect(0, 0, 80, 100);
+// makes grid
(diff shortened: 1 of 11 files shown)


js/experiment.js
```diff
@@ -56,60 +91,150 @@ function getPosition(elem) {
...
     }
 }
 
 
 function nodeClicked(e) {
-    'use-strict';
-
-    if (mode  === 'erase-mode') {
-        document.getElementById('mySVG').removeChild(e.currentTarget);
+    'use-stri`ct';
```
Ian-Stewart-Binks: `

Issue #341, opened 2015-02-10
Refactor CssGen into different files.


david-yz-liu:
Full disclosure: I (David) wrote CssGen.hs in a hurry, more as a proof of concept. Since we seem to be sticking to it, it would be good to refactor it a little.

@pkukulak You can start by reorganizing the file structure.

- There should be one file containing all the constants we've defined. Please include comments.
- The file structure should mirror the output css files: all the styles associated with an output file should be in the same file, and each output file should have its own "styleGen" file.

This is an excellent way to learn Haskell.

pkukulak:
Nice! I can totally do it. This will be a great thing to do over the break. I may come and see you to clarify a few things before I start working on it.

pkukulak:
Okay, commencing work on this as we speak. Hang tight; it may turn out to be troublesome.

PR #348, opened 2015-02-18
Issue 341 - Refactoring CssGen.hs

pkukulak:
#341. Thankfully, CssGen was written in a nice, modular fashion and so this was not nearly as difficult as I anticipated it to be. 

I have moved constants into a file name Consts.hs (might be a questionable name now that I think of it), and all functions that generate CSS for different pages into respective files: namely, the new files are AboutCssGen.hs, TimetableCssGen.hs, AboutCssGen.hs and GraphCssGen.hs.

I don't doubt that there are some rough edges to polish in these files.


hs/AboutCssGen.hs
```diff
@@ -0,0 +1,23 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module AboutCssGen where
+
+import Clay
+import Prelude hiding ((**))
+import Data.Monoid
+import Data.Text.Lazy
+import System.Directory
```
david-yz-liu: Please double-check the imports here... I'm not sure all of them are necessary.

hs/CssGen.hs
```diff
@@ -7,14 +7,18 @@ import Prelude hiding ((**))
 import Data.Monoid
 import Data.Text.Lazy
 import System.Directory
-
+import Consts
+import CommonCssGen as COMMON
```
david-yz-liu: Not big on the all-caps here. Just import the one style per module, e.g.

`import CommonCssGen (common)`

david-yz-liu:
Looks good, but let's drop the "Gen" from file names.

pkukulak:
I agree on the caps. Fixed.

Files have also been renamed.

Note that since the search symbol pull request has been merged, I made the according changes in TimetableCss.hs as well.

PR #384, opened 2015-02-25
Haskell Directory Structure

### LINKED ISSUE #377: Directory Structre: HS directory
The hs directory structure needs to be improved.

Ian-Stewart-Binks:
I have improved the directory structure of the hs directory. I still think that it can be improved further, but this is a good first step.

Fixes #377

Note that some things have been cleaned up as well, `post` is removed from server.hs, and `import SVGGen` statements have been removed.


Ian-Stewart-Binks:
@david-yz-liu Response directory has been removed.

david-yz-liu:
@Ian-Stewart-Binks I think I was wrong about getting rid of the `Response` directory. I can't recall why I wanted to do that. If you happen to remember, please let me know - I'm working on recreating that directory now.

Ian-Stewart-Binks:
@david-yz-liu I have a faint memory of it. I believe you wanted to group the responses with their corresponding models and views, or something like that. My apologies, the memory is not so clear.

david-yz-liu:
Ah you're right. I admire the intent, but some of the smaller responses seem too trivial to separate. I'll think more on this.

PR #394, opened 2015-02-27
database-path: Standardize database path

david-yz-liu:


david-yz-liu:
@christinem @Ian-Stewart-Binks Were you not having problems where the database was in the wrong location?

Ian-Stewart-Binks:
@david-yz-liu Yeap, was about to make an issue for it.

christinem:
@david-yz-liu yeah that issue still persists, where you have to run the database file from the hs directory,

Issue #396, opened 2015-02-28
Can't run parser.hs


christinem:
I think this has something to do with how the Haskell directory structure was recently changed. 

When I try to run ```runhaskell parser.hs``` from the ```SvgParsing``` directory, I get an error such as so:

```
ParserUtil.hs:1:8:
    File name does not match module name:
    Saw: ‘SvgParsing.ParserUtil’
    Expected: ‘ParserUtil’
```

When I try to run ```runhaskell SvgParsing\parser.hs``` from the ```hs``` directory, I get this error:

```
SvgParsing/parser.hs:24:8:
    Could not find module ‘SVGGenerator’
    Use -v to see a list of the files searched for.

SvgParsing/parser.hs:25:8:
    Could not find module ‘SVGBuilder’
    Use -v to see a list of the files searched for.

SvgParsing/parser.hs:26:8:
    Could not find module ‘SVGTypes’
    Use -v to see a list of the files searched for.

SvgParsing/parser.hs:27:8:
    Could not find module ‘ParserUtil’
    Use -v to see a list of the files searched for.
```

Ian-Stewart-Binks:
For the latter, I must have forgotten to add the module prefixes to the `parser.hs` file. The parser should be run from the hs directory.

christinem:
@Ian-Stewart-Binks ah yes, adding the module prefixes fixed it. So the permanent behaviour is to run it from ```hs``` now with ```runhaskell SvgParsing/parser.hs``` or will this be changed?

david-yz-liu:
Run from hs. One of you two should submit a pull request for this.

christinem:
I'll submit a PR for this.

Ian-Stewart-Binks:
Thanks Christine.

PR #397, opened 2015-02-28
Parser Fix for Imports

christinem:
This fixes Issue #396. The parser can now be run properly from the ```hs``` directory.


Ian-Stewart-Binks:
Looks good. Thanks @christinem !

Issue #413, opened 2015-03-03
Remove Old Post Code


Ian-Stewart-Binks:
Out with the old (`js/graph/tabs/post.js`) in with the new (`js/post/...`).

@christinem Does Courseography currently use any of the old post code?

christinem:
@Ian-Stewart-Binks nope, not at all!  It can be removed. In fact I can remove it  :)

PR #416, opened 2015-03-04
Remove post.js from Scripts.hs

christinem:
This also fixes Issue #413.


Ian-Stewart-Binks:
There might be some global variables that no longer need to live.

christinem:
Alright, I'll scour around the code and look for any.

christinem:
I think I've now caught all the now-unneeded variables and references to functions in post.js.

Ian-Stewart-Binks:
Good job on finding the mass of global variables in setup.js.

PR #419, opened 2015-03-05
Fix Cookies

Ian-Stewart-Binks:
This pull request removes the comma from Hybrid IDs. The IDs are used as cookie names, and commas are illegal characters in cookies. ([Source](http://curl.haxx.se/rfc/cookie_spec.html))

This is currently preventing `rqCookies` from Happstack to work properly, as Happstack cannot parse our cookies.

To see a cookie with a comma in its name, make sure that the cookie corresponding to the hybrid of CSC404 is being set.


david-yz-liu:
@Ian-Stewart-Binks Nice!!

PR #420, opened 2015-03-06
Image Conversion

Ian-Stewart-Binks:
This pull request allows images to be created by requesting `image` or `timetable-image`. The latter must be accompanied by an appropriate query string.

This pull request is built on top of #419, as the function `rqCookies` requires there to be no commas in the cookies.

Note that [Happstack appears to convert cookies to lowercase](http://www.happstack.com/docs/happstack-server-7.0.2/doc/html/happstack-server/src/Happstack-Server-Internal-Cookie.html#Cookie). As such, we now use only lowercase cookies for cookies that we use from `rqCookies`.


hs/ImageResponse.hs
```diff
@@ -0,0 +1,48 @@
...
+timetableImageResponse :: String -> ServerPart Response
+timetableImageResponse courses = liftIO $ getTimetableImage courses
+
+-- | Creates an image, and returns the base64 representation of that image.
+getGraphImage :: M.Map String String -> IO Response
+getGraphImage courseMap = do
+	buildSVG courseMap "Testfile2.svg"
+	liftIO $ createImageFile "Testfile2.svg" "INSERT_ID-graph.png"
+	imageData <- BS.readFile "INSERT_ID-graph.png"
+	liftIO $ removeImage "INSERT_ID-graph.png"
```
david-yz-liu: Can you also delete Testfile2.svg?

hs/MasterTemplate.hs
```diff
@@ -23,15 +23,16 @@ masterTemplate title headers body scripts =
...
-  H.h2 "Courseography"
-  H.ul ! A.id "nav-links" $ do
-    H.li $ makeA "" "" "graph" "" $ "Graph"
-    H.li $ makeA "" "" "grid" "" $ "Grid"
-    H.li $ makeA "" "" "draw" "" $ "Draw"
-    H.li $ makeA "" "" "post" "" $ "Check My POSt!"
-    H.li $ makeA "" "" "about" "" $ "About"
-
+header page = 
+  createTag H.nav "" "row header" $ do
```
david-yz-liu: This function is off.

1. Make the indentation 4 spaces.
2. What didn't you remove the `$` for "draw"?

Ian-Stewart-Binks: 1. Sure thing.
2. `$` came from an error in my merge of your `timetable-svg`.
3. The last line should also be removed.

hs/ImageResponse.hs
Ian-Stewart-Binks: Done. Does GitHubs 'outdated diff' feature only work if the exact line has been changed?

david-yz-liu: Yeah, but that's fine.

PR #432, opened 2015-03-09
lWeb request

shaftoes:
Added support for half hour time slots


david-yz-liu:
@shaftoes Looking good. I'm merging this in, though you'll be expected to do significant refactoring before the term is done. I'm very happy that the functionality seems to be complete at this point.

PR #433, opened 2015-03-09
Four oh four

jmatholia:
Created four oh four page and added a 404 page as a default in server.hs if no other links match. Design could use some work still. Fixes issue#411.


david-yz-liu:
1. Don't use `masterTemplate` (i.e., don't show the nav bar)
2. nice dragon!

david-yz-liu:
@jmatholia If you remove the picture, I'll merge this in. Thanks!

PR #438, opened 2015-03-10
Svg fonts

david-yz-liu:
@Ian-Stewart-Binks @pkukulak I looked into the font issue for how we're generating SVG. In the long run, we'll probably want to change our approach anyway, as the jpegs generated through ImageMagick don't seem to be that good. But I think you should both move on from this for now, at least until everything else works perfectly.


Ian-Stewart-Binks:
@david-yz-liu The width/height of the SVG and the size of the image may be a bit off. I noticed that I got clearer pictures when I increased the width of the output SVG.

david-yz-liu:
Cool. Please feel free to change the numbers around, I just chose them arbitrarily.

PR #442, opened 2015-03-11
Post Lowercase Cookie Fix

christinem:
This fixes the issue introduced now that cookies get stored in lowercase instead of uppercase. Also fixes a little bug I found when trying to autofill 200 level courses in the minor.

### DIFF
--- a/js/post/update_post.js
+++ b/js/post/update_post.js
@@ -182,8 +182,8 @@ function addExtraMinCourses() {
     var textboxesLeft = minor.textboxesExtra - minor.filledTextboxesExtra;
 
     for (var m = 0; m < Math.min(textboxesLeft, 3); m++) {
-        if (getCookie(minor.additionalMin200[m]) === 'active' || 
-            getCookie(minor.additionalMin200[m]) === 'overriden') {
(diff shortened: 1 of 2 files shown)


js/post/update_post.js
```diff
@@ -259,8 +259,8 @@ function updateActiveCourses() {
     for (var areaName in areas) {
         if (areas.hasOwnProperty(areaName)) {
             for (var i = 0; i < areas[areaName].length; i++) {
-                if ((getCookie(areas[areaName][i]) === 'active' || getCookie(areas[areaName][i]) === 'overridden') && 
-                    areas[areaName][i] !== 'CSC104') {
+                if ((getCookie(areas[areaName][i].toLowerCase()) === 'active' || getCookie(areas[areaName][i].toLowerCase()) === 'overridden') && 
+                    areas[areaName][i].toLowerCase() !== 'CSC104') {
```
Ian-Stewart-Binks: 'CSC104' is still in capitals.

christinem: Woops, you're right. Fixed it.

PR #448, opened 2015-03-13
Focuses 

christinem:
This PR fixes the bugs from last week - now courses on the graph are clickable again after a focus is closed, and the styling for the highlights (ellipses) are dealt with properly in the CSS. You can now click anywhere in the graph as well to close the sidebar. You can close a focus by clicking the button for the focus again, instead of the "Close Focus" button from before. 

I have also added focus descriptions in - when you click on a focuses in the sidebar, a div expands showing a scrollable view with the description and courses necessary for the focus. In the live version, these descriptions were hard coded in, so I decided to do the same for now, in a separate file. I'm not sure how good it looks right now, or if it seems too cramped in the sidebar - if you are able to, please try it out and let me know!


hs/Css/GraphCss.hs
```diff
@@ -193,7 +189,7 @@ sidebarCSS = do
         width (px 20)
         height (pct 100)
         float floatLeft
-        "background" -: "rgba(128, 0, 128, 0.8)"
+        "background" -: "rgb(128, 0, 128)"
```
david-yz-liu: make a colour constant

js/graph/sidebar/sidebar_divs.js
```diff
@@ -0,0 +1,85 @@
+var toggled = false;
+
+
+$("#sidebar-button").click(function(){
```
david-yz-liu: space before `{`

js/graph/sidebar/sidebar_divs.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+
+/**
+ * Fills the count of FCEs in the sidebar.
+**/
+function fillFCECount() {
+    'use strict';
+    
+    $("#fcecount").show();
```
david-yz-liu: single-quotes

david-yz-liu:
@christinem Don't set styles using `css` - use classes instead.

christinem:
@david-yz-liu I'm not sure I know exactly what you mean. The styles are statically set using classes at first. The only time I set styles using `css` is when I change the active tab - so the selected tab in the nav bar changes colour based on which one is active (whichever one is selected is highlighted purple). Is there a better way to do this using just classes?

david-yz-liu:
Yes, you should give the "active" tab a class rather than change the css directly. Always.

christinem:
@david-yz-liu Ah I see, understood! I've made the fixes.

Issue #423, opened 2015-03-06
Check My POSt: Changing header style


Ian-Stewart-Binks:
In the header, the 'Courseography' title appears differently in 'Check My POSt' than it does from /graph or /grid.

christinem:
Mmm you're right. Must be a conflict with stylesheets. I'll check it out.

PR #449, opened 2015-03-13
Issue 423 Header Style Fix

christinem:
This fixes Issue #423.


david-yz-liu:
@christinem Why isn't the code width just 100%?

christinem:
@david-yz-liu I actually tried 100% at first, but for some reason it extended too far and did not match the width of the full-name divs, but 99% seemed to make it fit. I might play around with the styling of them.

Issue #400, opened 2015-03-01
Remove Video JS


Ian-Stewart-Binks:
res/videojs should be removed, and all links to Video JS should point to the one hosted at http://www.videojs.com/.

christinem:
I can take this on.

PR #450, opened 2015-03-13
Issue #400 Removing Video JS

christinem:
This fixes Issue #400.

Out of curiosity, is there any reason there was no link to the javascript file video.js on master, compared to the live version? Should I add the link to the remotely hosted video.js back into master?


david-yz-liu:
@christinem Sure, you can add it back in. We'll eventually use it.

PR #453, opened 2015-03-16
Svg refactor

Ian-Stewart-Binks:
This pull request refactors the Haskell code by implementing a better parsing algorithm, cleaning out the database before reinserting a graph, cleaning up the haskell code, setting up support for multiple graphs, and more.

Note that the database must be rebuilt once this pull request is merged in.


hs/ImageConversion.hs
```diff
@@ -26,7 +26,6 @@ convertToImage inName outName = createProcess $ CreateProcess
                                   CreatePipe
                                   False
                                   False
-                                  False
```
david-yz-liu: Keep these

david-yz-liu:
@Ian-Stewart-Binks The only thing I'm unsure of is the new DataType file - why not just put that in Tables.hs?

Ian-Stewart-Binks:
@david-yz-liu I was having trouble declaring it in Tables.hs. I was getting a "ghc stage restriction" error, and importing the datatype solved the problem. I couldn't find a way to do it otherwise (even though we do it with `Time`).

david-yz-liu:
@Ian-Stewart-Binks Ah, got it. Do you really need all of the pragmas for that file, then?

Ian-Stewart-Binks:
Nope, I've taken them out.

PR #460, opened 2015-03-17
Graphs response

Ian-Stewart-Binks:
This pull request adds new functionality that allows the client to request graph title-id pairs in JSON format. This information will be used by the graph switching functionality to display graphs in the database on the sidebar.

The format looks like this: `[{"graph_title":"CSC","graph_id":1}]`.

This is a stepping stone for graph switching functionality.

@christinem This code can be used to get graph title/id pairs. It is of course, not yet complete, as I intend to use primary keys from the database to identify graphs. Currently, only CSC and 1 can be inserted into the graph, but this should soon change. If you wish to try on more graphs, you can change these two numbers and 'reinsert' the same graph.

### DIFF
--- a/hs/Database/CourseQueries.hs
+++ b/hs/Database/CourseQueries.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE ScopedTypeVariables, OverloadedStrings #-}
 
-module Database.CourseQueries (retrieveCourse, allCourses) where
+module Database.CourseQueries (retrieveCourse, allCourses, queryGraphs) where
 
 import Database.Persist
 import Database.Persist.Sqlite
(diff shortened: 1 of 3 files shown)


hs/Database/CourseQueries.hs
```diff
@@ -93,3 +93,10 @@ allCourses = do
       return $ T.unlines codes
   return $ toResponse response
 
+-- | Queries the graphs table and returns a JSON response of Graph JSON
+-- objects.
+queryGraphs :: IO Response
+queryGraphs =
+    runSqlite dbStr $
+        do graphs :: [Entity Graph] <- selectList [] []
+           return $ toResponse $ createJSONResponse $ encodeJSON $ Aeson.toJSON $ map (Aeson.toJSON . entityVal) graphs
```
david-yz-liu: Doesn't `createJSONResponse` already return a `Response`?

Ian-Stewart-Binks: Yeap. Fixed. There was also another place where that occurred...

christinem:
@Ian-Stewart-Binks Thanks. So right now the database still only stores one graph and ID? And where would the JSON end up after it's created?

david-yz-liu:
@christinem The JSON is received by the client (browser) as the response.

PR #463, opened 2015-03-19
Timetable - Update style and tutorials

foreverwander:
1. Created new dummy cell class that has Css set in the hs file

2. Rendered with Firefox display problems, now use % instead of px

3. Fixed bug relating to font-size

4. Now tutorials times are parsed correctly with '-'

5. Tutorials are now able to show up in timetable. Discovered UNDERLYING problem. The attribute of courses 'manualTutorialInrollment' is likely not parsed correctly from the internet. Now just shows as null for all the courses that i have checked, with or without tutorials. Please see attached for evidence.

Old:
old

New:
new
new2

### DIFF
--- a/hs/Css/TimetableCss.hs
+++ b/hs/Css/TimetableCss.hs
@@ -118,6 +118,8 @@ timetableCSS = do
             fontSize (em 0.9)
         ".timetable-time" ? do
             width (pct 12)
+        ".timetable-dummy-cell" ? do
+            width (pct 1)
     -- Overriding bootstrap
     ".col-md-2" ? do
         width (pct 14)
--- a/hs/Database/Tables.hs
+++ b/hs/Database/Tables.hs
@@ -228,9 +228,9 @@ instance FromJSON Tutorial where
 
 instance ToJSON Tutorial where
   toJSON (Tutorial Nothing times timeStr) =
-      Array $ V.fromList [toJSON times, toJSON timeStr]
+      Array $ V.fromList [toJSON (map convertTimeToString times), toJSON timeStr]
   toJSON (Tutorial (Just value) times timeStr) =
-      Array $ V.fromList [toJSON value, toJSON times, toJSON timeStr]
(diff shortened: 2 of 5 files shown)


david-yz-liu:
@foreverwander Looking good... though I would prefer it if we fixed the underlying problem and made the manualTutorialEnrolment field either true or false again... if you have time to look into it, that would be great.

foreverwander:
@david-yz-liu I would love to fix the underlying problem. However, the thing is that I can't even find the 'manual tutorial enrolment' indicator on the timetable or the calendars online.... Do you know when that data came from before?

david-yz-liu:
@foreverwander It's a computed indicator. If a course has tutorial or lab sections (e.g., "T0101" or "L0101") then we set it to True, otherwise it's false.

david-yz-liu:
@foreverwander I liked everything except the font-size change, which I took out. First, you need to stop using .css in your javascript but instead actually use css. But I also don't buy that this is a good fix; depending on the screen size this makes the text look too small.

foreverwander:
@david-yz-liu  sorry the font-size change was not intentional. It is because I set the rowspan 1 cells to have font-size of 0 to hide the text, so I have to set those back in order for the text to show again when the cells return to rowspan 2. I thought that I had used the same value as in the TableCss but I guess not.... Sorry again and I'll try to clean up the code to actually use classes and css.

david-yz-liu:
@foreverwander A better approach is just to not put any text in those cells in the first place. I suppose the font-size could have been 0.9em in the CSS as well, though.

foreverwander:
@david-yz-liu  I have tried the other approach as well. The only problem would be that the text in the table cell is used in various occasions, such as to check for whether the cell is selected, and or whether the cell is experiencing a conflict. If the text is removed then essentially the indicator that the cell is representing the course is lost. This would change the structure of functions quite a bit.

david-yz-liu:
Ah yes. Well, I guess the thing you can do is use an attribute CSS selector to match the tds that have rowspan="1" and rowspan="2".

PR #464, opened 2015-03-19
Facebook Modal Refactoring

Ian-Stewart-Binks:
This pull request refactors the facebook modal code and the course info modal code. The modal opening code is now a separate function that takes in a 'title' (what appears at the top of the modal) and a div (the contents of the modal).

This pull request also fixes a bug in the modal code, where the screen would partially black out if it was opened in the grid, and it no longer tries to change the style of the nodes when in the grid.


public/js/common/modal.js
```diff
@@ -218,7 +218,7 @@ function createG(nodeId) {
     g.attr('class', 'tooltip-group')
         .css('cursor', 'pointer')
         .click(function () {
-            openModal(nodeId);
+            openCourseModal(nodeId);
```
david-yz-liu: Eh, I don't think this needs to be a helper. The facebook one just does it inline, which is fine here too.

Ian-Stewart-Binks: Fixed.

PR #465, opened 2015-03-19
Graph IDs

Ian-Stewart-Binks:
This pull request adds the initial functionality for graph identification. Graph IDs will be based on primary keys. The code no longer deletes graphs when parsing a graph that is already in the database.

The primary key is inserted into the `Graph` table under the `graphsGId` attribute because, to my knowledge, the primary key cannot be retrieved from the Graph object.

For example, the following code needs access to the primary key:
```
instance ToJSON Graph where
    toJSON (Graph id_ title)
        = object ["graph_title" .= title,
                  "graph_id" .= id_
                 ]
```

Future pull requests will insert this key into the database for each corresponding element of the graph.

### DIFF
--- a/hs/Database/Tables.hs
+++ b/hs/Database/Tables.hs
@@ -69,19 +69,19 @@ Distribution
     deriving Show
 
 Graph
-    gId Int
+    gId String
     title String
     deriving Show
 
 Text
-    gId Int
+    gId String
     rId String
     pos Point
     text String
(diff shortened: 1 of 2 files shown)


Ian-Stewart-Binks:
This PR has been updated. It now uses `Int64`.

david-yz-liu:
Thanks, Ian.

PR #471, opened 2015-03-23
Added a searchable timetable

david-yz-liu:


david-yz-liu:
@Ian-Stewart-Binks Since you implemented the original version, if you have any thoughts about this, please let me know.

Ian-Stewart-Binks:
I have a question, but it has very little to do with the original implementation: Why are we creating the table in React rather than creating a response on the server? When is it preferable to create HTML client side vs. server side?

david-yz-liu:
If you're thinking about the initial load, it's certainly true that we would save time by saving an html file, like we used to. In general, the point is to use React to simplify the rendering implementation. Ideally, we won't load any timetable info on page load, but let the user "search" by department.

I think server-side vs. client-side rendering is a bigger and more complex question, but I don't think it matters too much for us right now.

Ian-Stewart-Binks:
I'm referring to using blaze-html rather than a static html file, wouldn't the rendering implementation be the very similar? Couldn't we load timetable information into the page by creating table responses in blaze-html?

david-yz-liu:
That would be fine for the initial load, but then filtering would be harder.

PR #473, opened 2015-03-24
populated manualTutorialEnrolment field in database and course records

shaftoes:


hs/WebParsing/TimeTableParser.hs
```diff
@@ -204,8 +205,12 @@ processCourseTable course = do
   print code
   runSqlite dbStr $ do
     runMigration migrateAll
+    setTutEnrol code (containsTut sesh)
+    --setPracEnrol code (containsPractical sesh)
     mapM_ (\l ->  insertLec session code l) (lectures sesh)
-    mapM_ (\t ->  insertTut session code t) (tutorials sesh)    
+    mapM_ (\t ->  insertTut session code t) (tutorials sesh)
```
Ian-Stewart-Binks: Do you need to use the lambda here?

david-yz-liu:
@shaftoes Thanks for doing this so quickly. I hope you'll be able to add in a manualPracticalEnrolment field as well (though it's less of a priority right now). Please be aware that we try to avoid trailing whitespace at the ends of lines.

@Ian-Stewart-Binks Thanks for your comment!

@foreverwander This now means manualTutorialEnrolment should always either be true or false, so you can merge master into half-hour and make the necessary updates!

foreverwander:
@shaftoes @david-yz-liu Sorry to say this but now it might be very necessary to add in a manualPracticalEnrolment. Because for now if a class has practicals but no tutorials, nothing will work... as evident below:

PR #474, opened 2015-03-24
graphs-json: use native ToJSON for graphs response.

Ian-Stewart-Binks:
This pull request takes out the redundant ToJSON behaviour redefinition.

@christinem the json format now looks like this:
```
[{"gId":1,"title":"CSC"},{"gId":2,"title":"CSC"}]
```
The names have been switched from
graph_title -> title
graph_id -> gId


hs/Database/Tables.hs
```diff
@@ -241,10 +241,4 @@ instance ToJSON Tutorial where
 convertTimeToString :: [Double] -> [T.Text]
 convertTimeToString [day, time] =
   [T.pack . show . floor $ day,
-   T.replace "." "-" . T.pack . show $ time]
```
david-yz-liu: It would be good to preserve the last newline, if it went missing

christinem:
@Ian-Stewart-Binks thanks! Once this is merged I'll make the necessary changes in my branch.

hs/Database/Tables.hs
Ian-Stewart-Binks: Added.

Issue #190, opened 2014-11-15
Transfer JSON to SQL


Ian-Stewart-Binks:
Transfer JSON to SQL, specifically postgres.

david-yz-liu:
Basically done now, due to @shaftoes

Ian-Stewart-Binks:
Can we expunge the (majority) of the web_parsing code?

Do we want to make an issue for converting to postgres in the future?

david-yz-liu:
Yeah, let's get rid of that directory entirely.

Let's hold off on switching databases for now... I think we need to do some more research.

Ian-Stewart-Binks:
Should we save enrol_parser?

david-yz-liu:
No, we can rewrite in Haskell this summer

PR #476, opened 2015-03-24
rm-utilities: Remove utilities directory.

Ian-Stewart-Binks:
Remove the utilities directory as decided in #190.


david-yz-liu:
Thanks, @Ian-Stewart-Binks

Ian-Stewart-Binks:
@david-yz-liu Not a problem. Should I remove the old code that would parse the JSON files? (Such as the FromJSON implementations in Tables.hs)?

david-yz-liu:
Hmmm, we want the FromJSON and ToJSON implementations to match up. If there was a specific FromJSON implementation that was specifically for the old JSON files, and this doesn't match the ToJSON, then it would be best to change it.

Ian-Stewart-Binks:
There's quite a bit of code that was used to parse the JSON files that were created from the web_parsing code. This code isn't needed any longer, which includes the FromJSON implementations.

david-yz-liu:
Okay, then feel free to remove it.

PR #485, opened 2015-03-27
Remove Post FCE Count from Graph page

christinem:
On suggestion from @david-yz-liu, this PR removes the FCE count for the active POSt from the Graph page nav bar. In the future it can be added somewhere else such as the sidebar.

David, would you like me to remove it from the nav bar on the POSt page as well?

### DIFF
--- a/public/js/post/update_post.js
+++ b/public/js/post/update_post.js
@@ -228,25 +228,6 @@ function updateNavPost() {
     } 
 }
 
-
-/**
- * Updates the Nav Bar on the Graph page.
-**/
-function updateNavGraph() {
-    'use strict';
-
-    var nav_graph = $('#nav-links')[0].getElementsByTagName('li')[3].getElementsByTagName('a')[0];
-
-    if (getCookie('specialist') === 'active') {
-        nav_graph.innerHTML = 'Check My POSt! (' + getCookie('activecount') + '/12.0)';
-    } else if (getCookie('major') === 'active') {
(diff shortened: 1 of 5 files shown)


public/js/post/update_post.js
```diff
@@ -214,7 +214,7 @@ function checkPostCompleted() {
 function updateNavPost() {
     'use strict';
 
-    var navPost = $('#nav-links')[0].getElementsByTagName('li')[3].getElementsByTagName('a')[0];
+    var navPost = $('#nav-links')[0].getElementsByTagName('li')[4].getElementsByTagName('a')[0];
```
Ian-Stewart-Binks: Is it possible to get `navPost` by adding an ID to to the proper `a` element in MasterTemplate?

public/js/post/update_post.js
```diff
@@ -228,25 +228,6 @@ function updateNavPost() {
     } 
 }
```
Ian-Stewart-Binks: It looks like there's now only one space between these two functions.

david-yz-liu:
@christinem Yes; I think the nav bar should look the same in all the tabs.

public/js/post/update_post.js
christinem: Yeah, I definitely think so. But in any case, right now this had been removed since it was decided we won't be updating the nav bar with the POSt credit count at the moment.

public/js/post/update_post.js
christinem: Done!

christinem:
@david-yz-liu Thanks, I have made the updates

PR #491, opened 2015-03-29
Graph Filename & Attribute Change

christinem:
Speaking to @Ian-Stewart-Binks , it seems that now we are going to use the IDs of the graphs as the filenames for the parsed graphs (such as 1.svg, 2.svg, etc .... ). I have adjusted the graph switching code to account for that, as well as the changed attribute names in the JSON objects.

This PR is built on #490.


david-yz-liu:
This change hasn't happened yet. I would wait.

Issue #351, opened 2015-02-20
Re-introduce Reset Button


christinem:
I think we should re-introduce the reset button on the graph (to clear all the selected courses), as is in the live version.

Ian-Stewart-Binks:
I agree. I think it's about time.

Should we keep the button on the graph, or move it over to some sort of... tool bar?

christinem:
@Ian-Stewart-Binks I'm not quite sure yet. Once I start working on the graph-switching, another toolbar or sidebar might be introduced, so perhaps it could go there? Though I don't see anything wrong with keeping it on the graph if that's not too difficult.

PR #494, opened 2015-03-30
Issue 351: Add Reset Button

christinem:
This PR fixes Issue #351 - it adds the reset button back into the graph. Right now it just sits in the graph, but once the sidebar makes it way into master, it can be moved there.


david-yz-liu:
Eh, I think you should just put it in the sidebar branch.

Issue #373, opened 2015-02-21
SVG Parser: Info Box


Ian-Stewart-Binks:
The Info Box tool tip does not display correctly in graphs generated by the SVG parser.

PR #495, opened 2015-03-30
Issue 373: Info Tooltip Fix

christinem:
This PR fixes Issue #373. Now the "Info" text displays properly inside the tooltip box.


Ian-Stewart-Binks:
The SVG equivalent of the 'z-index' also needs to be fixed for #373. This issue can be reproduced by hovering over the CSC463 node.

christinem:
@Ian-Stewart-Binks done! The tooltip should now display on top of everything else.

PR #504, opened 2015-04-01
Graph Loading with new Filenames/Attributes

christinem:
This PR changes the way graphs are grabbed and loaded, based on the new convention of files being named after their primary IDs. 

@Ian-Stewart-Binks the IDs right now are stored within the `graph-button` elements for each graph, with the key `id` - each respective ID stored in its respective button.  To note, however, the buttons are only dynamically created when you open the 'graphs' tab in the sidebar. If you need them stored in a different place, let me know, it should not be too hard, depending on how you want to use it.

### DIFF
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -21,7 +21,8 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph('1');
+    var active = getCookie('active-graph');
+    loadGraph(active);
     $('#fcecount').hide();
 });
 
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -21,7 +21,7 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph('graph-csc', 1);
+    loadGraph(1);
     $('#fcecount').hide();
 });
 
@@ -35,7 +35,7 @@ function getRemote(filepath) {
 
     var SVG = $.ajax({
         type: 'GET',
-        url: 'static/res/graphs/1.svg',
+        url: filepath,
         async: false
     }).responseText;
     $('#graph').append(SVG);
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -22,7 +22,13 @@ $(document).ready(function () {
     'use strict';
 
     var active = getCookie('active-graph');
-    loadGraph(active);
+
+    if (active !== '') {
+        loadGraph(active);
+    } else {
+        loadGraph('1');
+    }
+    
     $('#fcecount').hide();
 });
 
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -21,7 +21,7 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph(1);
+    loadGraph('1');
     $('#fcecount').hide();
 });
 
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -21,7 +21,7 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph('graph-csc');
+    loadGraph('graph-csc', 1);
     $('#fcecount').hide();
 });
 
--- a/public/js/graph/setup.js
+++ b/public/js/graph/setup.js
@@ -28,7 +28,7 @@ $(document).ready(function () {
     } else {
         loadGraph('1');
     }
-    
+
     $('#fcecount').hide();
 });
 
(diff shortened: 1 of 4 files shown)


public/js/graph/setup.js
```diff
@@ -21,7 +21,7 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph('graph-csc');
+    loadGraph(1);
```
david-yz-liu: Make this a string

public/js/graph/sidebar/sidebar_events.js
```diff
@@ -31,10 +30,11 @@ function createGraphButtons(graphs) {
     'use strict';
 
     for (var i = 0; i < graphs.length; i++) {
-        var graphTitle = graphs[i].graph_title.toLowerCase();
+        var graphTitle = graphs[i].title.toLowerCase();
         var graphButton = '<div id = "graph-' + graphTitle +'" class = "graph-button">';
         $('#graphs').append(graphButton);
         $('#graph-' + graphTitle).html(graphTitle.toUpperCase());
```
david-yz-liu: Why not use `graphs[i].title`?

public/js/graph/setup.js
christinem: Done!

Ian-Stewart-Binks:
@christinem Hmm. We'll need access to the current graph ID whenever the use is able to post a graph to Facebook, and we don't need access to every ID. Where do you think the best place to put it would be?

public/js/graph/sidebar/sidebar_events.js
christinem: Ah, yes I can definitely do so. Done!

christinem:
@Ian-Stewart-Binks if it's only the current graph ID you need, we can either use a cookie to denote the active graph ID or the data could be assigned to the `graph` element itself with a key `active-graph` or something of the sort.

david-yz-liu:
Seems like a perfect use for a cookie.

Ian-Stewart-Binks:
@christinem Since the information doesn't need to persist throughout multiple browsing sessions*, I don't know if a cookie would be the best house for it. Perhaps it would be best stored in the `.data` of the graph element. What do you think?

*I don't believe that's the correct use of the term 'browsing session', but you get the idea. Browser session: [A browser session is a continuous period of user activity in the browser, where successive events are separated by no more than 30 minutes.](https://blog.mozilla.org/metrics/2010/12/22/browsing-sessions/)

david-yz-liu:
If the current graph persists, users can pick up where they left off. Rather than always pull the CSC graph (or graph id '1'), check the cookie and grab that graph.

christinem:
@Ian-Stewart-Binks I actually think a cookie might be handy. I wanted to add in the feature of, if you switch tab, and go back to the graph tab, the active graph will load. Kind of like what happens when you switch the the POSt tab and whatever POSt was opened last, that's the one that opens right now.

christinem:
Ah @david-yz-liu beat me to it. 
I will now just add that functionality to pick up where someone left off.

Ian-Stewart-Binks:
@david-yz-liu @christinem Hadn't thought of that. I agree.

christinem:
Done!

public/js/graph/sidebar/sidebar_events.js
```diff
@@ -45,26 +45,29 @@ function createGraphButtons(graphs) {
...
+    setCookie('active-graph', id);
+    
     // Remove current graph
     $('#graph').empty();
 
-    getRemote(graph.toUpperCase() + '/' + graph + '_graph.svg');
+    getRemote('static/res/graphs/' + id + '.svg');
 
-    FCEPrerequisiteCourses = [csc318, csc454];
+    // Only create this if CSC graph loaded
```
Ian-Stewart-Binks: Good catch.

public/js/graph/setup.js
```diff
@@ -21,7 +21,8 @@ var nodes = [];               // List of all nodes
 $(document).ready(function () {
     'use strict';
 
-    loadGraph('graph-csc');
+    var active = getCookie('active-graph');
+    loadGraph(active);
```
david-yz-liu: The cookie might not exist; you need to check for this.

christinem: Ah right, done.

public/js/graph/sidebar/sidebar_events.js
```diff
@@ -45,26 +45,29 @@ function createGraphButtons(graphs) {
...
- * @param{string} id ID name of graph button clicked
+ * @param{string} id ID of graph in database
 **/
 function loadGraph(id) {
     'use-strict';
     
-    var graph = id.substring(6, id.length);
-
+    setCookie('active-graph', id);
+
```
david-yz-liu: Whitespace

christinem: Fixed.

david-yz-liu:
@christinem Okay, two more things before I merge this branch into master:

1) The FCE count bug needs to be fixed.
2) Please play around with making the sidebar look more attractive for clicking. E.g., give it a "question mark" icon, make it a bit thicker, and give it a different colour on hover.

PR #489, opened 2015-03-29
Facebook merge

david-yz-liu:

PR #505, opened 2015-04-01
Clean up fb

Ian-Stewart-Binks:
This pull request cleans up the Facebook code that was recently pulled into master in #489.

Note that I have also made a change in `graph/node.js`. Before, the code was checking the logical type against both `or` and `OR` (and `and` and `AND`). The code now converts the logical type to uppercase and compares it against `AND`.

### DIFF
--- a/public/js/grid/timetable_util.js
+++ b/public/js/grid/timetable_util.js
@@ -122,21 +122,21 @@ function restoreFromCookies() {
     var selectedSectionCookie = getCookie('selected-lectures');
 
     if (selectedCourseCookie === undefined || selectedCourseCookie.length === 0) {
-        selectedCourseCookie = '[]';
+        selectedCourseCookie = [];
     } else {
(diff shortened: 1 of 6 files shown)


public/js/grid/timetable_util.js
```diff
@@ -175,13 +174,13 @@ function saveCookies(courses, sections) {
     'use strict';
 
     if (courses !== undefined) {
-        setCookie("selected-courses", JSON.stringify(courses).replace(/,/g, '_').replace(/[\"\[\]]/g, ''));
+        setCookie("selected-courses", courses.join('_').replace(/[\"\[\]]/g, ''));
```
david-yz-liu: I don't know that the `replace` is necessary any more - did you check?

Ian-Stewart-Binks: It doesn't look like it, but when trying it out I've come across what appears to be an unrelated bug...

david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #527, opened 2015-04-12
Privacy Policy

Ian-Stewart-Binks:
This pull request adds in the initial privacy policy.


david-yz-liu:
Thanks for getting started on this, @Ian-Stewart-Binks. Only one comment right now - I don't think it should be a separate tab on the Nav bar. Link to it at the bottom of the About page.

Ian-Stewart-Binks:
@david-yz-liu I've added it to the README and removed it from MasterTemplate. Perhaps we should consider having a footer and add the Privacy Policy link to that, just like how GitHub has privacy in their footer.

david-yz-liu:
Thanks, @Ian-Stewart-Binks. Our current footer *certainly* can be improved.

david-yz-liu:
@Ian-Stewart-Binks I put in an amended, shorter privacy policy.

PR #528, opened 2015-04-12
Some facebook styling tweaks

david-yz-liu:
1. Styled the buttons (nav bar still needs work for spacing)
2. Removed the display of full name. My reasoning is it doesn't really do much for the user, and at this point another piece of information we need to be accountable for.
3. Only display "Post to Facebook" when on Graph/Grid, and logged in. In other situations, it looks misleading.

@Ian-Stewart-Binks Do you have any thoughts? My main hope is for our first version of a Privacy Policy to basically say "we don't use any of your personal information, we just let you post to Facebook."

### DIFF
--- a/hs/Css/CommonCss.hs
+++ b/hs/Css/CommonCss.hs
@@ -45,10 +45,11 @@ headerCSS = do
                 width $ px 200
                 display inlineBlock
                 margin (px 10) 0 (px 5) 0
+           height (px 50)
     "#nav-links" ?
         do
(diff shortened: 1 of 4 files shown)


Ian-Stewart-Binks:
1. Sounds good.
2. I agree, at this point it doesn't add much.
3. Sounds good, I agree.

Yes, I think that a simple(r) Privacy Policy is a good goal to have.

david-yz-liu:
Thanks!

PR #530, opened 2015-04-13
Graph viz

shaftoes:
Fleshed out graph generation. Running WebParsing/GraphVisualization.hs will now generate a series of svg's relating to each course code.


hs/Database/CourseQueries.hs
```diff
@@ -42,16 +57,13 @@ returnCourse lowerStr =
                                                                  TutorialsSession ==. "S"] []
         sqlTutorialsYear   :: [Entity Tutorials]  <- selectList [TutorialsCode ==. courseStr,
                                                                  TutorialsSession ==. "Y"] []
-        
+        let fallSession   = buildSession sqlLecturesFall sqlTutorialsFall
```
Ian-Stewart-Binks: If these let statements are being moved up here, I don't think we need the second `let`.

hs/Database/CourseQueries.hs
```diff
@@ -10,8 +19,13 @@ import Database.JsonParser
 import Happstack.Server
 import Data.List
 import qualified Data.Text as T
+import qualified Data.Aeson as Aeson
```
Ian-Stewart-Binks: I spent a long time removing unnecessary imports in a recent refactoring branch. `Aeson` is not used in this file, does it need to be imported? Are all of these imports necessary?

hs/SvgParsing/parser.hs
```diff
@@ -16,6 +16,7 @@ import qualified Data.Map as M (empty)
 main :: IO ()
 main = performParse "CSC" "graph_regions.svg"
 
+
```
Ian-Stewart-Binks: The majority of the hs files have been using a single blank line between functions. I don't mind if we use single or double, but since the rest of the code uses single, use single. We can change everything to double, but it needs to be brought up properly.

hs/WebParsing/GraphVisualization.hs
```diff
@@ -2,36 +2,67 @@
...
+junkCodes = ["CMS"]
 
+main :: IO ()
+main = do
+  rsp <- simpleHTTP (getRequest fasCalendarURL)
+  body <- getResponseBody rsp
+  let depts = filter (isPrefixOf "crs_")  (getDeptList $ parseTags body)
+  let codes = (map ((drop 4) . (takeWhile (/= '.')) . (map toUpper))  depts ) \\ junkCodes
+  mapM_ makeGraph codes
+  mapM_ (\x -> (runCommand $ "rm " ++ (map toUpper x))) codes
```
Ian-Stewart-Binks: Redundant parentheses.

hs/WebParsing/GraphVisualization.hs
```diff
@@ -2,36 +2,67 @@
...
+     (\g -> graphVizProcess filename (graphToDot graphParams g))
 
-main :: IO FilePath
-main = 
-  compsci >>= 
-  (\x -> return (graphToDot graphParams x)) >>=
-  graphVizProcess "compsci"
+-- | constructs a graph of all courses starting with code.
+makeGraph :: String ->  IO ProcessHandle
+makeGraph code = do
```
Ian-Stewart-Binks: I don't think the `do` is necessary here.

david-yz-liu:
Thank you, @shaftoes. I'd probably want to see a PR cleaning up the parsing code as the next major step.

PR #534, opened 2015-04-14
Cabal file

david-yz-liu:
@christinem You would be a good person to test this out on.

Try pulling this, and then do the following:

```
> cabal configure
> cabal build
```

You might need to `--force-reinstalls` the first command. Hopefully things won't go awry. This also attempts to install pandoc.


Ian-Stewart-Binks:
@david-yz-liu Pardon my memory, what was the cause of the Pandoc problem?

david-yz-liu:
I can't quite remember; perhaps conflicting dependencies.

christinem:
@Ian-Stewart-Binks yeah it was some kind of dependency error whenever I tried to run the server.
@david-yz-liu I tried running the first command on my Mac, which doesn't have pandoc installed, and I get this error and it stops:

```
christinemurad: ~/Documents/courseography/hs >> cabal configure
Resolving dependencies...
Configuring courseography-0.1.0.0...
cabal: At least the following dependencies are missing:
pandoc -any
```

Looks like its asking for it to already be installed, instead of attempting to install it.

I also tried running the first command on my Windows Machine and it gave a similar error, except for the dependencies `haxml, diagrams-lib, diagrams-svg, fb, and http-conduit` (which are not installed yet).

Must all dependencies be installed before I run this?

david-yz-liu:
Oh yeah. Try `cabal install only-dependencies` first

christinem:
That worked. Tested on Mac, after running the `cabal install --only-dependencies` command, it was able to configure and build fine, and the About Page is working fine. 

Windows is giving me a hard time with dependencies. Part of it was the gcc I had on my machine, and installing Cygwin binaries cleared up that issue and allowed me to install the network package, but a couple of other dependencies refuse to install that pandoc depends on. I'm getting things like `undefined reference to initWinSock` and `undefined reference to shutdownWinSock`. This might take some more work to figure out.

david-yz-liu:
@christinem Yeah, some preliminary research indicates this type of error comes from conflicting compilers. I would recommend uninstalling gcc, MinGW, Cygwin, Visual Studio, etc. and trying again. The Haskell Platform comes with its own MinGW, but you also need to install MSYS (http://www.mingw.org/wiki/msys)

christinem:
@david-yz-liu lo and behold, after many (many) hours later, finally being able to download all dependencies properly on my Windows Machine (I ended up uninstalling everything you suggested above, and installing a clean version of haskell, all cabal packages, and MSYS), during the build phase, I get the same error that was giving me issues before:

```
Loading package hslua-0.3.13 ... linking ... ghc.exe: C:\Users\Christine\AppData\Roaming\cabal\x86_64-windows-ghc-7.8.3\hslua-0.3.13\Hshslua-0.3.13.o: unknown symbol 'mingw_getsp'
ghc.exe: unable to load package 'hslua-0.3.13'
```

Same error happens when I try to directly run the server (which was where the issue occurred before).

david-yz-liu:
Interesting. Try using the Windows installer for pandoc: https://github.com/jgm/pandoc/releases/tag/1.13.2 (at the bottom). Don't know why I didn't think of this earlier.

PR #538, opened 2015-04-15
Database upgrade

david-yz-liu:
Parsing the new 2015-2016 timetable, plus a database name switch. @shaftoes You'll be interested in a couple of the minor parsing changes, as I had to resolve some runtime errors.

### DIFF
--- a/hs/Database/JsonParser.hs
+++ b/hs/Database/JsonParser.hs
@@ -26,7 +26,7 @@ fbdbStr :: T.Text
 fbdbStr = "fdatabase1.sqlite3"
 
 dbStr :: T.Text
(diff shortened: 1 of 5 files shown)


david-yz-liu:
Note to self - crs_cjs link seems to be be broken, which caused the parser to fail.

shaftoes:
The "lastH2" in the jewish studies is not the list of courses, so its failing. A quick fix can be done by modifying which header is extracted from the partition. Want me to make a commit? Or save it until after our talk on Monday.

david-yz-liu:
@shaftoes It's okay for now, I think that whole page is being skipped for now.

PR #547, opened 2015-04-16
Reset button disappear when switching tabs Fix

christinem:
@david-yz-liu The fix that you made for the reset button for Firefox ended up making the Reset button disappear whenever you switched between the focus or graph tab. So I just moved the line over to toggleSidebar (when the sidebar is closed). Seems to be working fine on both Chrome and Firefox.


david-yz-liu:
Thanks @christinem!

PR #557, opened 2015-04-18
Add development information to README

arkon:
For https://github.com/Courseography/courseography/issues/555

PR #560, opened 2015-04-19
Move 'Get Involved' from README to CONTRIBUTING

elliottsj:
It's a minor change, but GitHub will now add a notice when contributors make a pull request:

See <https://github.com/blog/1184-contributing-guidelines>

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Courseography was envisioned in late 2013 by [David Liu](http://www.cs.toronto.e
 
 See [CONTRIBUTING.md](https://github.com/Courseography/courseography/blob/master/CONTRIBUTING.md).
 
+Click [here](/privacy) to learn about our Privacy Policy.
+
 ### Development
 
 Courseography is powered by [Haskell](https://www.haskell.org/).
--- a/README.md
+++ b/README.md
@@ -1,20 +1,15 @@
 Courseography [![Slack][slackin-badge]][slackin]
-==================
-#About Courseography
+=============
+
 Here at the University of Toronto, we have hundreds of courses to choose from, and it can be hard to navigate prerequisite chains, program requirements, and term-by-term offerings all at once. That's where Courseography comes in: by presenting course and scheduling information in a set of graphical interactive tools, we make it easier to choose the right courses for your academic career. Whether it's making sure you'll satisfy all the prerequities for that 4th year course you really want to take, or fitting together fragments of your schedule for next term, we hope Courseography makes your life easier!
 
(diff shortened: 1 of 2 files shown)


Ian-Stewart-Binks:
@elliottsj Thanks for the addition! This is a feature of GitHub that I wasn't aware of, and I'm not certain what difficulties new contributors may face, so this really helps.

Some notes:
The README.md file is used as the about page on the website (`\about`). This can be seen in the [Response file](https://github.com/Courseography/courseography/blob/master/hs/AboutResponse.hs) (and also in the [server](https://github.com/Courseography/courseography/blob/master/hs/server.hs). The current implementation brings forth problems, such as the one that is faced right now: when the README.md file is parsed and output as `\about`, the privacy link (to `\privacy`, which is also implemented in the same way as `\about`/README.md are) won't work, as it links to PRIVACY.md, not `\privacy` (`http://localhost:8000/PRIVACY.md` doesn't exist). The same is also true when accessing the README.md file from GitHub, I don't believe that GitHub knows about `\privacy`.

The link to the privacy policy needs to be accessible from the website by users of all kinds, and CONTRIBUTING.md may not be the best home for it.

The 'Development' section of README.md might also be better placed in CONTRIBUTING.md.

arkon:
On the same topic, is it overkill to convert the Markdown files for the website using Pandoc? There isn't much of it, and it's not terribly hard to convert Markdown to HTML normally, nor should the files be changing _that_ often.

elliottsj:
I think using pandoc is fine, but we should probably use a different set of documents for the app's about/privacy pages than the ones used on GitHub. Especially with the addition of [developer-oriented content](https://github.com/Courseography/courseography/pull/557) and [badges](https://github.com/Courseography/courseography/pull/561), which don't really need to be shown on the `/about` page.

Something like:
- `pages/ABOUT.md`
- `pages/PRIVACY.md`

...which would just be used by the app.

And `README.md` and `CONTRIBUTING.md` would be used just for GitHub.

arkon:
Could also consider just putting the link to the privacy policy in the footer instead of in the README file.

david-yz-liu:
@elliottsj I like your idea about splitting the README off into a separate "About" page. Until that happens, though, we should keep the privacy link the way it is.

README.md
```diff
@@ -1,20 +1,17 @@
...
-#Get Involved
-One of our main goals for Courseography is to keep it a student-driven project. This means that if you want to get involved, you can! The best way to get started is by shooting us an email, but you should also feel free to check out all of the source code on Github and the accompanying Projects page.
-
-If you're coming from outside of the Department of Computer Science, we'd love to hear from you! We welcome help on making new graphs, understanding department- or program-specific policies, pointing out bugs or incorrect information, and general thoughts on the design of our tools. You don't need to be a programmer to get involved.
+### Get Involved
 
-We can be reached at courseography@cs.toronto.edu, and look forward to hearing from you!
+See [CONTRIBUTING.md](https://github.com/Courseography/courseography/blob/master/CONTRIBUTING.md).
 
 Click [here](/privacy) to learn about our Privacy Policy.
```
elliottsj: Ok, I've reverted the privacy link back to `[here](/privacy)`

Issue #569, opened 2015-04-24
Convert SVG graph rendering to React


elliottsj:
Work-in-progress over on https://github.com/Courseography/courseography/tree/react-graph

This requires some tooling changes in order to make components modular and unit-testable via [webpack](http://webpack.github.io/docs/) and [Jest](https://facebook.github.io/jest/):

- [x] Make the app a Node.js package (by adding `package.json` and specifying dependencies)
- ~~Configure [Brunch](http://brunch.io/) with webpack to bundle JavaScripts (starting with just the `/graph` page)~~ **Using just webpack instead**
- [x] Configure webpack to bundle JavaScripts (starting with just the `/graph` page)
- Create universal scripts, [like GitHub does](http://wynnnetherland.com/linked/2013012801/bootstrapping-consistency/), to install/run/test Courseography:
  - [x] `script/bootstrap`: ensure dependencies are satisfied (Haskell platform, Node.js) and install them if necessary
  - [x] `script/dev`: launch the Haskell server and Brunch tasks _concurrently_, hot-reloading files upon change (need [happstack-plugins](http://hackage.haskell.org/package/happstack-plugins) to hot-reload the Haskell server). Processes will run concurrently using a `Procfile.dev` file and [node-foreman](https://github.com/strongloop/node-foreman).
  - [ ] `script/server`: bundle assets and build the Haskell server for production, then launch the server
  - [ ] `script/test`: execute Haskell unit tests, then JavaScript unit tests

david-yz-liu:
Closed in #805

PR #571, opened 2015-04-25
Add package.json

elliottsj:
Ref #569


elliottsj:
Whoops, need to request merge into `react-graph` instead.

PR #572, opened 2015-04-25
Add bootstrap script

elliottsj:
Running `script/bootstrap` will ensure that you have the Haskell platform and Node.js installed correctly, then it will install cabal & npm dependencies

Ref #569


elliottsj:
Whoops, need to request merge into `react-graph` instead.

PR #574, opened 2015-04-25
Add bootstrap script

elliottsj:
Running `script/bootstrap` will ensure that you have the Haskell platform and Node.js installed correctly, then it will install cabal & npm dependencies

Ref #569


david-yz-liu:
@elliottsj It occurred to me to ask about Windows - correct me if I'm wrong, but it looks like `script/bootstrap` needs a Windows equivalent.

arkon:
It might be wishful thinking, but perhaps something like [Gow](https://github.com/bmatzelle/gow) might help?

elliottsj:
Mhm, the script depends on bash, so a bash distribution for Windows (like in Gow) is required.

Alternatively, we could use [Batsh](https://github.com/BYVoid/Batsh) to compile both a bash script and a Windows batch script.

For now though, I think it's enough to just document in the wiki that the bootstrap script does not support Windows and setup must be done manually.

Also worth noting: [nvm](https://github.com/creationix/nvm) does not support Windows; [nvmw](https://github.com/hakobera/nvmw) or [nvm-windows](https://github.com/coreybutler/nvm-windows) must be used instead.

david-yz-liu:
Okay, I'll keep that in mind. Some students have been having problems compiling the Haskell libraries on Windows, but I would like to continue supporting Windows (and making it just as easy to work on as Linux) as much as possible. Thanks for your input.

PR #582, opened 2015-04-27
Use `notFound` for 404 response

### LINKED ISSUE #581: 404 page returns HTTP status 200
Accessing any route that does not exist displays the 404 page, but returns HTTP status code 200. This can cause problems detecting errors when trying to load assets that do not exist.

e.g. Loading `static/res/grahs/404.svg` via AJAX will return 'success' even though the resulting data is just the content of the 404 page.

elliottsj:
Resolves #581


david-yz-liu:
Thanks, @elliottsj!

PR #585, opened 2015-04-28
Remove unused image assets and update others

arkon:


david-yz-liu:
I would like to start using the checkmark again at some point in the near future; would you mind keeping that one in there, or perhaps replacing it with a nicer-looking one?

arkon:
@david-yz-liu Added a checkmark icon.

david-yz-liu:
Thanks @arkon. FYI, you'd accidentally committed the `courseography` binary; I removed it when merging. This is more impetus to fix the dumb relative paths in the current build process.

arkon:
@david-yz-liu Whoops, didn't notice that I accidentally added that file.

PR #589, opened 2015-04-28
Use same font as rest of graph for bool nodes

arkon:
Doesn't really seem like an appropriate place to use Comic Sans.


david-yz-liu:
I'm fine with picking a different font, but I don't really like the look of Arial.

I would also prefer not changing the csc2014 graph.

arkon:
@david-yz-liu How about Trebuchet? It's the same font as the rest of the graph.

PR #590, opened 2015-04-28
Add request logging

elliottsj:

### DIFF
--- a/hs/server.hs
+++ b/hs/server.hs
@@ -2,6 +2,7 @@ module Main where
 
 import Control.Monad (msum)
 import Control.Monad.IO.Class (liftIO)
+import Data.Maybe (fromMaybe)
 import Happstack.Server
 import GridResponse
 import GraphResponse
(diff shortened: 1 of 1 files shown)


hs/server.hs
```diff
@@ -16,11 +18,32 @@ import Database.CourseQueries (retrieveCourse, allCourses, queryGraphs, courseIn
...
 import System.Directory
+import System.Environment (lookupEnv)
+import System.IO (hSetBuffering, stdout, stderr, BufferMode(LineBuffering))
+import System.Log.Logger (logM, updateGlobalLogger, rootLoggerName, setLevel, Priority(INFO))
 import CourseographyFacebook
 import qualified Data.Text as T
 
+-- | log access requests using hslogger and a condensed log formatting
+--
+logMAccessShort :: FormatTime t => LogAccess t
```
elliottsj: Note: see <https://hackage.haskell.org/package/happstack-server-7.4.3/docs/src/Happstack-Server-Internal-Types.html#logMAccess> for the default `logMAccess` function

PR #591, opened 2015-04-29
Update gitignore

arkon:
Ignore more generated graph SVGs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ cabal.sandbox.config
 # generated resources
 /public/res/graphs/1.svg
 /public/res/graphs/2.svg
+/public/res/graphs/3.svg
+/public/res/graphs/4.svg


christinem:
We probably want to eventually end up putting the generated graphs into a separate folder and just ignoring that folder. The number of graphs parsed will always change and there really should only be two graphs parsed and generated at one time currently: Computer Science and Statistics.

david-yz-liu:
There are two issues here:

1. As @christinem pointed out, we should move the hand-made graphs out of `public/res/graphs/` and into a different folder, e.g. `graphs/`. @arkon I would happily accept a pull request that does this, so that we can properly gitignore the `public/res/graphs` folder.
2. We aren't making a distinction between updating and regenerating the database, and so unfortunately every time we run `SvgParsing/parser.hs`, we get two new graphs in the database, and two new SVG files. That's not great.

PR #597, opened 2015-04-29
Switch from pandoc library to markdown library.

### LINKED ISSUE #594: Stop using Pandoc
Okay, the fact that @christinem, @jmatholia, @foreverwander, and now @akbiggs are all having the same problem means that I'm the odd one out when it comes to Windows installation.

One possibility is to simply rewrite the About and Privacy pages in some other format to eliminate the need for conversion.

My favoured solution, however, is to swap out Pandoc for the much lighter [`markdown`](https://hackage.haskell.org/package/markdown) package.

If someone I tagged on Windows could try this out and see if this works, that would be fantastic.

akbiggs:
Fixes #594

### DIFF
--- a/hs/AboutResponse.hs
+++ b/hs/AboutResponse.hs
@@ -5,9 +5,11 @@ module AboutResponse where
 import           Text.Blaze ((!))
 import qualified Text.Blaze.Html5 as H
 import qualified Text.Blaze.Html5.Attributes as A
+import           Text.Blaze.Html.Renderer.Text (renderHtml)
+import Data.Text.Lazy
+import Text.Markdown
 import Happstack.Server
 import MakeElements
-import Text.Pandoc
 import MasterTemplate
 
 aboutResponse :: String -> ServerPart Response
@@ -32,4 +34,4 @@ aboutHtml contents = H.div ! A.id "aboutDiv" $ mdToHTML contents
 -- | mdToHTML takes in the contents of a file written in Mark Down and converts it to 
 -- blaze-HTML.
 mdToHTML :: String -> H.Html
-mdToHTML contents = writeHtml def $ readMarkdown def contents
+mdToHTML contents = markdown def $ pack contents
--- a/hs/AboutResponse.hs
+++ b/hs/AboutResponse.hs
@@ -5,13 +5,13 @@ module AboutResponse where
 import           Text.Blaze ((!))
 import qualified Text.Blaze.Html5 as H
 import qualified Text.Blaze.Html5.Attributes as A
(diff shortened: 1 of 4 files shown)


hs/AboutResponse.hs
```diff
@@ -5,9 +5,11 @@ module AboutResponse where
 import           Text.Blaze ((!))
 import qualified Text.Blaze.Html5 as H
 import qualified Text.Blaze.Html5.Attributes as A
+import           Text.Blaze.Html.Renderer.Text (renderHtml)
```
david-yz-liu: I don't think this import is necessary

hs/AboutResponse.hs
```diff
@@ -5,9 +5,11 @@ module AboutResponse where
...
+import Data.Text.Lazy
```
david-yz-liu: For this and the one below it, please indicate which function is being imported, since I think there's only one.

hs/AboutResponse.hs
```diff
@@ -32,4 +34,5 @@ aboutHtml contents = H.div ! A.id "aboutDiv" $ mdToHTML contents
 -- | mdToHTML takes in the contents of a file written in Mark Down and converts it to 
 -- blaze-HTML.
 mdToHTML :: String -> H.Html
-mdToHTML contents = writeHtml def $ readMarkdown def contents
\ No newline at end of file
+mdToHTML contents = markdown def $ pack contents
```
david-yz-liu: But also, I wouldn't mind if we used the `Data.Text.Lazy` `Text` type all the way through, to avoid the call to `pack` here. This would involve using `Data.Text.Lazy.IO`'s `readFile` in `server.hs`.

akbiggs:
@david-yz-liu I have fixed the issues you mentioned.

hs/AboutResponse.hs
```diff
@@ -5,12 +5,13 @@ module AboutResponse where
 import           Text.Blaze ((!))
 import qualified Text.Blaze.Html5 as H
 import qualified Text.Blaze.Html5.Attributes as A
+import Data.Text.Lazy (Text, pack)
```
david-yz-liu: So I guess the `pack` is no longer necessary.

akbiggs:
@david-yz-liu Unused imports are gone now.

david-yz-liu:
Excellent. Let's make sure that @jmatholia and @christinem know Windows should work just fine again. Thanks, @akbiggs!

christinem:
Thanks @david-yz-liu and @akbiggs!

Issue #592, opened 2015-04-29
FCE Count should be shared between graphs


david-yz-liu:
Right now, it seems the two graphs track their own FCE counts.

christinem:
I'll take this on.

david-yz-liu:
Note: this is a more complex problem due to courses that appear in multiple graphs.

PR #602, opened 2015-05-02
Shared FCE Count fix

christinem:
This fixes issue #592.


public/js/graph/setup.js
```diff
@@ -2,13 +2,17 @@
 // Globals
 var timetable = 'static/res/timetable.html';
 var clickedCourses = [];
-var FCEs = 0;
+
+// FCEs for current graph
+var FCEs = 0;
```
david-yz-liu: Maybe this doesn't need to be global.

christinem: We could leave `totalFCEs` as the only global variable and just increase the credit count with that variable. Do you think we have any need of keeping track of credit counts for individual graphs? Right now I store those as cookies, but if not I could make it simple and remove that code, and keep `totalFCEs` as the only global. It's really the only reason `FCEs` is still global.

david-yz-liu: I see. If `FCEs` tracks the current FCEs selected on the current department, then why don't you:

1) rename the variables
2) display its value alongside the total FCE count (perhaps just in parentheses after the global total, for now)

christinem:
@david-yz-liu I've made it so that the current graph FCE total is in brackets beside the global FCE count (although I'm not sure how clear it is to the user's eye). Pressing "Reset Graph" only resets the FCEs of the current graph, and adjusts the global value accordingly to whatever credits are left selected in the other graphs. Is that functionality okay?

david-yz-liu:
Thanks, @christinem! The behaviour of the "Reset Graph" is fine. I disabled the total from actually appearing (even though it is computed) because some courses are shared between graphs, and they would be counted twice, which isn't great. A far more complicated solution is required for that.

PR #603, opened 2015-05-06
Graph Generating Fix

christinem:
Graphs were not being generated in the right "gen" folder. This fixes that issue.


david-yz-liu:
I'm confused, shouldn't this folder always exist? It's in the repo with a .gitkeep file

Issue #613, opened 2015-05-07
Fix GHC warnings


david-yz-liu:
If we compile our code with `-Wall` (this will be done by default in a pull request soon), we get lots of GHC warnings. We should eliminate these.

Ian-Stewart-Binks:
@david-yz-liu What are your thoughts on non-exhaustive pattern matches?

david-yz-liu:
Generally, the non-exhaustive pattern matches indicate some larger failure in the types. These are what we should be aiming to fix; but, I am open to the idea that this won't always be possible. We shall see.

david-yz-liu:
@PkHutch @RyanDJLee @AilsaF this is a good issue to work on for getting to know Haskell as well. It also has the benefit that you can all work on this, since it's just a matter of fixing the errors for certain files, and there's plenty of Haskell files to go around.

I *think* you can try to do something like `stack ghc <path-to-file>` to try to compile a particular file, but I could be wrong. Worth investigating and documenting that!

RyanDJLee:
Okay! I still have some reading to do I realized, but I will be on it shortly!

PR #615, opened 2015-05-07
Add type signatures, remove redundancies

cchens:
Added some missing type signatures and removed unused import statements and bindings.

Ref #613. Does not get rid of all the warnings, but hopefully it's a start.


david-yz-liu:
Thanks, @cchens. Good start. In general, try to name commit messages with a prefix like `issue-613: CSS type signatures` or even `ghc-warnings: issue-613`.

In the file JsonParser.hs, either delete the functions if they are not used any more, or do not comment them out at all.

cchens:
@david-yz-liu I've removed the unused functions, as well as an unused type declaration in `ParsingHelp.hs`

david-yz-liu:
Thanks, @cchens!

Issue #309, opened 2015-01-30
Add Foreign Key Constraints To Tables.hs


Ian-Stewart-Binks:
Tables.hs does not contain any foreign key constraints, but it should. I'm not certain if this is possible with Persistent, but it's worth looking into.

If I recall correctly, foreign key constraints need to be enabled for Persistent's SQLite.

david-yz-liu:
This can certainly be done (and should be). See #620 for initial work on this.

david-yz-liu:
@klmerryn Why don't you work on this.

Take a look at the Lecture and Course tables in app/Database/Tables.hs. The Lecture table has a column called "code", which is currently just text (you can see this in the contents of the database itself). Try removing this column and replacing it with a CourseId column instead. Note that this actually will break a few different things, so you'll need to look at the existing code carefully to see how we deal with existing keys, and read the Persistent documentation carefully.

klmerryn:
thought I had replied already -- on it, thank you!

PR #620, opened 2015-05-09
Work for #309 - Graph foreign key

david-yz-liu:
This removes the `gId` attribute from the Graph table, and makes the `gId` columns in other tables a foreign key rather than a number.

Others should be able to build on this work to use foreign keys in other places, as well as remove manual storage of id attributes in the tables.


Ian-Stewart-Binks:
Nice. Are we able to get rid of the Data.Int imports?

david-yz-liu:
Yes, thanks @Ian-Stewart-Binks!

Issue #566, opened 2015-04-22
Check My POSt: Using Text Boxes to 'Select' Courses


Ian-Stewart-Binks:
Entering courses into the text boxes in the 'Check My POSt' feature doesn't work. I can't seem to successfully add the course 'CSC494', nor '494'.

christinem:
Ah, right. Well the only textboxes that are meant to be editable are the ones in the category with the MAT and STA, but yes CSC49* is not recognized. I can add that in.

Issue #580, opened 2015-04-27
Errors in Check My POSt


cchens:
Some of these are simply typos, others are not. I've listed all the ones I can find here:

- MAT137 and 157 listed as H courses (should be Y)
- MAT240 not listed with 221 and 223 for Specialists
- STA248 is written as 249
- ECE489 should count as a 400-level course for Specialists
- Linear algebra does not currently count towards Majors (should be an extra course)
- CSC200 does not currently count for anything (should count for Majors and Minors)

christinem:
@cchens Thanks for catching these! The sheets that I was given from the undergrad department are missing a couple of these points that are on the online calendar, such as the ECE489 being in the 400 level category (it is technically in another category but Degree Explorer says it should be with the 400 levels too) and 200 level courses for the minor (@david-yz-liu I'll probably talk to Leanne about those). The others are oversights on my part. I will fix them.

cchens:
Apparently ECE385 and ECE489 don't count for anything

david-yz-liu:
@christinem The errors which are listed in this issue are much more tractable than #566; you should be able to fix these quite easily.

Ian-Stewart-Binks:
Has this been completed?

christinem:
@Ian-Stewart-Binks not yet, there's a couple more to tackle. Fixing these has proven a bit harder than anticipated.

david-yz-liu:
@christinem can you double-check the status of this issue when you get a chance?

PR #623, opened 2015-05-10
More Check My Post Fixes

christinem:
This fixes Issue #566 and #580, as well as making other bug fixes and additions such as:

* You can now type in BCB courses in each of the categories
* Editable text boxes become disabled after you type in them and press 'Update POST'
* Some edits to the actual POSt information using info from Degree Explorer. I specifically took out ECE489H from the first 300+ level category because it is included in the second one (the Extra category), and degree explorer groups them together differently anyways, so how it is now ends up being exactly the same. It just made the autofilling for ECE courses cleaner

Please feel free to test it out and let me know if you find a bug somewhere. I've tried testing it extensively but I could have missed a use case.

The code could probably use some cleaning up, but since this will be moved over soon to React, maybe that can wait.

### DIFF
--- a/hs/PostResponse.hs
+++ b/hs/PostResponse.hs
@@ -236,7 +236,7 @@ checkPost =
                     H.p ! A.class_ "full_name CSC236240" $ "CSC236H (Introduction to the Theory of Computation)"
                     H.p ! A.class_ "full_name CSC236240" $ "CSC240H (Enriched Introduction to the Theory of Computation)"
             H.div ! A.id "min_misc" $ do
-                H.p ! A.class_ "code" $ "Any 300/400-level CSC course (atleast 1.0 FCE), CSC209H, CSC258H, CSC263H/CSC265H (1.5 FCEs)"  
+                H.p ! A.class_ "code" $ "Any 200/300/400-level CSC course (atleast 1.0 FCE), CSC209H, CSC258H, CSC263H/CSC265H (1.5 FCEs)"  
                 H.div ! A.id "minextra" ! A.class_ "more-info" $ do
                     H.input ! A.type_ "text" 
                     H.input ! A.type_ "text"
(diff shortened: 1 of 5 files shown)


hs/PostResponse.hs
```diff
@@ -236,7 +236,7 @@ checkPost =
                     H.p ! A.class_ "full_name CSC236240" $ "CSC236H (Introduction to the Theory of Computation)"
                     H.p ! A.class_ "full_name CSC236240" $ "CSC240H (Enriched Introduction to the Theory of Computation)"
             H.div ! A.id "min_misc" $ do
-                H.p ! A.class_ "code" $ "Any 300/400-level CSC course (atleast 1.0 FCE), CSC209H, CSC258H, CSC263H/CSC265H (1.5 FCEs)"  
+                H.p ! A.class_ "code" $ "Any 200/300/400-level CSC course (atleast 1.0 FCE of 300/400-level) (1.5 FCEs)"
```
david-yz-liu: at least should be 2 words

public/js/post/fill_textboxes.js
```diff
@@ -11,10 +11,11 @@ function fill300Textboxes(post, postElement) {
 
         var course = activeCourses[m];
 
-        if (course.indexOf('CSC3') !== -1 &&
-            (post.name === 'major' ||
+        if ((course.indexOf('CSC3') !== -1 || course.indexOf('ECE3') !== -1 ||
```
david-yz-liu: use `course.startsWith`

david-yz-liu: change to `=== 0`

public/js/post/fill_textboxes.js
```diff
@@ -221,4 +236,4 @@ function fillMisc() {
 
     // Update category
     updateInqCategory();
-}
+}
```
david-yz-liu: Put back the new line

david-yz-liu:
#566 Tried typing CSC490 and CSC490H into the 400+ checkbox for the specialist, but they weren't saved.

#580: Linear Algebra still doesn't count for the major. CSC200 isn't listed in the major.

If the textboxes are disabled, people can't go back and remove courses that they no longer want to take. That's not good.

PR #624, opened 2015-05-10
Biochemistry graph

cchens:
Here's a graph for the biochemistry department, which includes BCB courses. A goal for the future could be to have the BCB nodes work for Check My POSt. There's been a few odd kinks that needed to be worked out (such as things not rendering properly at certain coordinates), so it would be great if people could help test this out.

There is one region without a label. It's not a bug; I just couldn't come up with a good label for it ("General Biochemistry"?). The department calls them "lecture courses", which sounds kind of eh.


david-yz-liu:
Looks great! Couple of comments:

1. I think you should just stick with the department's categorization of "Lecture Courses"
2. You'll need to hard-code the hybrid relationships for now. See `parse_graph.js`

I'll show the Biochemistry department, and see how they like it!

cchens:
@david-yz-liu Just curious, what does hard-coding the hybrid relationships do? The hybrid nodes seem to include all the grey dummy nodes (at least, for this graph), but they appear to behave properly without hard-coding.

david-yz-liu:
Ah, you're correct - each of your dummy nodes only has a single course, so no hard-coding is required.

david-yz-liu:
Thanks, @cchens! This is quite exciting.

Issue #518, opened 2015-04-04
Draw: Moving Region


Ian-Stewart-Binks:
The behaviour for moving a region is a bit off.

arkon:
In `getClickPosition` in [draw.js](https://github.com/Courseography/courseography/blob/master/public/js/draw/draw.js), I think removing  the `parentPosition` stuff fixes this issue.

I'm not super familiar with the code base though, so I might be wrong, since I'm just basing it off of playing around with the code in Chrome Dev Tools.

PR #625, opened 2015-05-12
Fixes issue 518: region movement

jmatholia:
Simple fix for the incorrect offset applied when moving a region (#518)  in draw.js.


david-yz-liu:
Thanks, @jmatholia! I did merge this, not sure why this wasn't picked up.

PR #626, opened 2015-05-14
Abstract constants that change between dev and prod into configuration file.

akbiggs:
Before running cabal build, and after any changes are made to the configuration file, the chosen configuration file needs to be copied from `<ENVIRONMENT>_Config.hs` to `Config.hs`. This file will then be used for the settings in the environment. This allows us to override the configuration variables with different values in production.

### DIFF
--- a/hs/Development_Config.hs
+++ b/hs/Development_Config.hs
@@ -8,7 +8,7 @@ When the project is deployed to the production environment, this file should be 
 containing the production values.
 -}
 
-module Config (dbStr) where
+module Config (dbStr, markdownPath, graphPath, genCssPath) where
 
 import Data.Text (Text)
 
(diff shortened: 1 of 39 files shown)


david-yz-liu:
Thanks, @akbiggs. Couple things:

1. Please remove `fbdbStr` from the config file and just put it into `CourseographyFacebook`.
2. Please add some Haddock documentation for the Config module and each value
3. Please move the other path-related strings into this file.

akbiggs:
@david-yz-liu Done

david-yz-liu:
Thanks, @akbiggs!

hs/Development_Config.hs
```diff
@@ -0,0 +1,36 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+{-
```
david-yz-liu: @akbiggs Didn't notice earlier, but if you want Haddock to know about this, you need to use `{-|`

PR #629, opened 2015-05-16
Modifications to CSS compilation

david-yz-liu:
This pull request changes how the CSS is compiled and used in Courseography. The main changes are:

- Running the server (`cabal run`) no longer recompiles the css. Instead, there is a separate task `cabal run css` which recompiles the css. This has the benefit of not having to restart the server to regenerate the css, although this is a pretty small savings in time overall.
- Added a configuration option (in `Development_Config.hs`) to minify the css, although by default it is still pretty-printed.
- All of the css is now contained in a single css file `public/style/app.css`. This is meant to reduce the number of requests (though again, not too many currently), and move us towards using a unified style across the different pages. So far I've only seen one difference: the text `input` elements in the Search, Post, and Draw pages look different. @christinem @jmatholia you might know better than I if any other stylistic things changed.


christinem:
@david-yz-liu I tried to build this and got this error:

` Css/Compiler.hs:17:28: Module ‘Config’ does not export ‘cssStyle’ `

david-yz-liu:
@christinem Thanks, I think I didn't recompile that file, so didn't catch that first error. Interesting.

Regarding the second error you mentioned, you need to copy the new `Development_Config.hs` file over to `Config.hs`.

christinem:
@david-yz-liu awesome, thanks, it build fine now :)

You're right, I can only see the text `input` elements that have changed stylistically.

Another one is the page background for all the pages have seemed to follow suite with the Check My POSt background (which is not really a bad thing, just a point made).

Other than that everything seems to look the same.

PR #632, opened 2015-05-18
Fixed non-exhaustive pattern match warnings

cchens:
Fixes non-exhaustive pattern match warnings for #613 by making all the unnecessary patterns return an error. Not sure if this is bad practice, though.


david-yz-liu:
We should be working at fixing the code to avoid pattern matching errors altogether, not raising more errors explicitly. It would be good for you to think about how to do this on a case-by-case basis.

PR #633, opened 2015-05-19
Issue 566 CSC49* Fix

christinem:
This fixes Issue #566. You can now type in CSC49* courses in each of the categories.


public/js/post/fill_textboxes.js
```diff
@@ -56,11 +56,17 @@ function fill300s() {
 
     // Clear textboxes
     for (var i = 0; i < 3; i++) {
+         if (spec300s[i].value.indexOf('CSC49') !== -1 
+            && activeCourses.indexOf(spec300s[i].value) === -1) {
```
david-yz-liu: Put the line breaks after the operators, not before them.

public/js/post/update_post.js
```diff
@@ -61,20 +61,20 @@ function resetValues() {
...
-                 'reqs': ['CSC108', 'CSC148', 'CSC165240', 'CSC207', 'CSC236240'], 'textboxesExtra': 3, 'categories': 6,
-                 'creditCount': 0, 'additionalMin200': ['CSC209', 'CSC258', 'CSC263265'], 'name': 'minor'};
+    major = {'index300': 0, 'index400': 0, 'categoriesCompleted': 0, 'filledTextboxes300': 0,
+             'filledTextboxes400': 0, 'filledTextboxesExtra': 0, 'majCount': 0, 'reqs': ['CSC108',
+             'CSC148', 'CSC165240', 'CSC207', 'CSC236240', 'CSC258', 'CSC263265', 'STA247255257Sta1', 
+             'MAT135136137157Calc1'], 'textboxes300': 2, 'extraTypedNextBox': 0,
+             'textboxes400': 1, 'textboxesExtra': 3, 'categories': 13, 'creditCount': 0, 'name': 'major'};
+    minor = {'index300': 0, 'index400': 0, 'categoriesCompleted': 0, 'filledTextboxesExtra': 0,
+             'reqs': ['CSC108', 'CSC148', 'CSC165240', 'CSC207', 'CSC236240'], 'textboxesExtra': 3, 'categories': 6,
+             'creditCount': 0, 'additionalMin200': ['CSC209', 'CSC258', 'CSC263253'], 'name': 'minor'};
```
david-yz-liu: Don't change this

david-yz-liu:
@christinem In a future pull request, it would be nice to not be checking `indexOf(...) > -1` but rather `indexOf(...) === 0`, since we really want to check that the *start* of the string is something.

public/js/post/update_post.js
christinem: Without this, the credit counts update incorrectly, since Check My POSt relies on the global variables, so the global variables need to be reset.

christinem:
@david-yz-liu Thanks. Made the line break changes. The removed `var`s need to remain as they are, as I mentioned above, unless I change the logic of how the post credit count is calculated. I can change the indexOfs in a future pull request - unless you would like me to do them in this one which I can.

public/js/post/update_post.js
david-yz-liu: I was referring to the `CSC263253`, which is incorrect.

christinem: Ah my mistake. I've fixed it.

david-yz-liu:
I was playing around with this and don't think it quite works. 

1. changes aren't saved when page refreshes
2 updates should be bound to a `keyUp` event so that users don't need to click on "Update POSts" (which should be removed altogether)

It has become clear to me that the structural problems with this portion of the code are preventing easy fixes like this from happening. It is quite possible to fix (2) within the current code, but not (1).

@christinem You can disable all of the input boxes, but don't try to fix the problem without consulting me first.

christinem:
Alright @david-yz-liu. Sorry about that, I guess I keep missing stuff. I have been thinking myself actually that I want to change some of the logic of how Check My POSt works, since it does seem to be making things more complicated at the moment the way it is structured now. We can discuss it more thoroughly on Saturday.

PR #641, opened 2015-06-04
Issue 580 - CSC200 fix

christinem:
This fixes part of Issue #580 - it autofills CSC200 in the Extra category for the major and minor.


david-yz-liu:
Thanks, @christinem!

PR #646, opened 2015-06-23
Logo And Color Scheme changes

christinem:
This pull request adds the logo to Courseography and (with the great help of @cchens) changes the color scheme and look of text of the top nav bar and sidebar. New sidebar icon is also added.


hs/Css/Graph.hs
```diff
@@ -231,14 +231,14 @@ sidebarCSS = do
         width (px 40)
         height100
         float floatLeft
-        backgroundColor purple2
+        backgroundColor $ purple10
```
cchens: Why does this one need the $ operator?

christinem: It doesn't, just fixed it. Thanks for catching it!

hs/Css/Graph.hs
```diff
@@ -330,6 +330,12 @@ sidebarCSS = do
         marginBottom (px 20)
         ":hover" & do
             backgroundColor grey2
+    ".flip" & do
+        "-webkit-transform" -: "scaleX(-1)"
```
david-yz-liu: Clay actually can generate vendor prefixes in many cases, including [transform](https://hackage.haskell.org/package/clay-0.10.1/docs/Clay-Transform.html). Try something like `transform $ scaleX (-1)`

david-yz-liu:
Thanks, @christinem and @cchens! I suggested one change in the CSS. Also, I wonder if perhaps we should use the nav bar colour for the sidebars on the `Grid` page - what do you think?

christinem:
@david-yz-liu I fixed the things that you mentioned. I want to also change the colors on the Check My POSt page because after the background was changed to white, it looks kind of funny now - but if you want to merge this beforehand I'll make a new PR for it.

hs/Css/Graph.hs
cchens: @christinem I think the `transform $ scaleX (-1)` was supposed to replace this entire block

david-yz-liu: Yes, that's correct - none of the others need to be there.

david-yz-liu:
@christinem There is one more change to make for this one. You can change the other colours in a separate pull request.

christinem:
@david-yz-liu Fixed! And alright, I'll make it in another PR.

david-yz-liu:
Thanks, @christinem!

PR #652, opened 2015-06-29
Refactor Haskell Code

Ian-Stewart-Binks:
This pull request minorly refactors some of the existing Haskell code. Meant as an incremental refactoring step.

Most of the changes that you will notice are indentation changes. Much of the code in the WebParsing module used an indentation style inconsistent with the rest of the codebase.

Some of the naming has been improved. You will see both files and variables renamed.

There have been other minor fixes as well.

PR #657, opened 2015-06-30
Issue 467

Ian-Stewart-Binks:
Lets native `ToJSON` functionality take over for `Session` and `Course` objects. Depends on #652.


david-yz-liu:
Nice work, @Ian-Stewart-Binks. Two comments:

1) I don't think this really depended on the previous pull request, and even if there were minor style incompatibilities, I would have preferred just seeing a pull request with the last commit.
2) I'm not entirely sure if we need a single type which contains all of a course information, as well as all of the timetable information at once. The `Courses` type we get for free from the schema.
3) Even if we do want a single type, I think it's better to create a wrapper around the `Courses` type.

Ian-Stewart-Binks:
> 2) I'm not entirely sure if we need a single type which contains all of a course information, as well as all of the timetable information at once. The Courses type we get for free from the schema.
3) Even if we do want a single type, I think it's better to create a wrapper around the Courses type.

There are inconsistencies with how we store data in the database and how we use it in the JavaScript/deliver it via JSON. For instance, a `Courses` has no idea which `Lectures` are associated with it. I agree, though.

PR #663, opened 2015-07-03
Fixes issues with grid sidebar scrolling

cchens:
There is a bug with the left sidebar of the grid where it will not scroll completely to the bottom if there is overflow. This fixes the bug.


cchens:
Might be a bit aesthetically weird in browsers that don't use webkit

Ian-Stewart-Binks:
What's with the new sidebar.png?

cchens:
@Ian-Stewart-Binks it's a compressed version of the old one that is honestly only in this PR because I didn't branch properly

hs/Css/Timetable.hs
```diff
@@ -178,13 +178,11 @@ courseSelectCSS = do
...
         color white
     "#course-select" ? do
-        padding0
-        margin (px 18) 0 0 (px 17)
-        width100
+        padding (px 18) 0 0 (px 17)
+        margin0
         height100
         alignCenter
         overflowY scroll
```
david-yz-liu: @cchens What do you think about making this `auto` rather than `scroll`?

cchens: @david-yz-liu the scrollbar adds to the width of the div (not sure if there's a way to fix it without Javascript), so the sidebar would constantly be resizing if someone is expanding/hiding a course repeatedly. If it's not an issue then I can change it to be `auto` instead

david-yz-liu: Does it always add to the width of the div? I thought it subtracted from the available width of the div contents.

cchens: I think it does. The contents of the div are displaced when the scrollbar appears.

david-yz-liu: Oh, yes. The inner contents will certainly change. I think that's okay.

david-yz-liu:
Thanks, @cchens!

PR #665, opened 2015-07-03
Stats graph repositioning

### LINKED ISSUE #662: Stats Graph is Cut Off
Similar to #551.

Photograph, look at top right:
screenshot from 2015-07-01 23 23 27

cchens:
Temporarily fixes #662, where a part of the legend was cut off


david-yz-liu:
@cchens We currently generate element-specific text-anchor attributes in the SVG region nodes; the problem with the stats graph is that even though the nodes have the correct attribute (`text-anchor="end"`), it's being overridden by the CSS (somehow, not sure why).

To fix this, we should remove the `"text-anchor" -: "start"` line from `regionCSS`. If you'd like to submit a pull request making that change, I'm happy to accept it.

PR #669, opened 2015-07-08
fast-db: Make db fast

Ian-Stewart-Binks:
Timetable parsing goes much faster now. Like caffeinated horse.

### DIFF
--- a/hs/WebParsing/TimeTableParser.hs
+++ b/hs/WebParsing/TimeTableParser.hs
@@ -14,6 +14,8 @@ import Database.CourseInsertion
 import WebParsing.HtmlTable
 import WebParsing.ParsingHelp
 import WebParsing.TimeConverter
+import Control.Monad.IO.Class (liftIO, MonadIO)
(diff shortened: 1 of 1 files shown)


hs/WebParsing/TimeTableParser.hs
```diff
@@ -198,4 +200,7 @@ parseTT = do
     rsp <- simpleHTTP (getRequest timetableUrl)
     body <- getResponseBody rsp
     let depts = getDeptList $ parseTags body
-    mapM_ getDeptTimetable depts
+
+    runSqlite databasePath $ do
+        runMigration migrateAll
```
david-yz-liu: So this seems to be the substantial change. Is a migration even necessary here if we're guaranteed to have performed it earlier?

Ian-Stewart-Binks: I don't believe so. I'll take it out.

david-yz-liu:
Thanks, @Ian-Stewart-Binks! This is a big win for such a (relatively) small change.

Issue #668, opened 2015-07-08
Remove Lecture/Tutorial datatypes in Tables.hs


Ian-Stewart-Binks:
Remove `Lecture` and replace it with the `Lectures` table datatype, same with `Tutorial`.

Ian-Stewart-Binks:
Currently working on this.

david-yz-liu:
Closed in #672

PR #672, opened 2015-07-09
Issue 668

Ian-Stewart-Binks:
Ref #668 
Replaces `Tutorial` and `Lecture` with Table datatypes.


public/js/common/objects/course.js
```diff
@@ -130,16 +130,10 @@ Course.prototype.parseTutorials = function (session, timeSuffix) {
...
 
     var tmp = this;
-    var tutorials = [];
-    var i;
-    for (i = 0; i < session.tutorials.length; i++) {
-        if (!inArray(session.tutorials[i], tutorials)) {
-            tutorials.push(session.tutorials[i]);
-        }
-    }
+    var tutorials = session.tutorials;
```
david-yz-liu: just use `session.tutorials` inline below

hs/WebParsing/TimeTableParser.hs
```diff
@@ -97,54 +99,61 @@ addLecture sesh lec = sesh {lectures = lec:(lectures sesh)}
...
 makeLecture slot =
-    Lecture { extra = 0,
-              section = (slotSection slot),
-              cap = 0,
-              time_str = (slotTime_str slot),
-              time = concatMap makeTimeSlots (T.split (== ' ') (slotTime_str slot)),
-              instructor = (slotInstructor slot),
-              enrol = Nothing,
-              wait = Nothing }
+    Lecture { lectureCode = slotCode slot,
```
david-yz-liu: Put the `Lecture {` on the line above

hs/Database/Tables.hs
```diff
@@ -55,25 +55,24 @@ Courses json
...
     deriving Show
 
-Lectures
+Lecture json
     code T.Text
     session T.Text
     section T.Text
-    times [Time]
-    capacity Int
+    time [Time]
```
david-yz-liu: `times`

hs/Database/Tables.hs
```diff
@@ -183,29 +162,19 @@ data Course =
...
-      Array $ V.fromList [toJSON (map convertTimeToString tutorialTimes), toJSON tutorialTimeStr]
-  toJSON (Tutorial (Just value) tutorialTimes tutorialTimeStr) =
-      Array $ V.fromList [toJSON value, toJSON (map convertTimeToString tutorialTimes), toJSON tutorialTimeStr]
+instance ToJSON Time where
+    toJSON (time) =
+        toJSON $ convertTimeToString time
+
+-- instance FromJSON required so that tables can be parsed into JSON,
+-- not necessary otherwise.
+instance FromJSON Time
```
david-yz-liu: Don't use generic here; use an inverse of `convertTimeToString`

Ian-Stewart-Binks: Hmm. Does it make much sense to 1) use a Double as the representation of a time, and 2) have this discrepancy between the way that we store data and the way that we serve that data - perhaps it would be better to make the conversion at the parsing level. What are your thoughts?

Ian-Stewart-Binks: I think it makes sense to store a _number_ as a _numeric_ type, however I think 2 is a bit less clear.

david-yz-liu: 1) Yes, neither the day of the week nor the time should be represented by Doubles. This is a separate issue, however.
2) It's not clear to me what you mean by "parsing level". If you're suggesting that we use the default (generic) FromJSON *and* ToJSON, and then parse the json in the Javascript when making the sections, that's fine with me, and probably better, too.

Ian-Stewart-Binks: By the "parsing level", I mean at the point in which the data is inserted into the database by the parser.

david-yz-liu: Okay, I'd rather not use the format returned by `convertTimeToString` to store the time information in the database.

Ian-Stewart-Binks: I have my own understanding why that wouldn't be the best idea, but I have no idea why you'd rather not. My guess would be that using a String/Text is too general a type to represent the time, and it makes us lose a bit of certainty about what exactly is in the database. However, making changes without clearly knowing the reason why I'm making the changes isn't very good in my opinion. I am now prompted to ask this question: Why would you rather not?

david-yz-liu: The loss of precision is part of it. But it also makes it harder to write queries about the times and days in which courses take place.

Really, we should be using the `PersistTimeOfDay` and `PersistDay` constructors to store this data.

Ian-Stewart-Binks: > Really, we should be using the `PersistTimeOfDay` and `PersistDay` constructors to store this data.
I'm not sure I understand. What do you mean?

david-yz-liu: Summary of discussion:

- Define new `Enum` instance `DayOfWeek`, and make it an instance of `toJSON` and `fromJSON`
- Change the `Time` type to be a record containing two fields, `day :: DayOfWeek` and `time :: TimeOfDay`
- We should then be able to use the generic `json` to handle `Time`

Javascript files will also need to be changed to handle the new json format, but overall everything should be cleaner.

david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #680, opened 2015-07-15
Specialist Extra Category fix on Check My POSt

christinem:
Linear Algebra should not be an option in the Extra category for specialists, since it is its own required category in the specialist. This fixes that.


david-yz-liu:
Thanks, @christinem!

PR #684, opened 2015-07-24
ICalendar Event Generator

GitGeminiano:
Dear Courseography members,

I have been working on the event generator for Google Calendar and ICalendar. This is a working version of my work. Please let me know if changes are needed.

Best Regards,
Christian  Garcia


Ian-Stewart-Binks:
@GitGeminiano Wondferful! Thanks for the contribution.

hs/Scripts.hs
```diff
@@ -100,5 +100,8 @@ searchScripts = do
          "https://cdnjs.cloudflare.com/ajax/libs/react/0.13.1/JSXTransformer.js"])
     H.script ! A.type_ "text/jsx" ! A.src "static/js/search/timetable.js" $ ""
 
+--calendarScripts :: H.Html
+--calendarScripts =
```
Ian-Stewart-Binks: Probably a good idea to remove this.

public/js/common/export/export.js
```diff
@@ -17,8 +17,11 @@ function createExportModalDiv() {
     var img = (context === 'graph') ? getGraphImage() : getGridImage(session);
     var contentDiv = $('<div></div>');
     var topContentDiv = $('<div></div>');
+    var calendarOption = $('<a href="http://localhost:8000/calendar">Download ICS</a>')
```
arkon: Should remove `http://localhost:8000`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Creates an event for each start/end time.
+eventsByTime :: String -> String -> [String] -> [String] -> (String, String) -> String -> [String]
+eventsByTime code sect start end date timeSystem = 
+    if fst date == "" || snd date == "" then []
+    else map (eventsByDate code sect date timeSystem) (zip start end)
+
+-- | Generates the string that represents the event for each course.
+eventsByDate :: String -> String -> (String, String) -> String -> (String, String) -> String
+eventsByDate code sect (startDate, endDate) timeSystem (start, end) = 
+    "BEGIN:VEVENT\n" ++ "DTSTAMP:" ++ timeSystem ++
```
david-yz-liu: I liked the earlier approach of storing the lines in a list of `String`. You can do that here and either return `[String]`, or use `unlines` to still return a `String`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Creates an event for each course.
+eventsByCourse :: String -> String -> String -> [[String]] -> [[String]] -> [[(String, String)]] -> String -> [[String]]
+eventsByCourse code session sect start end dates timeSystem = map (eventsByDay code session sect timeSystem) (zip' start end dates)
+
+-- | Join three lists together in tuples of three elements
+zip' :: [a] -> [b] -> [c] -> [(a,b,c)]
+zip' _ _ [] = []
+zip' _ [] _ = []
+zip' [] _ _ = []
+zip' (x:xs) (y2:ys) (z:zs) = (x,y2,z):zip' xs ys zs
```
david-yz-liu: If you put this line first, you'd only need one line after it: `zip' _ _ _ = []`

hs/DevelopmentConfig.hs
```diff
@@ -46,8 +52,30 @@ genCssPath = "../public/style/"
...
+
+-- | First Monday of the Winter term.
+firstMondayWinter :: Day
+firstMondayWinter = fromGregorian 2016 01 11
+
+-- | Last Monday of the Winter term.
+lastMondayWinter :: Day
+lastMondayWinter = fromGregorian 2016 04 04
+
+-- | Out of date day.
```
david-yz-liu: It's not very clear what this value is for.

hs/DevelopmentConfig.hs
```diff
@@ -46,8 +52,30 @@ genCssPath = "../public/style/"
...
+outDay :: Day
+outDay = fromGregorian 2014 01 01
+
 -- FACEBOOK CONFIGURATION
 
 -- | Enable Facebook integration. Should only be true on the production server.
 enableFb :: Bool
-enableFb = False
+enableFb = False
```
david-yz-liu: There should be a new line here.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- sequence $ map pullDatabase courseInfo
+    currentTime <- getCurrentTime
+    let timeSystem = getTime currentTime
+    let events = concat $ map (getEvent timeSystem) databaseInfo
+    return $ toResponse $ getCSV $ events
+
+-- | Generates a properly formatted current date and time.
+getTime :: UTCTime -> String
+getTime currentTime = date ++ "T" ++ hour ++ "Z"
```
david-yz-liu: I can't believe there isn't a way to do this using a built-in function, e.g. passing in a format string like `"%Y%m%dT%H%M%SZ"`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Returns a CSV file of events as requested by the user.
+calendarResponse :: String -> ServerPart Response
+calendarResponse courses =
+    liftIO $ getCalendar courses
+
+-- | Gets together all the pieces of the program.
+getCalendar :: String -> IO Response
+getCalendar courses = do
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- sequence $ map pullDatabase courseInfo
```
david-yz-liu: Look into `mapM` here

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    currentTime <- getCurrentTime
+    let timeSystem = getTime currentTime
+    let events = concat $ map (getEvent timeSystem) databaseInfo
```
david-yz-liu: You can use `fmap` here

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+getTime currentTime = date ++ "T" ++ hour ++ "Z"
+    where
+    date = formatTime defaultTimeLocale "%Y%m%d" (utctDay currentTime)
+    hour = formatTime defaultTimeLocale "%H%M%S" currentTime
+
+-- | Generates a string representing a CSV file.
+getCSV :: [String] -> String
+getCSV events =  if events == [] then "" else unlines $ header ++ events ++ bottom
+    where
+    header = ["BEGIN:VCALENDAR", "VERSION:2.0", "PRODID:-//Courseography//Calendar", "CALSCALE:GREGORIAN", "METHOD:PUBLISH"]
```
Ian-Stewart-Binks: Our formatting usually indents statements in a `where` block by 4 spaces.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    where
+    getData info = do
+        courseInfo <- info 
+        return (courseInfo, code, session, sect)
+
+-- ** Event Creation
+
+-- | Generates an event for each course.
+getEvent :: String -> ([Time], String, String, String) -> [String]
+getEvent timeSystem (timeFields, code, session, sect) = concat $ eventsByCourse code session sect start end dates timeSystem
```
Ian-Stewart-Binks: Some of the lines are a little long, and are difficult to read in some editors/github. A few one liners in this file are put onto the next line, it would be good to do that for some of the other ones as well.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    tuesday = let tues field = field !! 0 == 1.0 in filter tues lst
+    wednesday = let wed field = field !! 0 == 2.0 in filter wed lst
+    thursday = let thur field = field !! 0 == 3.0  in filter thur lst
+    friday = let fri field = field !! 0 == 4.0 in filter fri lst
+
+-- ** Start time
+
+-- | Obtains a list of start times for each course.
+startTime :: [[[Double]]] -> [[String]]
+startTime weekFields = map checkTimeStart weekFields
```
Ian-Stewart-Binks: I'm not entirely certain what this function does, or what `weekFields` is supposed to represent. Can you make this function a bit clearer?

Ian-Stewart-Binks: I think that you can Eta reduce here.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    halfWinter courseFields = getDate "S" courseFields
+    full session1 courseFields = getDate session1 courseFields
+
+-- | Gives the appropiate starting date for the course.
+-- | Obtains the starting date for each day.
+getDate :: String -> [[Double]] -> (String, String)
+getDate session dayFields = if session == "F" then getFallStr else getWinterStr
+    where
+    getFallStr = format $ getDayFall $ (concat dayFields) !! 0
+    getWinterStr = format $ getDayWinter $ (concat dayFields) !! 0
```
Ian-Stewart-Binks: I think these brackets are unnecessary.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Obtains all the dates for each course depending on its session.
+getDates :: String -> [[[Double]]] -> [[(String, String)]]
+getDates session dataInOrder = map (checkSession session) dataInOrder
+    where
+    checkSession session1 courseFields = if session1 == "Y" 
+                                        then [halfFall courseFields, halfWinter courseFields]
+                                        else [full session1 courseFields]
+    halfFall courseFields = getDate "F" courseFields
+    halfWinter courseFields = getDate "S" courseFields
+    full session1 courseFields = getDate session1 courseFields
```
Ian-Stewart-Binks: Can Eta reduce. Also, what is `full`?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    minutes = decimal * 6
+
+-- ** Start/End date
+
+-- | Obtains all the dates for each course depending on its session.
+getDates :: String -> [[[Double]]] -> [[(String, String)]]
+getDates session dataInOrder = map (checkSession session) dataInOrder
+    where
+    checkSession session1 courseFields = if session1 == "Y" 
+                                        then [halfFall courseFields, halfWinter courseFields]
```
Ian-Stewart-Binks: Indentation is a bit off.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+                  then hour ++ "0000"
+                  else hour ++ ratio minutes ++ "00"
+    where
+    hours = splitOn "." (show fullTime)
+    hour = hours !! 0
+    minutes = read (hours !! 1) :: Int
+
+-- | Gets the minutes out of a decimal part of my time.
+ratio :: Int -> String
+ratio decimal = if minutes >= 10 then show minutes else "0" ++ show minutes
```
Ian-Stewart-Binks: Can use `'0' : show minutes`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+checkTimeEnd :: [[Double]] -> [String]
+checkTimeEnd day = map getStr (map (+ 0.5) (getEndConsecutives sortedList ++ [last sortedList]))
+    where
+    sortedList = sort $ map (!! 1) day
+
+-- | Gets the end times that are not the very first end time. 
+getEndConsecutives :: [Double] -> [Double]
+getEndConsecutives lst = filter (/= 30.0) ([if lst !! i == (lst !! (i + 1)) - 0.5 then 30.0 else lst !! i| i <- [0 .. l]])
+    where
+    l = (length lst) - 2
```
Ian-Stewart-Binks: Brackets unnecessary.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+
+-- ** End time
+
+-- | Obtains a list of end times for each course.
+endTime :: [[[Double]]] -> [[String]]
+endTime weekFields = map checkTimeEnd weekFields
+
+-- | Gets all the end times for each day.
+checkTimeEnd :: [[Double]] -> [String]
+checkTimeEnd day = map getStr (map (+ 0.5) (getEndConsecutives sortedList ++ [last sortedList]))
```
Ian-Stewart-Binks: Can use `map (getStr . (+ 0.5)) (getEndConsecutives sortedList ++ [last sortedList])`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Generates a properly formatted current date and time.
+getTime :: UTCTime -> String
+getTime currentTime = date ++ "T" ++ hour ++ "Z"
+    where
+    date = formatTime defaultTimeLocale "%Y%m%d" (utctDay currentTime)
+    hour = formatTime defaultTimeLocale "%H%M%S" currentTime
+
+-- | Generates a string representing a CSV file.
+getCSV :: [String] -> String
+getCSV events =  if events == [] then "" else unlines $ header ++ events ++ bottom
```
Ian-Stewart-Binks: Can use `null` instead of `events == []`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+
+-- | Gets together all the pieces of the program.
+getCalendar :: String -> IO Response
+getCalendar courses = do
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- sequence $ map pullDatabase courseInfo
+    currentTime <- getCurrentTime
+    let timeSystem = getTime currentTime
+    let events = concat $ map (getEvent timeSystem) databaseInfo
+    return $ toResponse $ getCSV $ events
```
Ian-Stewart-Binks: Redundant `$`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+getStr :: Double -> String
+getStr fullTime = if minutes == 0
+                  then hour ++ "0000"
+                  else hour ++ ratio minutes ++ "00"
+    where
+    hours = splitOn "." (show fullTime)
+    hour = hours !! 0
+    minutes = read (hours !! 1) :: Int
+
+-- | Gets the minutes out of a decimal part of my time.
```
Ian-Stewart-Binks: What is `my time`.

hs/Database/CourseQueries.hs
```diff
@@ -28,6 +30,7 @@ import Data.String.Utils
 import Data.List
 import Config (databasePath)
 
+
```
Ian-Stewart-Binks: Remove blank line.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+-- | Obtains the starting date for each day.
+getDate :: String -> [[Double]] -> (String, String)
+getDate session dayFields = if session == "F" then getFallStr else getWinterStr
+    where
+    getFallStr = format $ getDayFall $ (concat dayFields) !! 0
+    getWinterStr = format $ getDayWinter $ (concat dayFields) !! 0
+
+-- | Formats the date in the following way: YearMonthDayT.
+-- For instance, 20150720T corresponds to July 20th, 2015. 
+
```
Ian-Stewart-Binks: Remove blank line.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    checkSession session1 courseFields = if session1 == "Y" 
+                                        then [halfFall courseFields, halfWinter courseFields]
+                                        else [full session1 courseFields]
+    halfFall courseFields = getDate "F" courseFields
+    halfWinter courseFields = getDate "S" courseFields
+    full session1 courseFields = getDate session1 courseFields
+
+-- | Gives the appropiate starting date for the course.
+-- | Obtains the starting date for each day.
+getDate :: String -> [[Double]] -> (String, String)
```
Ian-Stewart-Binks: Some of these functions can be given better names. Don't be afraid to give longer names, such that the longer name provides a reader with a quicker/deeper understanding of what the code does. In cases like this, it might also help to add information about the arguments (Some of the files in the Svg module are good examples). For instance, the documentation states something about a course, but I don't see how a course would be represented in the arguments. (This could also be a reflection on how Courseography should make better use of type synonyms.)

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+    where
+    checkSession session1 courseFields = if session1 == "Y" 
+                                        then [halfFall courseFields, halfWinter courseFields]
+                                        else [full session1 courseFields]
+    halfFall courseFields = getDate "F" courseFields
+    halfWinter courseFields = getDate "S" courseFields
+    full session1 courseFields = getDate session1 courseFields
+
+-- | Gives the appropiate starting date for the course.
+-- | Obtains the starting date for each day.
```
Ian-Stewart-Binks: This line has a pipe, which is inconsistent with the doc for `getStr`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+getEndConsecutives :: [Double] -> [Double]
+getEndConsecutives lst = filter (/= 30.0) ([if lst !! i == (lst !! (i + 1)) - 0.5 then 30.0 else lst !! i| i <- [0 .. l]])
+    where
+    l = (length lst) - 2
+
+-- ** Function that works for both start and end times
+
+-- | Creates the string time in the following way HourMinutesSeccondsZ.
+-- For instance, 133500Z corresponds to 1:35 pm.
+getStr :: Double -> String
```
Ian-Stewart-Binks: This function name is also a little ambiguous, and sticks out more.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+getDayFall _ = (outDay, outDay)
+
+-- | Gives the appropriate days for courses in the Winter.
+getDayWinter :: Double -> (Day, Day) 
+getDayWinter 0.0 = (firstMondayWinter, addDays 1 lastMondayWinter)
+getDayWinter 1.0 = (addDays 1 firstMondayWinter, addDays 2 lastMondayWinter)
+getDayWinter 2.0 = (addDays 2 firstMondayWinter, addDays 3 lastMondayWinter)
+getDayWinter 3.0 = (addDays 3 firstMondayWinter, addDays 4 lastMondayWinter)
+getDayWinter 4.0 = (addDays 4 firstMondayWinter, addDays 5 lastMondayWinter)
+getDayWinter _ = (outDay, outDay)
```
Ian-Stewart-Binks: Newline.

hs/Database/CourseQueries.hs
```diff
@@ -70,6 +73,30 @@ returnCourse lowerStr = runSqlite databasePath $ do
...
+-- returns a list of Time fields.
+returnLectureTimes :: T.Text -> T.Text -> T.Text -> IO [Time]
+returnLectureTimes lowerStr sect session = runSqlite databasePath $ do 
+    maybeEntityLectures <- selectFirst [LecturesCode ==. T.toUpper lowerStr,
+                                        LecturesSection ==. sect,
+                                        LecturesSession ==. session]
+                                       []
+    return $ maybe [] getLecture maybeEntityLectures
+    where
+    getLecture entity = lecturesTimes $ entityVal entity
```
Ian-Stewart-Binks: Indentation is off.

public/js/common/export/export.js
```diff
@@ -17,8 +17,11 @@ function createExportModalDiv() {
...
     contentDiv.attr('id', 'modal-content-container');
+    calendarOption.attr('target', '_blank');
     topContentDiv.html('<img id="post-image" src="data:image/png;base64,' + img + '" />');
+    contentDiv.append(calendarOption);
     contentDiv.append(topContentDiv);
```
Ian-Stewart-Binks: Due to jQuery's [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface), you may be able to 'chain' these calls.

hs/CalendarResponse.hs
Ian-Stewart-Binks: Also, too many spaces between `=` and `if`.

david-yz-liu: Use pattern matching here.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,230 @@
...
+orderData :: [Time] -> [[[Double]]]
+orderData courseFields = filter notNull (assignDay $ map timeField courseFields)
+    where
+    notNull field = not $ null field
+
+-- | Organizes the data by the day of the week, assigning a position based on the time field day.
+assignDay :: [[Double]] -> [[[Double]]]
+assignDay lst = [monday, tuesday, wednesday, thursday, friday]
+    where
+    monday = let mon field = field !! 0 == 0.0 in filter mon lst
```
Ian-Stewart-Binks: Might be more readable to use a lambda here.

hs/Scripts.hs
GitGeminiano: Yes, thank you.

hs/CalendarResponse.hs
GitGeminiano: I changed it. Thank you.

hs/DevelopmentConfig.hs
GitGeminiano: 'OutDay' is used to control forbidden inputs for days in the 'getDayFall' and 'getDayWinter' functions in CalendarResponse.hs. I will include more precise comments on this. Thank you.

hs/DevelopmentConfig.hs
GitGeminiano: I added one line at the end in all the files I changed. Thank you.

hs/CalendarResponse.hs
GitGeminiano: You are right. There is way. It looks like this: 'formatTime defaultTimeLocale "%Y%m%dT%H%M%SZ" currentTime'. Thank you for your comments.

hs/CalendarResponse.hs
GitGeminiano: If I try to use 'fmap' here I will end up using it three times. I think, it is a little bit clearer the way it is. I will be happy to change it if you think otherwise.

hs/CalendarResponse.hs
GitGeminiano: Thank you for all your comments. I will change the format for both the where statements and the one liners.

hs/CalendarResponse.hs
GitGeminiano: Yes, you are right. I do not need this function. Also, 'weekFields' is a little bit misleading. I meant the data I ordered by days. I will call it dataInOrder instead. Thank you.

hs/CalendarResponse.hs
GitGeminiano: Thank you. I will try not to overuse the brackets.

hs/CalendarResponse.hs
GitGeminiano: full is used for half year courses. When I get the information for full year courses from the database I get just the information for the Fall, since the information for the Winter is the same. Therefore, I have to produce events for both Fall and Winter with the same information. For this reason, I have to treat half and full year courses differently. I will change full for halfYearCourse to make it clearer.

public/js/common/export/export.js
GitGeminiano: I removed it. Thanks.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+getStrTime :: Double -> String
+getStrTime fullTime = if minutes == 0
+                  then hour ++ "0000"
+                  else hour ++ ratio minutes ++ "00"
+    where
+        hours = splitOn "." (show fullTime)
+        hour = hours !! 0
+        minutes = read $ hours !! 1 :: Int
+
+-- | Creates a string for the minutes.
```
Ian-Stewart-Binks: This function doc doesn't make much sense to me. The name can be improved too. Try to describe precisely what this function does in the name.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+        sortList dataDay = sort $ map (!! 1) dataDay
+
+-- | Gets the start times that are not the very first start time. 
+getStartConsecutives :: [Double] -> [Double]
+getStartConsecutives lst = 
+    filter (/= 30.0) ([if lst !! i == lst !! (i + 1) - 0.5
+                       then 30.0
+                       else lst !! (i + 1)| i <- [0 .. l]])
+    where
+        l = length lst - 2
```
Ian-Stewart-Binks: What is the significance of `length lst - 2`? A more appropriate name would drastically increase readability.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+          "CATEGORIES:EDUCATION", "END:VEVENT"]
+
+-- ** Ordering data
+
+-- | Orders the time fields for each course; which were obtained from the
+-- database
+orderData :: [Time] -> [[[Double]]]
+orderData courseFields = filter notNull (assignDay $ map timeField courseFields)
+    where
+        notNull field = not $ null field
```
Ian-Stewart-Binks: Is it possible to Eta reduce here? Perhaps using compose (`.`)?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+             [String], -- ^ End time
+             (String, String)) -- ^ Start/End date.
+             -> [String]
+eventsByDate code sect timeSystem (start, end, dates) =
+    (zip start end) >>= (eventsGenerator code sect timeSystem dates)
+
+-- | Generates the string that represents each event.
+eventsGenerator :: String -- ^ Course code.
+                -> String -- ^ Course section.
+                -> String -- ^ Current time.
```
Ian-Stewart-Binks: Why is the variable called `timeSystem`?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+
+-- | Obtains all the dates for each course depending on its session.
+getDatesByCourse :: String -- ^ Course session.
+                 -> [[[Double]]] -- ^ Time fields for the week ordered by day.
+                 -> [[(String, String)]]
+getDatesByCourse session dataInOrder =
+    if session == "Y"
+    then [map (\ dataByDay -> getDatesByDay "F" dataByDay) dataInOrder,
+          map (\ dataByDay -> getDatesByDay "S" dataByDay) dataInOrder]
+    else [map (\ dataByDay -> getDatesByDay session dataByDay) dataInOrder]
```
Ian-Stewart-Binks: No need for lambda.

Ian-Stewart-Binks: Same for above.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+        checkTimeEnd dataByDay =
+            map (getStrTime . (+ 0.5))
+                (getEndConsecutives (sortList dataByDay) ++
+                                     [last (sortList dataByDay)])
+        sortList dataDay = sort $ map (!! 1) dataDay
+
+-- | Gets the end times that are not the very first end time. 
+getEndConsecutives :: [Double] -> [Double]
+getEndConsecutives lst =
+    filter (/= 30.0) ([if lst !! i == lst !! (i + 1) - 0.5
```
Ian-Stewart-Binks: Redundant brackets.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+                       then 30.0
+                       else lst !! (i + 1)| i <- [0 .. l]])
+    where
+        l = length lst - 2
+
+-- ** End time
+
+-- | Obtains a list of end times for each course.
+endTime :: [[[Double]]] -> [[String]]
+endTime dataInOrder = map checkTimeEnd dataInOrder
```
Ian-Stewart-Binks: Can Eta reduce.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+    where
+        checkTimeStart dataByDay = 
+            map getStrTime ([head (sortList dataByDay)] ++
+                        getStartConsecutives (sortList dataByDay))
+        sortList dataDay = sort $ map (!! 1) dataDay
+
+-- | Gets the start times that are not the very first start time. 
+getStartConsecutives :: [Double] -> [Double]
+getStartConsecutives lst = 
+    filter (/= 30.0) ([if lst !! i == lst !! (i + 1) - 0.5
```
Ian-Stewart-Binks: Redundant bracket.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+        friday = filter (\field -> field !! 0 == 4.0) lst
+
+-- ** Start time
+
+-- | Obtains a list of start times for each course.
+startTime :: [[[Double]]] -> [[String]]
+startTime dataInOrder = map checkTimeStart dataInOrder
+    where
+        checkTimeStart dataByDay = 
+            map getStrTime ([head (sortList dataByDay)] ++
```
Ian-Stewart-Binks: Can use `:`.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+        tuesday = filter (\ field -> field !! 0 == 1.0) lst
+        wednesday = filter (\field -> field !! 0 == 2.0) lst
+        thursday = filter (\field -> field !! 0 == 3.0) lst
+        friday = filter (\field -> field !! 0 == 4.0) lst
+
+-- ** Start time
+
+-- | Obtains a list of start times for each course.
+startTime :: [[[Double]]] -> [[String]]
+startTime dataInOrder = map checkTimeStart dataInOrder
```
Ian-Stewart-Binks: Can Eta reduce.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+-- | Creates an event for each start/end time.
+eventsByDate :: String -- ^ Course code.
+             -> String -- ^ Course section.
+             -> String -- ^ Current time.
+             -> ([String], -- ^ Start time
+             [String], -- ^ End time
+             (String, String)) -- ^ Start/End date.
+             -> [String]
+eventsByDate code sect timeSystem (start, end, dates) =
+    (zip start end) >>= (eventsGenerator code sect timeSystem dates)
```
Ian-Stewart-Binks: Redundant brackets.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- mapM pullDatabase courseInfo
+    currentTime <- getCurrentTime
+    let timeSystem = getTime currentTime
+    let events = databaseInfo >>= (getEvent timeSystem)
+    return $ toResponse $ getICS events
+
+-- | Generates a properly formatted current date and time.
+getTime :: UTCTime -> String
+getTime currentTime = formatTime defaultTimeLocale "%Y%m%dT%H%M%SZ" currentTime
```
Ian-Stewart-Binks: Can Eta reduce.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+    liftIO $ getCalendar courses
+
+-- | Gets together all the pieces of the program.
+getCalendar :: String -> IO Response
+getCalendar courses = do
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- mapM pullDatabase courseInfo
+    currentTime <- getCurrentTime
+    let timeSystem = getTime currentTime
+    let events = databaseInfo >>= (getEvent timeSystem)
```
Ian-Stewart-Binks: Redundant brackets.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+             -> [[String]] -- ^ End times.
+             -> [(String, String)] -- ^ Start/End dates.
+             -> String -- ^ Current time.
+             -> [[String]]
+eventsByTime code sect start end dates timeSystem =
+    map (eventsByDate code sect timeSystem) (zip' start end dates)
+
+-- | Join three lists together in tuples of three elements
+zip' :: [a] -> [b] -> [c] -> [(a,b,c)]
+zip' (x:xs) (y2:ys) (z:zs) = (x,y2,z):zip' xs ys zs
```
Ian-Stewart-Binks: Why `y2`?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+    return $ toResponse $ getICS events
+
+-- | Generates a properly formatted current date and time.
+getTime :: UTCTime -> String
+getTime currentTime = formatTime defaultTimeLocale "%Y%m%dT%H%M%SZ" currentTime
+
+-- | Generates a string representing a ICS file.
+getICS :: [String] -> String
+getICS events =
+    if null events
```
Ian-Stewart-Binks: Can pattern match.

hs/CalendarResponse.hs
GitGeminiano: I called the variable timeSystem because the value to which it refers comes from calling the following function.
getCurrentTime :: IO UTCTimeSource
Get the current UTC time from the system clock.

Ian-Stewart-Binks: Do you think that this is a good name or a bad name, and why?

david-yz-liu: `systemTime` is probably a better name. You could change the description of this parameter to `Current system time.`

GitGeminiano: I will name it systemTime then. Thank you.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+orderData courseFields = filter notNull (assignDay $ map timeField courseFields)
+    where
+        notNull field = not $ null field
+
+-- | Organizes the data by the day of the week, assigning a position based on
+-- the day in the time field.
+assignDay :: [[Double]] -> [[[Double]]]
+assignDay lst = [monday, tuesday, wednesday, thursday, friday]
+    where
+        monday = filter (\ field -> field !! 0 == 0.0) lst
```
Ian-Stewart-Binks: Some of the lambdas have spaces, some don't. Courseography generally does not have spaces between the lambda and the argument.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+-- | Creates a string time in the following way HourMinutesSeccondsZ.
+-- For instance, 133500Z corresponds to 1:35 pm.
+getStrTime :: Double -> String
+getStrTime fullTime = if minutes == 0
+                  then hour ++ "0000"
+                  else hour ++ ratio minutes ++ "00"
+    where
+        hours = splitOn "." (show fullTime)
+        hour = hours !! 0
+        minutes = read $ hours !! 1 :: Int
```
Ian-Stewart-Binks: Use `readMaybe`.

hs/CalendarResponse.hs
david-yz-liu: Sorry @Ian-Stewart-Binks I didn't mean to step on your teachable moment. Go ahead!

hs/CalendarResponse.hs
GitGeminiano: The compiler was giving me an error because of an existing binding. However, I found a way around by importing only the functions I actually use in my program. Therefore I was able to change y2 for y1.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+
+import Data.List (sort)
+import Data.List.Split (splitOn)
+import Data.Time (Day, UTCTime, addDays, formatTime, getCurrentTime)
+import Happstack.Server (ServerPart, Response, toResponse)
+import Control.Monad.IO.Class (liftIO)
+import System.Locale
+import Database.CourseQueries (returnTutorialTimes, returnLectureTimes)
+import qualified Data.Text as T
+import Database.Tables as Tables
```
Ian-Stewart-Binks: Why `as Tables`?

hs/CalendarResponse.hs
Ian-Stewart-Binks: That's alright. I'll try again soon.

hs/CalendarResponse.hs
GitGeminiano: I will use 'import Database.Tables (Time, timeField)' instead. I imported Database.Tables as Tables because I thought that was the standard way for importing files that had to with the database. However, now that I think about it, such importing just adds readability in files such as courseQueries. Please let me know if you agree with the new way I am planning to use to import Database.Tables @Ian-Stewart-Binks.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+        bottom = ["END:VCALENDAR"]
+
+-- ** Cookie Information
+
+-- | Obtains a list with all the information about the courses obtained from
+-- the cookies.
+getCoursesInfo :: String -> [(String, String, String)]
+getCoursesInfo courses = map courseInfo allCourses
+    where
+        courseInfo course = (course !! 0, course !! 1, course !! 2)
```
david-yz-liu: Use pattern-matching to define `courseInfo`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+        allCourses = map (splitOn "-") byCourse
```
david-yz-liu: Combine this definition with the one below

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+        byCourse = splitOn "_" courses
+
+-- ** Database Information
+
+-- | Pulls out the information for each course from the database.
+pullDatabase :: (String, String, String) -> IO ([Time], String, String, String)
+pullDatabase (code, sect, session)
+    | take 1 sect == "L" = getData $ returnLectureTimes (T.pack code)
```
david-yz-liu: use `head`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+    concat $ eventsByCourse code session sect start end dates systemTime
+    where
+        dataInOrder = orderData timeFields
+        start = startTime dataInOrder
+        end = endTime dataInOrder
+        dates = getDatesByCourse session dataInOrder
+
+-- | Creates an event for each course depending on its session.
+eventsByCourse :: String -- ^ Course code.
+            -> String -- ^ Course session.
```
david-yz-liu: Indent to align the types

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+            -> String -- ^ Course section.
+            -> [[String]] -- ^ Start times.
+            -> [[String]] -- ^ End times.
+            -> [[(String, String)]] -- ^ Start/End dates.
+            -> String -- ^ Current system time.
+            -> [[String]]
+eventsByCourse code session sect start end dates systemTime
```
david-yz-liu: Use pattern-matching

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+zip' (x:xs) (y:ys) (z:zs) = (x,y,z):zip' xs ys zs
+zip' _ _ _ = []
+
+-- | Creates an event for each start/end time.
+eventsByDate :: String -- ^ Course code.
+             -> String -- ^ Course section.
+             -> String -- ^ Current system time.
+             -> [String] -- ^ Holidays dates.
+             -> ([String], -- ^ Start time
+             [String], -- ^ End time
```
david-yz-liu: align these types better

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+             (String, String)) -- ^ Start/End date.
+             -> [String]
+eventsByDate code sect systemTime holidays (start, end, dates) =
+    zip start end >>= eventsStr code sect systemTime dates holidays
```
david-yz-liu: This is a funny use of `>>=`, when you can just do `map`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+          -> String -- ^ Course section.
+          -> String -- ^ Current system time.
+          -> (String, String) -- ^ Start/End date.
+          -> [String] -- ^ Holidays dates.
+          -> (String, String) -- ^ Start/End time.
+          -> [String]
+eventsStr code sect systemTime (startDate, endDate) holidays (start, end)
+    | (startDate == "" || endDate == "") = []
+    | otherwise =
+        ["BEGIN:VEVENT", "DTSTAMP:" ++ systemTime,
```
david-yz-liu: Please put each one on a separate line.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+-- | Orders the time fields for each course; which were obtained from the
+-- database
+orderData :: [Time] -> [[[Double]]]
+orderData courseFields = filter (not . null)
+                                (assignDay $ map timeField courseFields)
+
+-- | Organizes the data by the day of the week, assigning a position based on
+-- the day in the time field.
+assignDay :: [[Double]] -> [[[Double]]]
+assignDay lst = [monday, tuesday, wednesday, thursday, friday]
```
david-yz-liu: This function is a little inelegant. Use `groupBy`, and `sortBy`

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,285 @@
...
+-- | Creates a string time in the following way HourMinutesSeccondsZ.
+-- For instance, 133500Z corresponds to 1:35 pm.
+getStrTime :: Double -> String
+getStrTime fullTime =
+                  if maybe True (== 0) minutes
+                  then hour ++ "0000"
+                  else hour ++ maybe "0000" getStrMinutes minutes ++ "00"
+    where
+        hours = splitOn "." (show fullTime)
+        hour = hours !! 0
```
david-yz-liu: In general, use `head`, not `!! 0`

Ian-Stewart-Binks: I disagree that `head` should be used in general. If the code uses `!!` to access elements of a list at different indices, then using `head` for the first element and `!! #` for the rest of the elements seems inconsistent. HLint typically warns against this, but in their [documentation](http://community.haskell.org/~ndm/darcs/hlint/hlint.htm), they comment on this topic:

> Warning - for example x !! 0 suggests head x as a "warning" severity hint. Typically head is a simpler way of expressing the first element of a list, especially if you are treating the list inductively. However, in the expression f (x !! 4) (x !! 0) (x !! 7), replacing the middle argument with head makes it harder to follow the pattern, and is probably a bad idea. Warning hints are often worthwhile, but should not be applied blindly.

This bit of code does indeed use `!! 1` on the subsequent line.

david-yz-liu: Okay, that's fine.

hs/CalendarResponse.hs
GitGeminiano: I changed it. However, I am using the information produced by this function to calculate both times and dates. For this reason I need to return [[[Double]]]. Also, the implementation using groupBy and sortBy looks cleaner this way. Thank you for showing these really nice functions to me.

hs/CalendarResponse.hs
GitGeminiano: Thank you both @david-yz-liu @Ian-Stewart-Binks.

Ian-Stewart-Binks:
Good work, @GitGeminiano! There are a few more things I'd like for us to work on with the code, which will hopefully be a valuable learning experience for the both of us.

Code readability is crucial. More readable code helps future developers produce better work, and allows them to become more familiar with the code at a quicker rate. There's quite a bit that can be done to improve the readability of the entire Courseography codebase, but let's try some new things here.

The first thing that I think we should focus on are function signatures. Function signatures can be used to communicate to a developer what exactly the parameters to a function are, and what the function returns. The more expressive the arguments in the signature are, the better. One of the nice things about Haskell is its types. It's possible to write better function signatures by defining more expressive types.

I'll write some comments on the existing code.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,284 @@
...
+                  "PRODID:-//Courseography//Calendar",
+                  "CALSCALE:GREGORIAN",
+                  "METHOD:PUBLISH"]
+        bottom = ["END:VCALENDAR"]
+
+-- ** Cookie Information
+
+-- | Obtains a list with all the information about the courses obtained from
+-- the cookies.
+getCoursesInfo :: String -> [(String, String, String)]
```
Ian-Stewart-Binks: The function signature here doesn't tell me what information about the courses I would be getting.

Suppose you had a function, such as:
```
getHeadChefInformation :: String -> [(String, String)]
```
What information are you getting about the head chefs?

It's possible to define new type synonyms, such as `ID` and `Name`, to make this more expressive.

Now, with the function:

```
getHeadChefInformation :: String -> [(Name, ID)]
```
I'll ask the same question: What information are you getting about the head chefs?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,284 @@
...
+                      else listTimes !! i| i <- [0 .. lenForComparison]]
+    where
+        lenForComparison = length listTimes - 2
+
+-- ** Function that works for both start and end times
+
+-- | Creates a string time in the following way HourMinutesSeccondsZ.
+-- For instance, 133500Z corresponds to 1:35 pm.
+getStrTime :: Double -> String
+getStrTime fullTime =
```
Ian-Stewart-Binks: I have two notes here:
1. Encoding the type of the input/output in a variable/function name is not always good, and generally should be avoided. If the type of the function changes, then so must the name.
2. There is a better way to define a function that gets the string representation of a type in Haskell- can you find which one it is? You will need to change the types to do this.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,284 @@
...
+    filter (/= 30.0) [if listTimes !! i == listTimes !! (i + 1) - 0.5
+                      then 30.0
+                      else listTimes !! (i + 1)| i <- [0 .. lenForComparison]]
+    where
+        lenForComparison = length listTimes - 2
+
+-- ** End time
+
+-- | Obtains the end times for each course by day.
+endTime :: [[[Double]]] -> [[String]]
```
Ian-Stewart-Binks: What is a `[[[Double]]]`?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,284 @@
...
+-- the day in the time field.
+assignDay :: [[Double]] -> [[[Double]]]
+assignDay lst = groupBy (\x y -> head x == head y) sortedList
+    where
+        sortedList = sortBy compare lst
+
+-- ** Start time
+
+-- | Obtains the start times for each course by day.
+startTime :: [[[Double]]] -> [[String]]
```
Ian-Stewart-Binks: Is this the same `[[[Double]]]` as the one below?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,284 @@
...
+eventsByDate code sect systemTime holidays (start, end, dates) =
+    concatMap (eventsStr code sect systemTime dates holidays) (zip start end)
+
+-- | Generates the string that represents each event.
+eventsStr :: String -- ^ Course code.
+          -> String -- ^ Course section.
+          -> String -- ^ Current system time.
+          -> (String, String) -- ^ Start/End date.
+          -> [String] -- ^ Holidays dates.
+          -> (String, String) -- ^ Start/End time.
```
Ian-Stewart-Binks: Many of the docs here can be replaced by type synonyms. Is it a good idea to do so, or is it a better idea to keep it as is?

hs/Database/CourseQueries.hs
```diff
@@ -70,6 +72,30 @@ returnCourse lowerStr = runSqlite databasePath $ do
...
+-- returns a list of Time fields.
+returnLectureTimes :: T.Text -> T.Text -> T.Text -> IO [Time]
+returnLectureTimes lowerStr sect session = runSqlite databasePath $ do 
+    maybeEntityLectures <- selectFirst [LecturesCode ==. T.toUpper lowerStr,
+                                        LecturesSection ==. sect,
+                                        LecturesSession ==. session]
+                                       []
+    return $ maybe [] getLecture maybeEntityLectures
+    where
+        getLecture entity = lecturesTimes $ entityVal entity
```
Ian-Stewart-Binks: This doesn't have to do with types: compose (`.`) may make this more elegant.

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,281 @@
...
+        courseInfo course = (course !! 0, course !! 1, course !! 2)
+        allCourses = map (splitOn "-") byCourse
+        byCourse = splitOn "_" courses
+
+-- ** Database Information
+
+-- | Pulls out the information for each course from the database.
+pullDatabase :: (String, String, String) -> IO ([Time], String, String, String)
+pullDatabase (code, sect, session) =
+    if take 1 sect == "L"
```
Ian-Stewart-Binks: You can use pattern matching here. I believe @david-yz-liu suggested this before.

hs/CalendarResponse.hs
GitGeminiano: Thank you @Ian-Stewart-Binks. I agree that creating type synonyms can help us gain a deeper understanding of the code. Moreover, I will be happy to work on writing new function signatures.
For this specific case getCoursesInfo :: String -> [(String, String, String)] could be replaced by getCoursesInfo :: String -> [(Code, Section, Session)].

hs/CalendarResponse.hs
GitGeminiano: Yes, it is the same.

hs/CalendarResponse.hs
GitGeminiano: [[[Double]]] is a list containing the time fields ordered by day. Each [[Double]] is a list that contains the time fields for just one day of the week. endTime :: [[[Double]]] -> [[String]] could be replaced by endTime :: TimeFieldsByDay -> EndTimesByDay

hs/CalendarResponse.hs
GitGeminiano: In my opinion, it does not hurt to have more information about the function. However, if we use type synonyms, we should avoid writing repetitive information in the comments.

GitGeminiano: I do not think that we need Type Synonyms here because I cannot think of any more meaningful information that we could add in the comments.

GitGeminiano: Thanks for all your comments. I added some new types. Please do let me know if changes are needed.

Ian-Stewart-Binks: > In my opinion, it does not hurt to have more information about the function. However, if we use type synonyms, we should avoid writing repetitive information in the comments.

Correct. Redundancy is usually bad.

> I do not think that we need Type Synonyms here because I cannot think of any more meaningful information that we could add in the comments.

A principle can be derived here: if we have the opportunity to say something in the comments or through more expressive code, which one is better?

Also, there is now a type called `Code`, which is a synonym for a `String`. This function uses a `String` to represent a code. This may be a good opportunity for a change.

Ian-Stewart-Binks:
@GitGeminiano Very good. The code is now much easier to read.

Is there a better way to organize type synonyms? Perhaps a convention that other Haskell libraries use?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,336 @@
...
+-- | The current date and time as obtained from the system.
+type SystemTime = String
+
+-- | A list of the information within the time fields
+-- ordered by day.
+type TimeFieldsByDay = [[TimeFieldInfo]]
+
+-- | The information contained in a time Field. The first
+-- Double represents the day and the second one the time.
+type TimeFieldInfo = [Double]
```
Ian-Stewart-Binks: The organization of some of these types feels a bit strange to read. Do you think that it's possible to clarify some of the types? Are some of these types more unnecessary than others?

hs/CalendarResponse.hs
```diff
@@ -0,0 +1,336 @@
...
+    liftIO $ getCalendar courses
+
+-- | Gets together all the pieces of the program.
+getCalendar :: String -> IO Response
+getCalendar courses = do
+    let courseInfo = getCoursesInfo courses
+    databaseInfo <- mapM pullDatabase courseInfo
+    currentTime <- getCurrentTime
+    let systemTime = formatTime defaultTimeLocale "%Y%m%dT%H%M%SZ" currentTime
+    let events = databaseInfo >>= getEvents systemTime
```
Ian-Stewart-Binks: The second `let` is unnecessary.

hs/CalendarResponse.hs
GitGeminiano: It seems that types which are primary means of expression are preferred over secondary means such as comments. I guess using primary means over secondary ones is the principle to be derived.

david-yz-liu:
@GitGeminiano I merged this in, but I squashed all of the commits, so it didn't register on Github. Thanks for your work!

FYI, I put in the two major updates to make in issues #687 and #688.

PR #689, opened 2015-08-20
Response dir

david-yz-liu:
Moved all Haskell response modules into a subdirectory.

@Ian-Stewart-Binks, while I do generally prefer component-based organization, I think we should work to make the response files thin enough that they're basically just serving content computed elsewhere, and this "elsewhere" can be organized by component. I also suspect that for beginners, it's helpful to put all the files which use a particular library (like blaze-html or persistent) in one place. Please let me know if you have any thoughts on this.


hs/Response.hs
```diff
@@ -0,0 +1,35 @@
...
+import Response.About
+import Response.Calendar
+import Response.Draw
+import Response.Graph
+import Response.Grid
+import Response.Image
+import Response.NotFound
+import Response.Post
+import Response.Privacy
+import Response.Search
```
Ian-Stewart-Binks: Newline.

hs/Response/README.md
```diff
@@ -0,0 +1,12 @@
+Responses
+=========
+
+This directory contains all of the Haskell files responsible for creating server responses.
+These modules should all be exported by `hs/Response.hs`.
+
+All of the responses contain a value `???response` of type `ServerPart Response`,
```
Ian-Stewart-Binks: Perhaps `???Response`.

Ian-Stewart-Binks:
@david-yz-liu I agree. It makes it much more digestible.

david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #690, opened 2015-08-21
Created a utilities directory and simplified some Blaze & Happstack code

david-yz-liu:
The new `hs/Util` directory is meant for modules which have general-purpose helpers that we'd like to use in Courseography. This should be used to separate domain-specific code from the rest.

I also took the opportunity to clean up the two files I put in there (especially removing some of the old Blaze templating code).


hs/MasterTemplate.hs
```diff
@@ -19,27 +20,30 @@ masterTemplate title headers body scripts =
             H.title (H.toHtml title)
             H.link ! A.rel "icon" ! A.type_ "image/png"
                    ! A.href "static/res/ico/favicon.png"
-            sequence_ $ headers ++ (map stylesheet [
+            sequence_ $ headers ++ (map toStylesheet [
```
Ian-Stewart-Binks: Good call.

PR #692, opened 2015-08-21
Start using Circle Continuous Integration

david-yz-liu:
We don't have any tests right now, but at the very least this will check if the Haskell component successfully compiles on the latest versions of the packages.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,16 @@
+machine:
+  ghc:
+    version: 7.8.3
+
+general:
+  build_dir: hs
+
+dependencies:
+  override:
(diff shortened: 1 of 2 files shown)


Ian-Stewart-Binks:
The `process` dependency version needs to be changed in the cabal file, that may be a reason why the build fails. Or could become a reason in the future.

david-yz-liu:
Thanks for pointing that out, @Ian-Stewart-Binks. I've changed the cabal version back -- CircleCI now has no difficulty with cabal v1.22, so either something has changed or you're correct. Either way, looks like things are good now!

PR #711, opened 2015-09-13
issue-277: Fix static directory access

### LINKED ISSUE #277: Static Directory Access
When retrieving files from a location such as `http://localhost:8000/graph/something-else`, the server attempts to grab files from the static directory `something-else/static/...` instead of `/static/...`. We want to grab from `/static/...`.

Ian-Stewart-Binks:
Fixes #277


david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #712, opened 2015-09-13
fix-database-doc: Fix 'Time' doc in Tables.hs

Ian-Stewart-Binks:
A while ago, I noticed that the documentation for the `Time` type is incorrect. This fixes it, and adds a TODO as well.


david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #715, opened 2015-09-20
Tests using HUnit (not working) and QuickCheck.

pkukulak:
Do not merge. Pushing so @david-yz-liu can comment on which testing module (i.e. HUnit V. QuickCheck) is more suited to our needs. Could possibly be both. Note that QuickCheck in particular is effective for testing function invariants, while HUnit is good at, well, unit testing.

### DIFF
--- a/hs/SvgTests/BuilderTests.hs
+++ b/hs/SvgTests/BuilderTests.hs
@@ -1,3 +1,10 @@
-module BuilderTests where
+module SvgTests.BuilderTests where
 
 import Svg.Builder
+import Test.QuickCheck
+
+prop1_sanitizeId str =
+    let sanitized = sanitizeId str
+        illegals  = map (\c -> not $ elem c ",()/<>% ") sanitized
+        allTrue = and illegals
+    in  allTrue
--- a/hs/SvgTests/BuilderTests.hs
+++ b/hs/SvgTests/BuilderTests.hs
@@ -1,0 +1,3 @@
+module BuilderTests where
+
+import Svg.Builder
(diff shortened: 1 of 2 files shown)


pkukulak:
After doing a bit more reading, I've come across the following guide for a common test framework. It seems like a straightforward method designed for apps of all shapes and sizes. Let me know what you guys think. I'd like to write a comprehensive set of tests of Courseography.

http://batterseapower.github.io/test-framework/

hs/SvgTests/BuilderTests.hs
```diff
@@ -0,0 +1,10 @@
+module SvgTests.BuilderTests where
+
+import Svg.Builder
+import Test.QuickCheck
+
+prop1_sanitizeId str =
+    let sanitized = sanitizeId str
+        illegals  = map (\c -> not $ elem c ",()/<>% ") sanitized
+        allTrue = and illegals
```
david-yz-liu: Note: use `all` instead of `and $ map ...`

david-yz-liu:
@pkukulak Okay I like this approach. The only thing that's really missing here is a `main` function that acts as the entry point for tests. 

In terms of structure, I think you should create something like

```
hs/
  Test/
    Tester.hs
    Svg/
      Builder.hs
      Parser.hs
```

where the directory structure of `Test/Svg` mimics the structure of the `Svg` folder, with `Tester.hs` being the main test runner.

pkukulak:
Gotcha. Thanks for the guidance. Knowing the structure will make it easier to approach this methodically.

PR #723, opened 2015-10-04
cleanup: Remove feedback form references and graph dragging code

Ian-Stewart-Binks:
* `nodes` removed. Declared twice: https://github.com/Courseography/courseography/blob/f80e559b3240169b53ba3c0a06bed85d9e17e453/public/js/graph/setup.js#L22
* Feedback form references removed. They no longer exist.
* Graph dragging removed. It's old, bulky, intimidating, not in use, and exists in the history. The commit message explicitly says 'graph dragging', so it's searchable if we ever need it again.


david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #725, opened 2015-10-06
Conflict dialog fades after 3 seconds

### LINKED ISSUE #724: Grid conflict overlay blocks search box
When conflicting courses are selected in the grid, the warning overlay prevents the user from searching for more courses until one of the conflicting courses is removed.

pkukulak:
This fixes #724 .

### DIFF
--- a/public/js/grid/mouse_events.js
+++ b/public/js/grid/mouse_events.js
@@ -74,7 +74,7 @@ function alertUserOfConflict() {
     'use strict';
 
     var dialogSelector = $('#dialog');
-    getInConflict() ? dialogSelector.fadeIn(750) :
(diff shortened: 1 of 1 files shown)


Ian-Stewart-Binks:
I would be cool scrapping it. What are your thoughts @pkukulak?

pkukulak:
I kind of agree. It's always nice to have a "don't tell us we didn't say so" disclaimer, but the red cells are already enough of an indication that there's a conflict. Reducing clutter is always a good thing.

Ian-Stewart-Binks:
Yeah, I think that the red cells probably do the job. Clutter reduction would be nice.

david-yz-liu:
Thanks, @pkukulak. I agree with the discussion here, and removed the conflict dialog.

PR #726, opened 2015-10-06
PDF Exporting WIP

pkukulak:
@david-yz-liu : .tex files are generated with timetable images in them. They are then compiled into .pdf, where the only remaining step is to serve them to the client. This is not the best way to do this. It would be easier and cleaner to store timetable info in cookies. The modal is yet to be tabbed for selection between graph/grid exporting. and Facebook exporting

### DIFF
--- a/hs/ImageConversion.hs
+++ b/hs/ImageConversion.hs
@@ -1,8 +1,28 @@
 module ImageConversion
-    (createImageFile, removeImage) where
+    (createImageFile, removeImage, compileTexToPdf) where
 
 import System.Process
 import GHC.IO.Handle.Types
+
+-- | Compiles a TeX file to a PDF file.
+compileTexToPdf :: String -> IO
+                      (Maybe Handle,
+                       Maybe Handle,
+                       Maybe Handle,
+                       ProcessHandle)
+
+compileTexToPdf inName = createProcess $ CreateProcess
+                                   (ShellCommand $ "pdflatex " ++
+                                                   inName
+                                   )
+                                   Nothing
+                                   Nothing
+                                   CreatePipe
+                                   CreatePipe
+                                   CreatePipe
+                                   False
+                                   False
+                                   False
 
 -- | Converts an SVG file to a PNG file. Note that image magik's 'convert' command
 -- can take in file descriptors.
--- a/hs/Latex.hs
+++ b/hs/Latex.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Latex where
+
+import Text.LaTeX
+import Text.LaTeX.Packages.Graphicx
+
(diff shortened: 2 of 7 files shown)


david-yz-liu:
You need to add a route to `Server.hs`

hs/ImageConversion.hs
```diff
@@ -1,9 +1,29 @@
...
 import GHC.IO.Handle.Types
 
+-- | Compiles a TeX file to a PDF file.
+compileTexToPdf :: String -> IO
+                      (Maybe Handle,
+                       Maybe Handle,
+                       Maybe Handle,
+                       ProcessHandle)
+
+compileTexToPdf inName = createProcess $ CreateProcess
```
Ian-Stewart-Binks: Perhaps the filename should be changed to something that accounts for non-image-converting functions.

hs/Latex.hs
```diff
@@ -0,0 +1,36 @@
...
+    usepackage [] graphicx
+
+body imageFilename = do
+    addImage imageFilename
+
+addImage imageFilename =
+    includegraphics fitImageToPage imageFilename
+
+
+
```
Ian-Stewart-Binks: Lots of lines.

public/js/common/export/export.js
```diff
@@ -17,11 +17,16 @@ function createExportModalDiv() {
     var img = (context === 'graph') ? getGraphImage() : getGridImage(session);
     var contentDiv = $('<div></div>');
     var topContentDiv = $('<div></div>');
-    var calendarOption = $('<a href="calendar">Download ICS</a>')
+    var calendarOption = $('<a href="calendar">Download ICS</a>');
+    var gridOption = $('<a href="pdf">Export to PDF</a><br>');
```
david-yz-liu: This URL has no query parameters, but the server expects some. You can generate them yourself and put them in the URL here if you want - just repeat the code in `getGridImage` and call `encodeURIComponent` to encode them to create a URL.

hs/Response/Export.hs
```diff
@@ -0,0 +1,45 @@
...
+
+returnPdfData :: String -> String -> String -> String -> IO Response
+returnPdfData svgFilename imageFilename pdfFilename texFilename = do
+    createImageFile svgFilename imageFilename
+    compileTex texFilename imageFilename
+    _ <- compileTexToPdf texFilename
+    pdfData <- BS.readFile texFilename
+    _ <- removeImage svgFilename
+    _ <- removeImage imageFilename
+    _ <- removeImage pdfFilename
```
david-yz-liu: I would remove this for now until you're sure the PDF generation is working.

david-yz-liu:
@pkukulak I don't care what the "best way" to do this is right now. I want to see this working from the client side in any way possible before our next meeting.

PR #731, opened 2015-10-06
Added Estonian graph

tamaralipowski:
This Estonian graph won't parse properly.


david-yz-liu:
@tamaralipowski Please add the yEd .graphml file to this pull request (you can just commit the file and "Publish").

cchens:
@tamaralipowski I suspect that the problem is with creating the file and not with the parsing. I messed with the .svg file a bit in Inkscape and have been able to get this:
Imgur
Try doing the following:
- Select the regions, and click `Path > Object to Path`
- Select the text labels, and click `Text > Convert to Text`

Hope it works for you!

PR #736, opened 2015-10-12
Added Graphs - Parsing Problems

tamaralipowski:
Added Estonian, Italian, Finnish, and German graphs. None parse properly.

### DIFF
--- a/public/res/graphs/est2015.svg
+++ b/public/res/graphs/est2015.svg
@@ -1,0 +1,513 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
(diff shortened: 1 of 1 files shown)


cchens:
I made a comment in #731 about why your graphs might not be parsing correctly. Also, it seems that you've only committed the Estonian graph.

david-yz-liu:
@tamaralipowski - you don't actually need to close the pull request. After you make changes to the graph, you can commit them and then "Publish" - your pull request will automatically update, as long as you're on the right branch when you commit them.

tamaralipowski:
@david-yz-liu Thanks! I finally figured out what I was doing wrong on the app - hopefully I won't have problems with this anymore! I've submitted a pull request with the updated graphs after taking @cchens advice.

PR #737, opened 2015-10-13
Added Graphs - Inkscape files updated

tamaralipowski:
Added Estonian, Finnish, German, and Italian graphs after updating the Inkscape files to fix the parsing problems.


david-yz-liu:
@tamaralipowski Okay, the graphs are looking better, but still not perfect. Please make the following changes to all graphs you have created so far:

1) In the yEd file, make all the nodes have solid edges, not dotted edges.
2) In Inkscape: You seem to have converted all of the course nodes to paths. This isn't good, they need to be rectangles. Make sure that all of the *background coloured regions* (like "Independent Study") are paths, but make sure all of the individual nodes (like "EST100") are rectangles. You need to individually click on each region, don't just select all of the objects.

david-yz-liu:
@tamaralipowski Good progress. Can you please make the following changes:

1) Rather use the "OR" (and "AND") nodes you've been using, can you replace them with the style used in the CSC graph?
2) Can you fix the region boundaries in the Italian graph?

david-yz-liu:
@tamaralipowski Great!

I fixed two things, one of which I'll tell you about in person, and one which I can tell you right now:

- please make sure there's > 10px or so space between region labels and the course nodes. Otherwise the labels will be parsed as being part of the course nodes

PR #742, opened 2015-10-20
Remove infoPanel from Grid.hs

cchens:
Also cleans up any lingering code related to the infoPanel CSS

### DIFF
--- a/hs/Css/Timetable.hs
+++ b/hs/Css/Timetable.hs
@@ -19,7 +19,6 @@ timetableStyles = do
     timetableCSS
     courseSelectCSS
     tdColours
-    --infoCSS
 
 {- searchCss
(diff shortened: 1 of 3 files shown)


cchens:
Never mind, I figured out what it does. `infoPanel` was a terrible name in my opinion, though.

Ian-Stewart-Binks:
@cchens Yes, the naming is awful. This is something that I was bad at when starting Courseography development, and something that I think I could work on a lot. Sorry for any confusion. 

If ever you see something that could be renamed, I encourage you to rename it.

Btw, what is it for?

cchens:
@Ian-Stewart-Binks It's all good. I think it's a wrapper of sorts for the grid part in the middle, as the grid doesn't render properly without it.

PR #743, opened 2015-10-26
Specialist POSt React Component

christinem:
This PR has a visually-complete Specialist React component implemented. Functionality is half complete. The only issue I have is for MAT137 and MAT157 - apparently calling getCourse() on either of those courses just returns an object with a lot of nulls, so getCourseTitle() doesn't work - not sure if they are missing from the database or not. The code can use some cleaning up.

### DIFF
--- a/public/js/post/post.js
+++ b/public/js/post/post.js
@@ -1,6 +1,9 @@
 var CourseCode = React.createClass({
     getInitialState: function() {
-        return {selected: false}
+        return {
+            selected: false,
+            infoOpened: false
+        }
     },
 
     componentWillMount: function() {
@@ -8,7 +11,7 @@ var CourseCode = React.createClass({
     },
 
     toggleFullInfo: function() {
-        $('#' + this.props.courseIDs[0] + '_info').toggle();
+        this.setState({infoOpened: !this.state.infoOpened});
     },
 
     getCategoryName: function() {
@@ -32,16 +35,21 @@ var CourseCode = React.createClass({
 
         var categoryClasses = 'code';
         var courseClasses = 'full_name';
+        var infoClasses = 'more-info';
 
         if (this.state.selected) {
             categoryClasses += ' category_selected';
             courseClasses += ' course_selected';
         }
 
+        if (this.state.infoOpened) {
+            infoClasses += ' info_opened'
+        }
+
         return (
             <div id ={this.props.courseIDs[0]} className='course'>
                 <p className={categoryClasses} onClick={this.toggleFullInfo}> {this.getCategoryName()} </p>
-                <div id= {this.props.courseIDs[0] + '_info'} className='more-info'>
+                <div id = {this.props.courseIDs[0] + '_info'} className={infoClasses}>
                     {this.props.courseIDs.map(function (course) {
                          return <p className={courseClasses}> {getCourseTitle(course)} </p>
                     })}
--- a/public/js/post/post.js
+++ b/public/js/post/post.js
@@ -16,7 +16,7 @@ var CourseCode = React.createClass({
 
         if (this.props.courseIDs[0] === 'mat135') {
             // special case for calculus requirement since it doesn't fit the same pattern
-            categoryName = '(MAT135 and MAT136) or MAT137Y or MAT157Y';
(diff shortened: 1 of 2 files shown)


public/js/post/post.js
```diff
@@ -4,11 +4,28 @@ var CourseCode = React.createClass({
     },
 
     componentWillMount: function() {
-        this.setState({selected: getCookie(this.props.courseID) === 'active'});
+        this.setState({selected: getCookie(this.props.courseIDs[0]) === 'active'});
     },
 
     toggleFullInfo: function() {
-        $('#' + this.props.courseID + '_info').toggle();
+        $('#' + this.props.courseIDs[0] + '_info').toggle();
```
arkon: You should just use the component's state to handle the visibility of the element, instead of relying on DOM manipulation.

public/js/post/post.js
```diff
@@ -4,11 +4,28 @@ var CourseCode = React.createClass({
...
+        var categoryName = '';
+
+        if (this.props.courseIDs[0] === 'mat135') {
+            // special case for calculus requirement since it doesn't fit the same pattern
+            categoryName = '(MAT135H and MAT136H) or MAT137Y or MAT157Y';
+        } else {
+            categoryName += this.props.courseIDs[0].toUpperCase() + 'H';
+
+            for (i = 1; i < this.props.courseIDs.length; i++) {
+                categoryName += " or " + this.props.courseIDs[i].toUpperCase() + 'H';
```
david-yz-liu: 1) Maybe start by doing a map onto `this.props.courseIDs` to convert to uppercase and add an 'H'.
2) Use the Javascript `join` method for arrays.

public/js/post/post.js
```diff
@@ -22,15 +39,121 @@ var CourseCode = React.createClass({
...
+        }
+
+        var me = this;
+
+        return (
+            <div id={this.props.courseID} className='course'>
+                <p className = {categoryClasses} onClick={this.toggleFullInfo}> {this.props.data.categoryName} </p>
+                <div id = {'spec' + this.props.courseID.substring(5, this.props.courseID.length)} className='more-info'>
+                    <p className = {courseClasses}> 
+                        {Array.apply(0, Array(this.props.data.textBoxNumber)).map(function (x, i) {
```
david-yz-liu: @christinem I thought we talked about this and saw that the `Array` constructor and `apply` methods weren't actually needed here?

public/js/post/post.js
```diff
@@ -22,15 +39,121 @@ var CourseCode = React.createClass({
...
+            this.checkIfCompleted();
+        }  
+    },
+
+    render: function() {
+
+        var categoryClasses = 'code';
+        var courseClasses = 'full_name';
+
+        if (this.state.completed) {
```
david-yz-liu: Can you change the CSS selector so that you can just make the outer div have a class of `selected` rather than apply it to both inner parts?

public/js/post/post.js
christinem: @david-yz-liu we talked about this last meeting and I think we agreed that this was actually alright.

christinem:
@david-yz-liu I made the requested changes!

public/js/post/post.js
```diff
@@ -1,36 +1,168 @@
...
+    },
+
+    checkIfCompleted: function() {
+       this.setState({completed: this.state.completedTextBoxes === this.props.data.textBoxNumber});
+    },
+
+    handleKeyDown: function(e) {
+        if (e.keyCode === 13) {
+            if (this.state.completedTextBoxes < this.props.data.textBoxNumber) {
+                 this.setState({completedTextBoxes: this.state.completedTextBoxes +=1})
```
david-yz-liu: space missing at the end, and this line is indented by 1 too many spaces

public/js/post/post.js
```diff
@@ -1,36 +1,168 @@
...
+        }  
+    },
+
+    render: function() {
+
+        var classes = 'course';
+        var infoClasses = 'more-info';
+
+        if (this.state.completed) {
+            classes += " .selected";
```
david-yz-liu: shouldn't have a period

david-yz-liu:
Thanks, @christinem. I merged this in, although I left a few minor comments.

PR #747, opened 2015-11-02
Finish specialist textbox functionality

christinem:
Functionality for textboxes for MultipleCourseCode component is now about complete. Textboxes autofill based on selected courses, and you can type in courses for the third (extra) category. The category will detect when all it's textboxes are filled (therefore complete).

### DIFF
--- a/public/js/post/post.js
+++ b/public/js/post/post.js
@@ -129,6 +129,7 @@ var SpecialistPost = React.createClass({
         var level400Courses = [];
         var level300Courses = [];
         var levelExtraCourses = [];
+        var inquiryCourse = [];
 
         this.state.activeCourses.map(function (course) {
             if (notSpecialistCourse(course)) {
@@ -139,10 +140,14 @@ var SpecialistPost = React.createClass({
                 } else if (course.substring(3, 4) >= '3' && levelExtraCourses.length < 4) {
                     levelExtraCourses.push(course);
                 }
+
+                if (CSCinq.indexOf(course) >= 0 && inquiryCourse.length < 1) {
+                    inquiryCourse.push(course);
+                }
             }
         });
 
-        return [level400Courses, level300Courses, levelExtraCourses];
+        return [level400Courses, level300Courses, levelExtraCourses, inquiryCourse];
     },
 
     render: function() {
@@ -176,7 +181,7 @@ var SpecialistPost = React.createClass({
                     categoryName: 'Any of the following: 300+ level CSC course; MAT: 235/237/257, any 300+ \
                                      except for 329, 390, & 391; STA: 248, 261, any 300+; ECE: 385H/489H; \
                                      BCB: 410H/420H/430Y (2.0 FCEs)'}} />
-                <MultipleCourseCode courseID="spec_inq" data={{textBoxNumber: 1, 
+                <MultipleCourseCode courseID="spec_inq" data={{textBoxNumber: 1, courses: courseCategoryArrays[3],
                     categoryName: 'Any from this list: CSC301H, CSC318H, CSC404H, CSC411H, CSC418H, CSC420H, \
                     CSC428H, CSC454H, CSC485H, CSC490H, CSC491H, CSC494H, or PEY (0.5 FCEs) \
                     ** Note: Type "PEY" for Check my POSt to recognize it **'}} />
--- a/public/js/post/post.js
+++ b/public/js/post/post.js
@@ -108,7 +108,8 @@ var MultipleCourseCode = React.createClass({
 var SpecialistPost = React.createClass({
     getInitialState: function() {
         return {
-            selected: true
+            selected: true,
+            activeCourses: activeCourses.slice()
         }
     },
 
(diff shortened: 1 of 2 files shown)


public/js/post/post.js
```diff
@@ -63,47 +63,59 @@ var MultipleCourseCode = React.createClass({
             infoOpened: false
         }
     },
+
+    componentDidMount: function() {
+        for (i = 0; i < this.props.data.courses.length; i++) {
+            this.setState({completedTextBoxes: this.state.completedTextBoxes += 1});
```
david-yz-liu: Use ` + 1` not `+= 1`

public/js/post/post.js
```diff
@@ -124,6 +137,36 @@ var SpecialistPost = React.createClass({
...
 
+    getCourses: function () {
+        var level400Courses = [];
+        var level300Courses = [];
+        var levelExtraCourses = [];
+        var inquiryCourse = [];
+
+        this.state.activeCourses.map(function (course) {
+            if (notSpecialistCourse(course)) {
+                if (course.substring(3, 4) === '4' && level400Courses.length < 3) {
```
david-yz-liu: Better, but I want to see each of these checks be a separate function.

public/js/post/post.js
arkon: Why is this in a loop anyway? Seems like you could just do something like
```JavaScript
this.setState({
    completedTextBoxes: this.state.completedTextBoxes + this.props.data.courses.length
});
```

christinem: @arkon I actually tried that. For some really strange reason, when I try this, `this.state.completedTextboxes` is still 0 right after, so the `this.checkIfCompleted()` call right after it returns false even when it's really true. However when I check the state in Chrome React tools after it's all been rendered, `this.state.completedTextboxes` has properly changed. I can't figure out why that is. Somehow the for loop causes the state to update automatically.

arkon: @christinem Ah. That's because `setState` is asynchronous. Try:
```JavaScript
this.setState({
    completedTextBoxes: this.state.completedTextBoxes + this.props.data.courses.length
}, this.checkIfCompleted);
```

david-yz-liu: Thanks, @arkon!

christinem: Yes, thanks for the advice @arkon! That did fix the problem.

christinem:
@david-yz-liu I made changes to how the course arrays are populated based on what we discussed two weeks ago. Is this what you had in mind when you were discussing it with me?

public/js/post/post.js
```diff
@@ -63,47 +63,57 @@ var MultipleCourseCode = React.createClass({
...
+        this.setState({completed: this.state.completedTextBoxes === this.props.data.textBoxNumber});
     },
 
     handleKeyDown: function(e) {
         if (e.keyCode === 13) {
-            if (this.state.completedTextBoxes < this.props.data.textBoxNumber) {
-                 this.setState({completedTextBoxes: this.state.completedTextBoxes +=1})
+            if (this.state.completedTextBoxes <= this.props.data.textBoxNumber + 1) {
+                if (e.target.defaultValue === '' && e.target.value !== '') {
+                    this.setState({completedTextBoxes: this.state.completedTextBoxes += 1});
```
david-yz-liu: Still not sure why `+=` is used here.

public/js/post/post.js
```diff
@@ -116,7 +126,8 @@ var MultipleCourseCode = React.createClass({
 var SpecialistPost = React.createClass({
     getInitialState: function() {
         return {
-            selected: true
+            selected: true,
+            activeCourses: activeCourses.slice()
```
david-yz-liu: I'm not sure what the `activeCourses` on the right is. It seems like this would be a good place to call the function which computes which courses are active.

public/js/post/post.js
```diff
@@ -124,6 +135,44 @@ var SpecialistPost = React.createClass({
...
+    },
+
+    getCourses: function () {
+        // [level400Courses, level300Courses, levelExtraCourses, inquiryCourse]
+        var courseArrays = [[], [], [], []];
+        var courseChecks = [this.isLevel400, this.isLevel300, this.isLevelExtra];
+        var me = this;
+
+        this.state.activeCourses.map(function (course) {
+            for (var i = 0; i < 3; i++) {
```
david-yz-liu: Use `courseChecks.length`

public/js/post/post.js
```diff
@@ -124,6 +135,44 @@ var SpecialistPost = React.createClass({
...
+        return notSpecialistCourse(course) && course.substring(3, 4) >= '3' && levelExtraArray.length < 4;
+    },
+
+    isInquiryCourse: function(course, inquiryArray) {
+        return CSCinq.indexOf(course) >= 0 && inquiryArray < 1;
+    },
+
+    getCourses: function () {
+        // [level400Courses, level300Courses, levelExtraCourses, inquiryCourse]
+        var courseArrays = [[], [], [], []];
```
david-yz-liu: Use a loop (or something) to create this dyanmically, so that the length isn't hard-coded to 3.

public/js/post/post.js
christinem: Oops, different part of the code than the last time, but still same error. Thanks!

PR #748, opened 2015-11-05
svg and defs attributes added into react-graph

ryanfan:
`(#arrow)` renders properly now


public/js/graph/parse_graph.js
```diff
@@ -90,14 +90,16 @@ function buildGraph(gId) {
     });
 
     $('.path').each(function () {
-        if ($(this).attr('source-node') && $(this).attr('target-node')) {
-            makeEdge(window[$(this).attr('source-node')],
-                     window[$(this).attr('target-node')],
+        if ($(this).attr('data-source-node') && $(this).attr('data-target-node')) {
```
arkon: You _could_ just forego jQuery and do `this.dataset.sourceNode`, for example.

Ian-Stewart-Binks:
@ryanfan On your last pull request, I suggested some style changes. It appears that some of these style changes haven't been applied across the PR, and some suggestions were ignored.

Before making PRs in the future, for whatever repository you may be contributing to, it's a good idea to review your PR for style and clarity before submitting it. This way, reviewers can review the code without having to get caught up in the style changes. For instance, this pull request tries to pull in some redundant whitespace, which I'm assuming is not the intention of the PR.

Can you fix the style of the PR such that it is consistent with the rest of Courseography? Can you remove the comment about moving the setup code to postsetup.js.jsx?

Also, I'm a bit out of the loop. Why is this being merged into master and not react-graph?

Thanks.

ryanfan:
@Ian-Stewart-Binks Ok, I'll fix these things, close this PR and make another.

PR #751, opened 2015-11-18
Search bar in Grid done in React

cchens:
The React components for the search functionality of the Grid page. Does not support onclick functionality. Yet.


public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
+var CoursePanel = React.createClass({
+    render: function() {
+        return (
+        	<ul className="trapScroll-enabled" id="course-select">
```
Ian-Stewart-Binks: Tabs => Spaces

public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
...
+        );
+    }
+});
+
+var SearchPanel = React.createClass({
+    render: function() {
+        return (
+            <div>
+                <div id="filter-container">
+                    <form onsubmit="return false;">
```
david-yz-liu: Hmm okay since we're cleaning up, can we remove the outer "form" element?

public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
...
+    getInitialState: function() {
+        return {
+            courses: [],
+            searchList: []
+        };
+    },
+
+    componentDidMount: function() {
+        $.ajax({
+            url: "all-courses",
```
david-yz-liu: Try to prefer single quotes.

public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
...
+        var filter = $('#course-filter').val().toUpperCase();
+
+        if (filter !== '') {
+            var searchList = this.state.courses.filter(function (course) {
+                return course.indexOf(filter) > -1;
+            }).map(function (course) {
+                return <CourseEntry course={course} />
+            });
+        }
+        this.setState({searchList: searchList});
```
david-yz-liu: I'd rather see the state only keep track of the value of the the input (#course-filter), and the filtering of courses and generation of <CourseEntry> components happen inside the render function.

public/js/grid/react_util.js
cchens: @david-yz-liu there's some CSS tied to the "form" element, so removing it would require some modification to existing CSS. Should I do that?

david-yz-liu: Yeah sure

public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
...
+            courses: [],
+            searchList: []
+        };
+    },
+
+    componentDidMount: function() {
+        $.ajax({
+            url: "all-courses",
+            dataType: "text",
+            async: false,
```
arkon: Why isn't this asynchronous?

public/js/grid/react_util.js
```diff
@@ -0,0 +1,115 @@
...
+        });
+    },
+
+    resetSearchList: function() {
+        'use strict';
+
+        var filter = $('#course-filter').val().toUpperCase();
+
+        if (filter !== '') {
+            var searchList = this.state.courses.filter(function (course) {
```
arkon: Couldn't you just do an `if` check within a `map` instead of filtering, then mapping?

david-yz-liu: @arkon Wouldn't that result in a list of the wrong length? Unless you mean to use an accumulator for the list.

arkon: @david-yz-liu Ah, you're right.

Alternatively, so you can avoid doing 2 loops, you could do something like:

```JSX
var searchList = [];

this.state.courses.forEach(function (course) {
    if (course.indexOf(filter) > -1) {
        searchList.push(<CourseEntry course={course} />);
    }
});
```

david-yz-liu: I'd rather stick with this approach for now. Later on I'll switch it to lodash, so we get the streaming behaviour and not "two loops".

public/js/grid/react_util.js
cchens: I don't know, to be honest. Blindly copied this code from the original Javascript file.

david-yz-liu: You should remove it. :)

public/js/grid/react_util.js
cchens: Never mind. In order to get the search icon to show up properly, you'd need either the form or another wrapper div (otherwise it would just be used as a background image for the input box)

david-yz-liu:
Thanks, @cchens!

PR #753, opened 2015-11-19
Economics

tamaralipowski:
Economics graph added - multiple "flowRoot" tags were found in the svg code - only first was deleted. Should all be deleted?


david-yz-liu:
Yes, all of them should be deleted.

david-yz-liu:
@tamaralipowski There is still one flowRoot remaining. I asked you to remove all of them. Please remove.

PR #754, opened 2015-11-22
Implemented web parsing for department names and codes.

alexbaluta:
Removed DepartmentInformation.hs and implemented web parsing in TimeTableParser.hs to populate database. Removed unnecessary set up steps in Database.hs.


Ian-Stewart-Binks:
Thanks Alex! I'll go through and add some questions/comments.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+import Database.Tables (Department(Department))
 
 parseTT :: IO ()
 parseTT = do
     rsp <- simpleHTTP (getRequest timetableUrl)
     body <- getResponseBody rsp
+
     let depts = getDeptList $ parseTags body
+    let deptCode = getDeptCode $ parseTags body
+    let deptName = getDeptName $ parseTags body
```
Ian-Stewart-Binks: You can remove this `let` and the one above it. See another place in the same PR where these are removed [here](https://github.com/Courseography/courseography/pull/754/files#diff-51b14c0659ad1c06571c3bf833a68171R67).

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
 
     runSqlite databasePath $
         mapM_ getDeptTimetable depts
+    runSqlite databasePath $ 
+        mapM_ insertDepartment $ init $ constructAssociationList deptName deptCode
```
Ian-Stewart-Binks: `runSqlite databasePath $ ` doesn't need to be declared twice when performing two distinct actions on the database. This snippet opens up a connection to the database, does some stuff, and then closes it. An easier way to think about it is that it just runs whatever you feed into it. So, I'm pretty sure you can just get rid of line 34. Take a look at some other code in Courseography that does the same thing.

I made this mistake at first, too. Inserting all of the courses into the database took a lot longer when we would perform one `runSqlite` call for each course. Now, it runs like lightning.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+-- | Takes a Char and a String and returns whether or not that Char is in the String.
+containsChar :: Char -> String -> Bool
+containsChar _ [] = False
+containsChar y (x:xs)
+    | y == x = True
+    | otherwise = containsChar y xs  
+
+-- | Takes a Char and a String and checks whether that char is in the String. If it
+-- is, it returns the original string, otherwise it returns an empty string.
+filterList :: Char -> String -> String
```
Ian-Stewart-Binks: This name doesn't really reveal the intention of the function. `filterList` could really mean many different things, and it doesn't have much to do with the doc above it.

Also, this encodes the list type into the function name, but the function doesn't operate on a list (well, not any list. A string is still a list, but this function wouldn't work on a list of Ints).

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+    | otherwise = containsChar y xs  
+
+-- | Takes a Char and a String and checks whether that char is in the String. If it
+-- is, it returns the original string, otherwise it returns an empty string.
+filterList :: Char -> String -> String 
+filterList _ [] = ""
+filterList y (x:xs) 
+    | y == x = (t:ts)
+    | otherwise = filterList y xs  
+    where (t:ts) = (x:xs)
```
Ian-Stewart-Binks: Why is `(t:ts)` defined as `(x:xs)`?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+containsChar y (x:xs)
+    | y == x = True
+    | otherwise = containsChar y xs  
+
+-- | Takes a Char and a String and checks whether that char is in the String. If it
+-- is, it returns the original string, otherwise it returns an empty string.
+filterList :: Char -> String -> String 
+filterList _ [] = ""
+filterList y (x:xs) 
+    | y == x = (t:ts)
```
Ian-Stewart-Binks: Would it make more sense to return `(x:xs)` here?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+bracketLeft :: Char
+bracketLeft = '['
+
+-- | Char right bracket to remove from the end of the last words in tagTexts.
+bracketRight :: Char
+bracketRight = ']'
+
+-- | List of unwanted words to remove when getting department codes.
+notDeptCodes :: [String]
+notDeptCodes = ["courses",
```
Ian-Stewart-Binks: Is it possible to improve this name?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+-- | Char left bracket used to call removeUnwantedChar. Removes all char prior to 
+-- and including itself from given String. 
+bracketLeft :: Char
+bracketLeft = '['
+
+-- | Char right bracket to remove from the end of the last words in tagTexts.
+bracketRight :: Char
+bracketRight = ']'
+
+-- | List of unwanted words to remove when getting department codes.
```
Ian-Stewart-Binks: Is it possible to make this a bit clearer, perhaps add in an example of when you would want to append a word to this list in the future?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+getDeptName :: [Tag String] -> [T.Text]
+getDeptName tags = 
+    let tagsText = map fromTagText $ filter isTagText tags
+        containsBrackLeft = filter (\x -> containsChar bracketLeft x == True) tagsText
+        convertToStr = map T.pack containsBrackLeft
+        deptNamesText = map head $ map (T.splitOn (T.pack bracketLeftString)) convertToStr
+    in deptNamesText
+
+-- | Takes a Char and a String and returns whether or not that Char is in the String.
+containsChar :: Char -> String -> Bool
```
Ian-Stewart-Binks: Okay. So, as a recursive function that asserts whether an item is an element of a list, this is a job well done. The name and docs are good too.

There's one problem, which is that we're redefining the wheel. Haskell provides functions that do this already. It's easier to use those, because many Haskell programmers (including future Courseography programmers) will know those functions too. Take a look [here](https://hackage.haskell.org/package/base-4.8.1.0/docs/Data-List.html). You can apply this to many other functions here. Try to see how much code you can simplify.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+removeUnwantedChar _ [] = ""
+removeUnwantedChar y (x:xs)
+    | y == x = xs
+    | otherwise = (x:xs)
+
+-- | Returns whether all the elements of the String are uppercase.
+isUpperString :: String -> Bool
+isUpperString [] = True
+isUpperString (x:xs)
+    | Data.Char.isUpper x == True && isUpperString xs == True = True
```
Ian-Stewart-Binks: Why do you give the full qualified name?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+    | otherwise = False
```
Ian-Stewart-Binks: There's a way to simplify this function greatly. Can you spot it?

hs/WebParsing/TimeTableParser.hs
alexbaluta: Thanks Ian, will do!

hs/WebParsing/TimeTableParser.hs
alexbaluta: Thanks again Ian, great to know!

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+-- | Takes a Char and a String and checks whether that char is in the String. If it
+-- is, it returns the original string, otherwise it returns an empty string.
+filterList :: Char -> String -> String 
+filterList _ [] = ""
+filterList y (x:xs) 
+    | y == x = (t:ts)
+    | otherwise = filterList y xs  
+    where (t:ts) = (x:xs)
+
+-- | Removes the chars before and including the Char input in the given String.
```
Ian-Stewart-Binks: I'm not sure what this function is intended to do. It looks as though it just removes the first character of the string if `x == y`, but I'm not getting that feeling from the doc/name.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+-- | Constructs an association list given a list of T.Text and a list of lists of T.Text.
+constructAssociationList :: [T.Text] -> [[T.Text]] -> [([T.Text], T.Text)]
+constructAssociationList _ [] = []
+constructAssociationList [] _ = []
+constructAssociationList (x:xs) (y:ys) = [(y, x)] ++ constructAssociationList xs ys
+
+-- | Inserts the Department association list in the database.
+insertDepartment :: MonadIO m => ([T.Text], T.Text) -> ReaderT SqlBackend m (Key Department)
+insertDepartment (code, name) = 
+    Database.Persist.Sqlite.insert (Department code name)
```
Ian-Stewart-Binks: Why qualified name?

hs/WebParsing/TimeTableParser.hs
alexbaluta: Thanks, I see how the names misleading. I'll change it to a more suitable one.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+                "courses]", 
+                "includes",
+                "and",
+                "-",
+                "[",
+                "]",
+                ""]
+
+-- | Parses the timetable main page, retrieves the tag text and gets the department
+-- codes in a list of lists of T.Text.
```
Ian-Stewart-Binks: Better to leave the function declaration to tell the reader what the function returns.

hs/WebParsing/TimeTableParser.hs
alexbaluta: I tried (x:xs) at first but the recursive call would only return the string characters from the point where the condition was met. For example if our Char was L and our string was APPLE, it would return LE. So setting (t:ts) equal to (x:xs) circumvented that problem.

hs/WebParsing/TimeTableParser.hs
alexbaluta: Sure thing, will do!

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+getDeptCode :: [Tag String] -> [[T.Text]]
+getDeptCode tags = 
+    let tagsText = map fromTagText $ filter isTagText tags
+        filterDeptInfo = map (filterList bracketLeft) tagsText
+        removedEmptyStr = filter (\x -> x /= "") filterDeptInfo
```
Ian-Stewart-Binks: You can section here.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+        removedChars = map (removeUnwantedChar bracketLeft) removedEmptyStr
+        listOfWords = map words removedChars
+        relevantWords = map (filter (`notElem` notDeptCodes)) listOfWords
+        deptCodesStr = map (filter (\x -> isUpperString x /= False)) relevantWords
```
Ian-Stewart-Binks: The function `getDeptCode` is a bit of a mouthful, and is taking me longer to understand. Are you able to improve the readability of this code? Comments and naming would be well appreciated.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+    let deptName = getDeptName $ parseTags body
 
     runSqlite databasePath $
         mapM_ getDeptTimetable depts
+    runSqlite databasePath $ 
+        mapM_ insertDepartment $ init $ constructAssociationList deptName deptCode
+
+-- | String left bracket to split tagText into two strings. One containing the 
+-- department name and the other the department codes. 
+bracketLeftString :: String
```
Ian-Stewart-Binks: I'm not certain that we need to bind "[" to an identifier. This seems a little redundant.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+    | otherwise = (x:xs)
+
+-- | Returns whether all the elements of the String are uppercase.
+isUpperString :: String -> Bool
+isUpperString [] = True
+isUpperString (x:xs)
+    | Data.Char.isUpper x == True && isUpperString xs == True = True
+    | otherwise = False
+
+-- | Constructs an association list given a list of T.Text and a list of lists of T.Text.
```
Ian-Stewart-Binks: Pardon my ignorance: What's an association?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+                "includes",
+                "and",
+                "-",
+                "[",
+                "]",
+                ""]
+
+-- | Parses the timetable main page, retrieves the tag text and gets the department
+-- codes in a list of lists of T.Text.
+getDeptCode :: [Tag String] -> [[T.Text]]
```
Ian-Stewart-Binks: Why a list of list of `T.Text`?

hs/WebParsing/TimeTableParser.hs
alexbaluta: Sure thing, Ill improve the documentation!

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,111 @@ import WebParsing.TimeConverter
...
+getDeptCode tags =
```
Ian-Stewart-Binks: I assume we're getting multiple codes, does it make more sense to pluralize this?

hs/WebParsing/TimeTableParser.hs
alexbaluta: Ill take a look at the built in functions again! Thanks for the heads up.

hs/WebParsing/TimeTableParser.hs
alexbaluta: My mistake, will fix. Thanks again!

Ian-Stewart-Binks:
@alexbaluta Nice work! I've added some comments, please address them and we can iterate from there.

hs/WebParsing/TimeTableParser.hs
alexbaluta: Ill reword the doc but for instance if we have the string "University" and the char 'v', the function will return "ersity". The way the Tag Texts were made it simple to break up the dept names from the dept codes with this function.

hs/WebParsing/TimeTableParser.hs
alexbaluta: The insert function is defined in Data.List as well as Database.Persist.Sqlite. Thanks again! I made this mistake earlier as you pointed out.

hs/WebParsing/TimeTableParser.hs
alexbaluta: Will make it easier to follow, thanks again!

hs/WebParsing/TimeTableParser.hs
alexbaluta: Makes sense, I'll remove that thanks!

hs/WebParsing/TimeTableParser.hs
alexbaluta: It's essentially like a map in Java or a dictionary in Python. It's a list of tuples and each tuple holds a key and a value.

hs/WebParsing/TimeTableParser.hs
alexbaluta: There are some special cases where some departments have multiple department codes and so we store them in the list of list.

hs/WebParsing/TimeTableParser.hs
alexbaluta: Ya I'll change that, thanks again!

alexbaluta:
Thanks for the comments Ian! They were very helpful and I'll make the necessary changes!

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: Ah, hmm. Is there any reason why we return a `Key Department` here?

Most of the other code doesn't utilize keys after inserting into the database. There are two functions that _Persist_ provides, [here](insertDepartment). One is `insert_` which discards the key. The other, `insert` returns the key.

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: I'm not certain that I understand how this function maps the input "University" to "ersity". Can you walk me through the code?

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: I still get _LE_ when I run this function on _APPLE_. Am I missing something?

You can use some of the Data.List functions to make this work, and to make this clearer.

alexbaluta: Oh sorry, my mistake you're right fixing that right now.

hs/WebParsing/TimeTableParser.hs
alexbaluta: insert_ is better suited in this case. Thanks again, I'm new to Haskell and the different libraries but your comments are really helpful!

hs/WebParsing/TimeTableParser.hs
alexbaluta: Ya no problem. The function looks at the first character in the string and let's assume it's not the char we're looking for. Then the recursive call is made on xs. Now the function goes through the xs similarly. Once it's found it returns the (x:xs) at that point. For instance, lets say we want to get P from APPLE. A is x and the rest is the xs. Then we call the function recursively on the string PPLE. Now x is P and the xs is PLE. So we return PPLE.

hs/WebParsing/TimeTableParser.hs
alexbaluta: Is this what you had in mind?
filter isUpper (x:xs) == (x:xs) = True
Thanks again for the help!

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: Hmm. Is there any way that we can better represent the data? Why do some departments have more than one code?

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: I'm new too ;) No problem!

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: Ah, but the code and your description of the code are different.

hs/WebParsing/TimeTableParser.hs
Ian-Stewart-Binks: Not quite. However, in your example, you'd be better off to use [all](https://www.haskell.org/hoogle/?hoogle=all). 

As a similar exercise:
In Python, what would be an easy way to clean up this function:
```python
def is_upper_string(x, xs):
    if is_upper_char(x) and is_upper_string(xs):
        return True
    else:
        return False
```

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+                "[",
+                "]",
+                ""]
+
+-- | Extracts a list of lists of department codes from the main website
+getDeptCodes :: [Tag String] -> [[T.Text]]
+getDeptCodes tags = 
+    let tagsText = map fromTagText $ filter isTagText tags
+        -- filterDeptInfo returns only the strings that contain dept codes
+        filterDeptInfo = map (charInString '[') tagsText
```
david-yz-liu: These two lines (and the helper function below) are too complex.

Using filter is good; make use of the `elem` function rather than defining `charInString`.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+        removedEmptyStr = filter (\x -> x /= "") filterDeptInfo
+        -- removedChars gets substring that contains dept codes
+        removedChars = map (removeCharsBefore '[') removedEmptyStr
```
david-yz-liu: Anything along the lines of "remove elements before..." can probably be done using `dropWhile`.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+        listOfWords = map words removedChars
+        relevantWords = map (filter (`notElem` wordsToRemove)) listOfWords
+        -- deptCodesStr removes words from wordsToRemove to isolate dept Codes
+        deptCodesStr = map (filter (\x -> all isUpper x == True)) relevantWords
```
david-yz-liu: Using `== True` is redundant. In fact, you can also eliminate the `x` and write `filter (all isUpper)` here.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+
+-- | Returns the chars after the first instance of the Char input from the given String.
+removeCharsBefore :: Char -> String -> String
+removeCharsBefore _ [] = ""
+removeCharsBefore y (x:xs)
+    | y == x = xs
+    | otherwise = (x:xs)
+
+-- | Constructs an association list given a list of T.Text and a list of lists of T.Text.
+constructAssociationList :: [T.Text] -> [[T.Text]] -> [([T.Text], T.Text)]
```
david-yz-liu: Use the `zipWith` function instead of defining this function.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
 import Control.Monad.IO.Class (liftIO, MonadIO)
 import Control.Monad.Trans.Reader (ReaderT)
 import Config (databasePath, timetableUrl)
+import Database.Tables (Department(Department))
```
david-yz-liu: Don't repeat the import here (see line 13). You should just use `Tables.Department` in your code.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+    | y == x = xs
+    | otherwise = (x:xs)
+
+-- | Constructs an association list given a list of T.Text and a list of lists of T.Text.
+constructAssociationList :: [T.Text] -> [[T.Text]] -> [([T.Text], T.Text)]
+constructAssociationList _ [] = []
+constructAssociationList [] _ = []
+constructAssociationList (x:xs) (y:ys) = [(y, x)] ++ constructAssociationList xs ys
+
+-- | Inserts the Department association list in the database.
```
david-yz-liu: Put this function right after `parseTT`.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+        -- deptCodesStr removes words from wordsToRemove to isolate dept Codes
+        deptCodesStr = map (filter (\x -> all isUpper x == True)) relevantWords
+        deptCodesText = map (map T.pack) deptCodesStr
+    in deptCodesText
+
+-- | Extracts a list of department names from the main website
+getDeptName :: [Tag String] -> [T.Text]
+getDeptName tags = 
+    let tagsText = map fromTagText $ filter isTagText tags
+        containsBrackLeft = filter (\x -> '[' `elem` x) tagsText
```
david-yz-liu: You can use `filter (elem '[')` here.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -17,15 +18,84 @@ import WebParsing.TimeConverter
...
+                "courses]", 
+                "includes",
+                "and",
+                "-",
+                "[",
+                "]",
+                ""]
+
+-- | Extracts a list of lists of department codes from the main website
+getDeptCodes :: [Tag String] -> [[T.Text]]
```
david-yz-liu: @alexbaluta A good way to simplify this code is to break up the parsing of Tags with the parsing of strings.

I want to see one function with the following type, which does the string parsing:

```haskell
parseLinkText :: String -> ([T.Text], -- the department codes
                            [T.Text] -- the department)
```

This has the advantage of being very easy to test, because you can just call it directly in the interpreter.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -22,10 +23,52 @@ parseTT :: IO ()
 parseTT = do
     rsp <- simpleHTTP (getRequest timetableUrl)
     body <- getResponseBody rsp
+
     let depts = getDeptList $ parseTags body
+        deptInfo = parseLinkText $ parseTags body
 
     runSqlite databasePath $
         mapM_ getDeptTimetable depts
+    runSqlite databasePath $
```
Ian-Stewart-Binks: Remove this.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -22,10 +23,52 @@ parseTT :: IO ()
...
 
     runSqlite databasePath $
         mapM_ getDeptTimetable depts
+    runSqlite databasePath $ 
+        mapM_ insertDepartment deptInfo
+
+-- | Inserts the Department association list in the database.
+insertDepartment :: MonadIO m => ([T.Text], T.Text) -> ReaderT SqlBackend m ()
+insertDepartment (code, name) = 
+    Database.Persist.Sqlite.insert_ (Tables.Department code name)
```
Ian-Stewart-Binks: Since we're doing `insert_`, is the qualified name still necessary?

hs/WebParsing/TimeTableParser.hs
```diff
@@ -22,10 +23,52 @@ parseTT :: IO ()
...
+                "-",
+                "[",
+                "]",
+                ""]
+
+-- | Extracts a list of tuples containing a list of dept codes and the respective dept name.
+parseLinkText :: [Tag String] -> [([T.Text],T.Text)]
+parseLinkText tags = 
+    let tagsText = map fromTagText $ filter isTagText tags
+        containsBrackLeft = filter (\x -> '[' `elem` x) tagsText
```
Ian-Stewart-Binks: You can section here.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -22,10 +23,52 @@ parseTT :: IO ()
...
+        convertToStr = map T.pack containsBrackLeft
+        -- deptNamesAndCodes splits lists into two strings; one containing
+        -- dept names and the other containing the unfiltered dept codes
+        deptNamesAndCodes = map (T.splitOn (T.pack "[")) convertToStr
```
Ian-Stewart-Binks: It might be more readable if the code separates defining intermediate functions from defining intermediate steps. For instance, `convertToStr` doesn't actually do anything when its declared, so it might be better to put that into a `where` clause. This way, I can understand what lines manipulate data and what lines declare functions. Just a thought.

hs/WebParsing/TimeTableParser.hs
```diff
@@ -22,10 +23,52 @@ parseTT :: IO ()
...
+        -- listOfWords takes the unfiltered string containing dept codes and 
+        -- makes a list of the words in that string
+        deptCodesUnfiltered = map T.unpack $ map last deptNamesAndCodes 
+        listOfWords = map words deptCodesUnfiltered
+        relevantWords = map (filter (`notElem` wordsToRemove)) listOfWords
+        -- deptCodesStr filters the dept codes
+        deptCodesStr = map (filter (all isUpper)) relevantWords
+        deptCodesText = map (map T.pack) deptCodesStr
+        -- deptAssocList zips the lists of dept codes and their dept name into a tuple 
+        deptAssocList = init $ zipWith ( , ) deptCodesText deptNamesText
```
Ian-Stewart-Binks: Use `zip`.

hs/WebParsing/TimeTableParser.hs
alexbaluta: I tried but it wont compile.

Ian-Stewart-Binks: What's the error message? You likely need to just `do` it.

alexbaluta: Oh right, sorry my oversight.

Ian-Stewart-Binks: No problem :-)

PR #755, opened 2015-11-24
Specialist Component Final Bug Fixes/Additions

christinem:
This PR adds some final fixes and additions to the Specialist component, such as having a separate component for each Course Year category, disabling appropriate textboxes, and some bug fixes. I also fixed the bug where MAT137 and MAT157 course codes wouldn't render. The only thing remaining to be done is Credit Count.

This PR is built alongside #747 (since it was created when I first pushed #747, and before I started making changes to it, so this minimizes conflicts), so that one should be merged before looking at this one.

### DIFF
--- a/public/js/post/post.js
+++ b/public/js/post/post.js
@@ -92,7 +92,8 @@ var CourseCategory = React.createClass({
         return {
             completed: false,
             completedTextBoxes: 0,
-            infoOpened: false
+            infoOpened: false,
+            textboxValues: this.createInitialValueArray()
         }
     },
 
@@ -109,20 +110,30 @@ var CourseCategory = React.createClass({
         this.setState({completed: this.state.completedTextBoxes === this.props.textBoxNumber});
     },
 
-    handleKeyDown: function(e) {
-        if (e.keyCode === 13) {
-            if (this.state.completedTextBoxes <= this.props.textBoxNumber + 1) {
-                if (e.target.defaultValue === '' && e.target.value !== '') {
-                    this.setState({completedTextBoxes: this.state.completedTextBoxes + 1},
-                        this.checkIfCompleted);
-                } else if (e.target.defaultValue !== '' && e.target.value === '') { 
-                    this.setState({completedTextBoxes: this.state.completedTextBoxes - 1},
-                        this.checkIfCompleted);
-                }
-            }
+    createInitialValueArray: function() {
+        var array = Array(this.props.textBoxNumber).join(".").split(".");
+        for (i = 0; i < this.props.courses.length; i++) {
+            array[i] = this.props.courses[i];
+        }
+        return array;
+    },
 
-            e.target.defaultValue = e.target.value;
-        }  
+    handleOnChange: function(e) {
+        var newValues = this.state.textboxValues;
+        newValues[e.target.id] = e.target.value.substring(0, 6);
+        this.setState({textboxValues: newValues});
+        this.setState({completedTextBoxes: this.countCompletedTextBoxes()}, this.checkIfCompleted);
+    },
+
+    countCompletedTextBoxes: function() {
+        var count = 0;
+        for (i = 0; i < this.state.textboxValues.length; i++) {
+            if (this.state.textboxValues[i].length === 6) {
+                count += 1;
+            }
+        }
+
+        return count;
     },
 
     render: function() {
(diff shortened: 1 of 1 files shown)


public/js/post/post.js
```diff
@@ -109,11 +141,15 @@ var MultipleCourseCode = React.createClass({
...
             <div id={this.props.courseID} className={classes}>
-                <p className="code" onClick={this.toggleFullInfo}> {this.props.data.categoryName} </p>
+                <p className="code" onClick={this.toggleFullInfo}> {this.props.categoryName} </p>
                 <div id = {'spec' + this.props.courseID.substring(5, this.props.courseID.length)} className={infoClasses}>
                     <p className="full_name"> 
-                        {Array.apply(0, Array(this.props.data.textBoxNumber)).map(function (x, i) {
-                            return <input type='text' defaultValue={me.props.data.courses[i]} onKeyDown={me.handleKeyDown} />;
+                        {Array.apply(0, Array(this.props.textBoxNumber)).map(function (x, i) {
+                            if (me.props.textboxesDisabled) {
+                                return <input type='text' defaultValue={me.props.courses[i]} onKeyDown={me.handleKeyDown} disabled />;
```
arkon: Could just do `disabled={me.props.textboxesDisabled}` instead of the `if`/`else`.

public/js/post/post.js
```diff
@@ -122,6 +158,22 @@ var MultipleCourseCode = React.createClass({
     }
 })
 
+var CourseYear = React.createClass({
+    getInitialState: function() {
```
arkon: Is this necessary if there's no state?

public/js/post/post.js
```diff
@@ -29,8 +30,39 @@ var CourseCode = React.createClass({
         }
     },
 
+    getIdName: function() {
+        var idName = this.props.courseIDs[0].substring(0, 3);
+        
+        this.props.courseIDs.map(function (course) {
```
david-yz-liu: use `forEach`, not `map`

public/js/post/post.js
```diff
@@ -207,30 +259,21 @@ var SpecialistPost = React.createClass({
...
-                    categoryName: 'Any 300+ level CSC course, BCB410H, BCB420H, BCB430Y, ECE385H, ECE489H (1.5 FCEs)'}} />
-                <MultipleCourseCode courseID="spec_extra" data={{textBoxNumber: 4, courses: courseCategoryArrays[2],
-                    categoryName: 'Any of the following: 300+ level CSC course; MAT: 235/237/257, any 300+ \
-                                     except for 329, 390, & 391; STA: 248, 261, any 300+; ECE: 385H/489H; \
-                                     BCB: 410H/420H/430Y (2.0 FCEs)'}} />
-                <MultipleCourseCode courseID="spec_inq" data={{textBoxNumber: 1, courses: courseCategoryArrays[3],
-                    categoryName: 'Any from this list: CSC301H, CSC318H, CSC404H, CSC411H, CSC418H, CSC420H, \
-                    CSC428H, CSC454H, CSC485H, CSC490H, CSC491H, CSC494H, or PEY (0.5 FCEs) \
-                    ** Note: Type "PEY" for Check my POSt to recognize it **'}} />
+                <CourseYear yearName='First Year' courses={firstYearCourses} />
```
david-yz-liu: Rename this component to "CourseCategory".

public/js/post/post.js
christinem: I looked into this beforehand. The HTML5 spec says that the disabled attribute should either be in the element, or not be in the element (https://github.com/facebook/react/issues/961). But what you suggest does work. Should I just go with that?

arkon: You're right -- the HTML5 spec does say that. It's important to remember that JSX has some differences from HTML's syntax though, and this is actually one of those things (as mentioned in that issue). It'd certainly be cleaner to go with JSX's behaviour.

christinem: Cool! I was only skeptical since the link I quoted above actually is a React Issue where they discuss making people use the bare disable attributes instead of the "disabled=true, disabled=false" functionality. I do agree using it for now is cleaner.

public/js/post/post.js
christinem: Ah, I was supposed to this remove this. Thanks!

david-yz-liu:
@christinem I need you to get rid of the "Course Year" component in favour of "CourseCategory."

christinem:
@david-yz-liu ah my mistake, I thought you meant to rename MultipleCourseCode to CourseCategory. Or did you mean get rid of Course Year completely?

david-yz-liu:
Comments generally refer to particular lines. In this case, the "replace" comment is attached to a "CourseYear" line.

PR #757, opened 2015-12-01
Portuguese graph added

tamaralipowski:
Portuguese graph added


david-yz-liu:
@tamaralipowski The Portuguese graph has two regions, the bottom left and bottom right, which are rectangles instead of paths. This is not acceptable; please delete them and replace them.

PR #758, opened 2015-12-01
Spanish Graph Added

tamaralipowski:


Ian-Stewart-Binks:
I think that we should represent the MAT courses consistently in all graphs. Are we going to split the MAT node into four separate nodes in the CSC graph?

david-yz-liu:
@Ian-Stewart-Binks I'll think about it.

@tamaralipowski I need you to follow the naming conventions in the original graphs. Please rename all the files as soon as possible.

david-yz-liu:
@tamaralipowski The spanish graph also has several regions which are rectangles instead of paths. We have already discussed this; please redo them.

The middle blue region has some weird coordinates. Please remake that region as well.

PR #759, opened 2015-12-03
React graph mouse events added.

ryanfan:
Note : Hybrids are currently understood as `inactive` because they were hardcoded in the old code, and I need to think of a way to add them as children of it's actual node.

### DIFF
--- a/public/js/graph/react_node.js.jsx
+++ b/public/js/graph/react_node.js.jsx
@@ -104,6 +104,12 @@ var ReactSVG = React.createClass({
         var currentNode = this.refs['nodes'].refs[courseID];
         
         currentNode.focusPrereqs(this);
+        
+        //Old hover modal code
+        if ($(".modal").length === 0 && activeFocus === '') {
+            removeToolTips();
+            displayTooltip(courseID);
+        }
 
     },
     
@@ -113,6 +119,15 @@ var ReactSVG = React.createClass({
         var currentNode = this.refs['nodes'].refs[courseID];
             
         currentNode.unfocusPrereqs(this);
+        
+        //Old hover modal code
+        if ($(".modal").length === 0) {
+            var timeout = setTimeout(function () {
+                $('.tooltip-group').hide('slow', function () { $(this).remove();});
+            }, 100);
+
+            timeouts.push(timeout);
+        }
     },
     
     render: function() {
@@ -120,6 +135,7 @@ var ReactSVG = React.createClass({
         var svgAttrs = {'width': this.props.width, 'height': this.props.height};
         var markerAttrs = {'id': 'arrow'};
         var polylineAttrs = {'points': '0,1 10,5 0,9', 'fill': 'black'};
+        var svg = this;
         return (
             <svg {... svgAttrs} ref='svg'>
                 <defs>
@@ -128,7 +144,7 @@ var ReactSVG = React.createClass({
                     </marker>
                 </defs>
                 <ReactRegions/>
-                <ReactNodes ref='nodes' onClick={this.nodeClick} onMouseEnter={this.nodeMouseEnter} onMouseLeave={this.nodeMouseLeave}/>
+                <ReactNodes ref='nodes' onClick={this.nodeClick} onMouseEnter={this.nodeMouseEnter} onMouseLeave={this.nodeMouseLeave} svg={svg}/>
                 <ReactBools ref='bools'/>
                 <ReactEdges ref='edges'/>
                 <ReactRegionLabels/>
@@ -214,6 +230,7 @@ var ReactNodes = React.createClass({
     },
     
     render: function() {
+        var svg = this.props.svg;
         return (
             <g id='nodes' stroke='black'>
                 {this.state.nodesList.map(function(entry, value) {
@@ -244,7 +261,9 @@ var ReactNodes = React.createClass({
                             childs={childs}
                             inEdges={inEdges}
                             outEdges={outEdges}
-                            {... this.props}/>
+                            {... this.props}
+                            svg={svg}
+                            hybrid={false}/>
                 }, this)}
     
                 {this.state.hybridsList.map(function(entry, value) {
@@ -274,7 +293,9 @@ var ReactNodes = React.createClass({
                             parents={parents}
                             childs={childs}
                             inEdges={inEdges}
-                            outEdges={outEdges}/>
+                            outEdges={outEdges}
+                            svg={svg}
+                            hybrid={true}/>
                 }, this)}
             </g>
         );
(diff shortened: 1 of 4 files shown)


public/js/graph/react_node.js.jsx
```diff
@@ -204,59 +306,168 @@ var ReactNode = React.createClass({
...
+        }
+        return sat;
     },
     
-    handleMouseLeave: function(event) {
-        //code here
+    updateNode: function(svg, newState) {
+        var newNewState =''
+        if (this.arePrereqsSatisfied(svg)) {
+            if (this.isSelected() || this.props.hybrid) {
```
david-yz-liu: @ryanfan I thought the point was to use `newState` here instead of `this.isSelected()`?

ryanfan: @david-yz-liu I cleaned this up in my newest commit, `newState` won't be needed anymore.  I'm using `bind` in my callback function and it know the new updated state to work with.

public/js/graph/react_node.js.jsx
```diff
@@ -187,14 +226,71 @@ var ReactNodes = React.createClass({
...
                 {this.state.nodesList.map(function(entry, value) {
-                    return <ReactNode attributes={entry['attributes']} children={entry['children']} className='node' key={entry['id']} styles={entry['style']} hybrid={false}/>
-                })}
+                    var parents = [];
+                    var childs = [];
+                    var outEdges = [];
+                    var inEdges = [];
+
+                    $('.path').map(function(key, element) {
+                        if (entry['id'] == element.getAttribute('data-target-node')){
```
david-yz-liu: Use `===`, and put a space before the `{` (everywhere)

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
-                children.push(element.getAttribute('data-target-node'));
-                outEdges.push(element.id);
-            }
-        });
-        
-        if (parents.length == 0){
-            status = 'takeable';
-        }
-        
+        var type = 'AND';
```
david-yz-liu: Make `type` a prop, and give it a default value of `'AND'`

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
+            if (svg.refs['nodes'].refs[element]){
+                return svg.refs['nodes'].refs[element].isSelected();
+            }
+            if (svg.refs['bools'].refs[element]){
+                return svg.refs['bools'].refs[element].isSelected();
+            }
+        }
         
+        if (this.state.logicalType === 'AND') {
+            sat = this.props.parents.every(isAllTrue);
```
david-yz-liu: Don't use a temporary variable; just `return` directly here.

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
+        return sat;
     },
     
-    handleMouseLeave: function(event) {
-        //code here
+    updateNode: function(svg) {
+        var newState =''
+        if (this.arePrereqsSatisfied(svg)) {
+            if (this.isSelected() || this.props.hybrid) {
+                newState='active';
```
david-yz-liu: Missing some spaces around `=` (check everywhere).

david-yz-liu: Also, if you don't think you need the `newState` any more, you can remove it entirely.

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
+        
+        //setCookie(this.id, this.status);
+
+        // Always update children of hybrids
+        if (this.state.hybrid) {
+            $.each(this.props.childs, function(i, node) {
+                var currentNode = svg.refs['nodes'].refs[node] ? svg.refs['nodes'].refs[node] :
+                                                                 svg.refs['bools'].refs[node];
+                currentNode.setState({status: currentNode.state.status},
+                                     function(){this.updateNode(svg)}.bind(currentNode));
```
david-yz-liu: Put spaces around the `()` and `{` and `}`.

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
+        $.each(this.props.childs, function (i, node) {
+            var currentNode = svg.refs['nodes'].refs[node] ? svg.refs['nodes'].refs[node] :
+                                                             svg.refs['bools'].refs[node];
+            currentNode.setState({status: currentNode.state.status},
+                                 function(){this.updateNode(svg)}.bind(currentNode));
+        });
+        
+        $.each(this.props.outEdges, function (i, edge) {
+            var currentEdge = svg.refs['edges'].refs[edge];
+            currentEdge.setState({status: currentEdge.state.status},
```
david-yz-liu: This line just screams like something is wrong, because it seems to be updating a state with its current value.

public/js/graph/react_node.js.jsx
```diff
@@ -203,60 +299,165 @@ var ReactNodes = React.createClass({
...
+                var currentNode = svg.refs['nodes'].refs[node] ? svg.refs['nodes'].refs[node] :
+                                                                 svg.refs['bools'].refs[node];
+                currentNode.focusPrereqs(svg);
+            });
+        }
+    },
+    
+    unfocusPrereqs: function(svg){
+        if (!this.isSelected()) {
+            /*if (activeFocus === '' ||
```
david-yz-liu: I don't really know what this code is for, but if you think you can remove it, please do so.

public/js/graph/react_node.js.jsx
```diff
@@ -292,12 +508,128 @@ var ReactBools = React.createClass({
 });
 
 var ReactBool = React.createClass({
+    getInitialState: function() {
+        var id = this.props.attributes['id'];
+        var type = this.props.children[1].innerHTML.toUpperCase();
```
david-yz-liu: Make the `type` a prop

public/js/graph/react_node.js.jsx
```diff
@@ -174,93 +212,223 @@ var ReactRegion = React.createClass({
...
             hybridsList: []
         };
     },
     
-    componentDidMount: function() {
-        this.setState({nodesList:getNodes('.node')});
-        this.setState({hybridsList:getNodes('.hybrid')});  
+    componentDidMount: function () {
+        this.setState({nodesList: getNodes('.node')});
+        this.setState({hybridsList: getNodes('.hybrid')});
```
arkon: You should just set both at the same time.

public/js/graph/react_node.js.jsx
```diff
@@ -174,93 +212,223 @@ var ReactRegion = React.createClass({
...
+        var svg = this.props.svg;
         return (
             <g id='nodes' stroke='black'>
-                {this.state.nodesList.map(function(entry, value) {
-                    return <ReactNode attributes={entry['attributes']} children={entry['children']} className='node' key={entry['id']} styles={entry['style']} hybrid={false}/>
-                })}
-                {this.state.hybridsList.map(function(entry, value) {
-                    return <ReactNode attributes={entry['attributes']} children={entry['children']} className='hybrid' key={entry['id']} styles={entry['style']} hybrid={true}/>
-                })}
+                {this.state.nodesList.map(function (entry, value) {
```
arkon: This and the block below seem pretty similar, perhaps create a helper function to reduce redundancy?

public/js/graph/react_node.js.jsx
```diff
@@ -174,93 +212,223 @@ var ReactRegion = React.createClass({
...
-            inEdges: inEdges,
-            logicalType: type,
-            updated: false,
-            hybrid: this.props.hybrid,
-            status: status
-        };
+
+        // Always update children of hybrids
+        if (this.props.hybrid) {
+            $.each(this.props.childs, function (i, node) {
```
arkon: Looks like you could just use a regular non-jQuery `forEach` loop here (and below)?

public/js/graph/react_node.js.jsx
```diff
@@ -174,93 +212,223 @@ var ReactRegion = React.createClass({
...
+            $.each(this.props.outEdges, function (i, edge) {
+                var currentEdge = svg.refs['edges'].refs[edge];
+                currentEdge.forceUpdate(function () { currentEdge.updateEdge(svg) });
+            });
+        }
     },
     
-    handleClick: function(event) {
-        //code here
+    turn: function (svg) {
```
arkon: The name `turn` gives me the impression that this function rotates a node or something.

public/js/graph/react_node.js.jsx
```diff
@@ -174,93 +212,223 @@ var ReactRegion = React.createClass({
...
+        if (this.isSelected()) {  
+            if (this.props.parents.length === 0) {
+                this.setState({status: 'takeable'}, function () { this.updateNode(svg) });
```
arkon: Could just use a ternary operator instead of if/else.

david-yz-liu:
@arkon Thanks for your comments. I have already given @ryanfan enough to work on.

david-yz-liu:
@ryanfan Just tried it. Seems like no nodes have any parents. Also doesn't seem like edges have any "attributes" in the props.

PR #760, opened 2015-12-06
Economics Graphs - Parsing Fixed

tamaralipowski:
All parsing problems fixed - problem with text only showing up black also fixed.


david-yz-liu:
Looks great, @tamaralipowski!

PR #762, opened 2015-12-06
Portuguese Graphs - Parsing Fixed

tamaralipowski:
All parsing errors fixed.


david-yz-liu:
I think the region labels are missing on this one - can you please check, @tamaralipowski?

PR #765, opened 2015-12-09
Slavic Dept Graph - Parsing Problems

tamaralipowski:
The  coordinates of the coloured regions always mess up when parsing.
Not sure how to fix this.


david-yz-liu:
@tamaralipowski So unfortunately I couldn't figure out *why* it was happening, but I did manage to fix the coordinate problems.

1. I first loaded the graph in my browser, and then used the developer tools to identify the weird regions, and find their coordinates.
2. Opened the SVG file in a text editor, did a search for those coordinates (to find the right region), and then deleted the whole path element.
3. Redrew it in Inkscape.

The process worked for me on my first try, though maybe I was lucky. I used the "Create a sequence of paraxial lines" tool, if that helps.

PR #766, opened 2015-12-10
stack: introduce stack configuration and reorganize directory structure

david-yz-liu:
This pull request enables courseography to be built using [stack](http://docs.haskellstack.org/en/stable/README.html).

New installation instructions:

1. Install stack (see above link).
2. ~~Download [ghc 7.10.3](https://www.haskell.org/ghc/download_ghc_7_10_3). This seems to be required on Windows, but everyone should do it.~~ 
    
    Run `stack setup 7.10.3`.
3. Navigate to the Courseography directory, *top-level*.
4. **Update Config.hs: `cp hs/DevelopmentConfig.hs hs/Config.hs`.** Probably also need to do a `mkdir db`.
5. Run `stack build`. It will take some time to build everything.
6. Run the usual commands in a slightly different way, all from the top-level (**not** hs/):

```
> stack exec courseography css
> stack exec courseography graphs
> stack exec courseography database
> stack exec courseography
```

You should now be able to go to `localhost:8000` and view the graph.


david-yz-liu:
@Ian-Stewart-Binks @christinem @alexbaluta I would appreciate if the above instructions were tested on all three of Windows, Mac, and Linux. Not urgent, but hopefully some time in December!.

Ian-Stewart-Binks:
Tried updating to ghc 7.10.3 a bit ago, was taking far too long so I pushed it to another time.

Why does this pull in Haxml?

christinem:
Mac users on El Capitan definitely need 7.10.3 in order for stack to work, because of the new System Integrity Protection mode (according to the documentation), so it's good for everyone to update to it.

When running the `stack build` command, I'm getting this error:
```
Warning: /Users/christinemurad/Documents/courseography/stack.yaml: Unrecognized field in ProjectAndConfigMonoid: extra-package-dbs
Could not resolve directory /Users/christinemurad/Documents/courseography/dependencies\HaXml-1.25.3
```

The directory definitely exists.

christinem:
Ah I figured out the problem - Windows and Mac/Linux use different brackets. I had to go to stack.yaml and change this line:

` - dependencies/HaXml-1.25.3`

to this

`- dependencies\HaXml-1.25.3`

Ian-Stewart-Binks:
> Why does this pull in Haxml?

https://github.com/david-yz-liu/courseography/blob/stack/stack.yaml#L9

christinem:
So, after the install, turns out I can't run the stack commands from top-level. Since the file it tries to access is `../public/res/graphs/csc2015.svg`, where `..` would take me out of the courseography folder all-together, so for me it tries to access `Documents/public/res/graphs/csc2015.svg`. Tried to run the command from `hs` folder, and it started working, but ran into this:

```
Migrating: CREATE TABLE "department"("id" INTEGER PRIMARY KEY,"code" VARCHAR NOT NULL,"name" VARCHAR NOT NULL)
courseography: SQLite3 returned ErrorReadOnly while attempting to perform step.
```
Not sure which file or line it's failing on. I also get this warning for every command still:
```
Warning: /Users/christinemurad/Documents/courseography/stack.yaml: Unrecognized field in ProjectAndConfigMonoid: extra-package-dbs
```

I'll explore a little more.

Ian-Stewart-Binks:
[Never crossed my mind as to how weird it is that Microsoft still uses the escape char.](https://blogs.msdn.microsoft.com/larryosterman/2005/06/24/why-is-the-dos-path-character/)

christinem:
Update: Fixed the error with a `sudo` before every command. Still has to be run from within some folder _second_ from top-level (such as `hs`), not from `courseography`. And I still get that warning when running the commands. Graphs and database parsed, and was able to get it running, but there's a bunch of 404s:

screen shot 2015-12-17 at 4 47 50 am

david-yz-liu:
@Ian-Stewart-Binks HaXml didn't compile on either of my Windows machines - had a dumb error having to do with line endings, or perhaps the C preprocessor for Haskell. I changed line 79 of the cabal file to make it work.

@christinem Sorry, I forgot to mention that since `DevelopmentConfig.hs` has been updated, you should copy that over to `Config.hs` first before starting. I'll update that!

christinem:
@david-yz-liu ahhhh I see, thanks - there's only one problem, is that the database file URL in `DevelopmentConfig.hs` is relative to `hs`, not `courseography` - so the graph files are searched relative to `courseography` (as desired), but the database file is searched relative to `hs`. I had to change this line:

```
-- | The path to the database file, relative to @hs/@.
databasePath :: Text
databasePath = "db/database2015.sqlite3"
```

to this:

```
-- | The path to the database file, relative to @hs/@.
databasePath :: Text
databasePath = "hs/db/database2015.sqlite3"
```

Also, had to rename `Database` folder to `db` folder (since there's no `db` folder in the repository). Then I could finally run the commands from `courseography` top-level. And after that courseography worked properly. Still get that weird warning though.

christinem:
_Comment Edit: Missed the mkdir db part in the updated original post when writing my last comment_

Thanks @david-yz-liu, that solved my problems! It does work properly now with the updated instruction, and I was able to get it to run.

david-yz-liu:
Okay, thanks @christinem!

PR #767, opened 2015-12-15
French Graphs Added

tamaralipowski:
Coloured regions removed due to extensive parsing problems. Parsing
problems with highlighted nodes fixed.


david-yz-liu:
Thanks, @tamaralipowski!

PR #770, opened 2016-01-06
Credit Count fixes and File Seperation

christinem:
This PR seperates course components and and post components into their own files which follow the new naming convention, and adds some credit count fixes for typed courses. I also made Inquiry Category its own component as was discussed before, to avoid the mutual exclusion issue with categories.


public/js/post/post_components.js.jsx
```diff
@@ -0,0 +1,132 @@
...
+                }
+            }
+        });
+
+        return courseArrays;
+    },
+
+    getInquiryCourse: function () {
+        var inquiryCourses = this.state.activeCourses.filter(this.isInquiryCourse);
+        return inquiryCourses == [] ? '' : inquiryCourses[0];
```
david-yz-liu: The check shouldn't work. Use `inquiryCourses.length === 0`.

david-yz-liu:
@christinem Putting React JSX components in different files without a module loading system creates a race condition because of the jsx compilation, and so isn't good, even if it seems to work. I will fix this, but just letting you know.

david-yz-liu:
@christinem Okay I've pushed various changes to the branch. Please review the last few commits, and then work on replacing the major and minor POSts.

christinem:
@david-yz-liu ah, thanks for letting me know, I didn't realize that. And thanks! Will review and get started on the other POSts.

PR #771, opened 2016-01-12
East Asian Graph - Parsing Problems

tamaralipowski:
The parser does not pick up the nodes or paths and all or-nodes are
filled in black.
I have: deleted flow-roots, deleted invisible rectangles,  ensured and-
& or- nodes are identical to those previously used, converted all text,
nodes, and paths, to their corresponding types.


david-yz-liu:
Thanks, @tamaralipowski. I'll look at this today.

david-yz-liu:
@tamaralipowski I just ran the parser and it worked beautifully on this graph! So your setup might have general parsing problems. Let's meet offline to see if we can resolve the problem.

PR #779, opened 2016-01-23
reset-graph: clear graphs tables before running 'graphs' command

david-yz-liu:
Now when you run `courseography graphs` the graph tables will reset, so there won't be any duplicates.


Ian-Stewart-Binks:
Looks good to me!

david-yz-liu:
Thanks, @Ian-Stewart-Binks!

PR #780, opened 2016-01-28
React code for Minor POSt

christinem:
This PR adds the code for the Minor POSt in React. MultipleCourseCode components are now created dynamically since the Minor POSt has less "categories" than the Major or Specialist, so this keeps us still being able to use a generic Post component that others can extend from.


public/js/post/course_components.js
```diff
@@ -178,7 +178,7 @@ export var MultipleCourseCode = React.createClass({
                 <div id = {'spec' + this.props.courseID.substring(5, this.props.courseID.length)} className='more-info'>
                     <p className="full_name"> 
                         {Array.apply(0, Array(this.props.textBoxNumber)).map(function (x, i) {
-                            if (me.isValidExtraCourse(me.state.textboxValues[i])) {
+                            if (me.isValidExtraCourse(me.state.textboxValues[i].substring(0, 6))) {
```
Ian-Stewart-Binks: `.substring(0, 6)` is fairly ambiguous to new eyes. Perhaps defining a function could make this easier to understand? I noticed this in a recent pass over some of the current code in Courseography.

public/js/post/post_components.js
```diff
@@ -139,9 +147,11 @@ var SpecialistPost = React.createClass({
                   secondYearCourses={secondYearCourses} 
                   laterYearCourses={laterYearCourses} 
                   textBoxNumbers={[3, 3, 4]} 
+                  disabledTextboxes={[true, true, false]}
```
david-yz-liu: @christinem I don't like having two parallel lists that must be the same length. Probably pass in a nested list instead.

PR #784, opened 2016-02-03
Graph of History and Philosophy of Science

tamaralipowski:
Saved as plain svg format. No parsing errors.


david-yz-liu:
Thanks, @tamaralipowski!

PR #785, opened 2016-02-03
Geography Graph - Parsing Error

tamaralipowski:
Graph does not parse.


david-yz-liu:
Hi @tamaralipowski, this is a transform problem. Please open the svg file in Inkscape and try the following:

1. Edit -> Select All in All Layers.
2. While everything is selected, go to Object -> Transform
3. In the Transform Panel, make sure "Relative move" is unchecked. Click Apply.
4. Save the file, then try to reparse.

Let me know how it goes!

PR #788, opened 2016-02-06
Json request: source and targets added to JSON response

ryanfan:


david-yz-liu:
@ryanfan it would be great to remove the corresponding code from Builder.hs (or Generator.hs, can't remember which).

PR #789, opened 2016-02-11
added SVG dimensions to database so that generated graphs can be sized to fit container

bell-kelly:


david-yz-liu:
@bell-kelly good job - remember to create a new branch next time, rather than working directly from your master branch.

PR #794, opened 2016-02-18
Aboriginal Graphs - Parsing Error

tamaralipowski:
"no parse" message given.
I have checked for transform errors.


david-yz-liu:
@tamaralipowski - I've updated the parser so it no longer gives an error on this graph.

Note that you'll need to do a pull from master to get all the changes. I'm not sure you've been pulling regularly, so it might require some effort.

PR #795, opened 2016-02-26
Loading icon

R0ss:
Loading icon server responses (Large full res, and small 256x256 pixel sizes)


hs/Response/Draw.hs
```diff
@@ -47,3 +46,12 @@ modePanel = H.div ! A.id "side-panel-wrap" $ do
...
     H.div ! A.id "change-mode" ! A.class_ "mode" $ "SELECT/MOVE (m)"
     H.div ! A.id "erase-mode" ! A.class_ "mode" $ "ERASE (e)"
+    H.input ! A.id "area-of-study"
+            ! A.class_ "course-code"
+            ! A.name "course-code"
+            ! A.placeholder "Enter area of study."
+            ! A.autocomplete "off"
+            ! A.type_ "text"
+            ! A.size "30"
+    H.div ! A.id "submit-gid" ! A.class_ "button" $ "Submit"
```
Ian-Stewart-Binks: Are the `$` necessary here?

Ian-Stewart-Binks: I recall there being an issue with removing the `$` if attributes were specified. I forget why though.

hs/Server.hs
```diff
@@ -12,7 +12,9 @@ import Control.Monad (msum)
 import Control.Monad.IO.Class (liftIO)
 import Happstack.Server hiding (host)
 import Response
+
 import Database.CourseQueries (retrieveCourse, allCourses, queryGraphs, courseInfo, deptList, getGraphJSON)
+
```
Ian-Stewart-Binks: Inconsistent blank lines.

public/js/draw/setup.js
```diff
@@ -86,6 +86,20 @@ $('#finish-region').click(function () {
     finishRegion();
     });
 
+$('#submit-gid').click(function() {
+       $.ajax({
+            url: 'get-json-data',
+            data: {gid : $('#area-of-study').val()},
+            dataType: 'json', 
+            success: function(data){
```
Ian-Stewart-Binks: Space.

hs/Response/Draw.hs
R0ss: Not sure, the code was like this prior to my changes.

Should I remove them?

stack.yaml
```diff
@@ -1,7 +1,8 @@
 # For more information, see: https://github.com/commercialhaskell/stack/blob/release/doc/yaml_configuration.md
 
 # Specifies the GHC version and set of packages available (e.g., lts-3.5, nightly-2015-09-21, ghc-7.10.2)
-resolver: lts-3.17
+resolver: lts-5.1
```
david-yz-liu: Don't include this change in the pull request.

hs/Response.hs
```diff
@@ -12,6 +12,8 @@ module Response
         module Response.Grid,
         -- | @/image@
         module Response.Image,
+        -- | @/timesearch@
```
david-yz-liu: The comment should specify the route used. This isn't `timesearch`.

hs/Response/Draw.hs
david-yz-liu: I think you need to leave them... @Ian-Stewart-Binks or @R0ss, either of you could actually try to change it yourself.

R0ss: Based on the documentation: https://hackage.haskell.org/package/blaze-html-0.1/docs/Text-Blaze.html#2


From HTML 5 onwards, the user is able to specify custom data attributes.

An example:

 <p data-foo="bar">Hello.</p>
We support this in BlazeHtml using this funcion. The above fragment could be described using BlazeHtml with:

 p ! dataAttribute "foo" "bar" $ "Hello."

Ian-Stewart-Binks: Thanks @R0ss! This happens because `$` forces the two `!` operators to evaluate before it evaluates, as priority of `!` >  priority of `$`, correct? Otherwise, we likely are trying to compute `A.class_ "button" "Submit"`?

R0ss: Yes. The `$` is the only thing that distinguishes inner-text from multi-argument attributes.

hs/Css/Loading.hs
```diff
@@ -0,0 +1,45 @@
...
+{- loadingStyles
+ - Generates all CSS for the draw page. -}
+loadingStyles :: Css
+loadingStyles = do
+    spinnerCSS
+    smallCompassPosCSS
+    largeCompassPosCSS
+    spinCSS
+
+{- Properties of rotation animation. -}
```
Ian-Stewart-Binks: This style of documentation is different than what I usually use. I'm not certain how uniform the style is across the codebase, is there any reason why you decided to use `{- ... -}`?

R0ss: I copied `css/draw.hs` which used that style for commenting, and made my changes.

PR #797, opened 2016-03-09
Stored all css and js dependencies locally (except Google Analytics)

R0ss:
Stored all css and js dependencies locally (except Google Analytics)


arkon:
CDNs are typically much faster, though.

R0ss:
Yeah, it seems that the local dependencies are much slower...

image

But hey! At least now I can do work on the train (without Internet) :+1:

christinem:
I'm pretty sure we used to store dependencies locally (like almost a year ago) and then we purposely switched to the CDNs, so I don't know if we should go back.

R0ss:
The best solution is to use BOTH:

Like so:

```<script src="//ajax.aspnetcdn.com/ajax/jquery/jquery-2.0.0.min.js"></script>```
```<script>window.jQuery || document.write('<script src="js/jquery-2.0.0.min.js">\x3C/script>')</script>```

This way, the local copies will only be loaded if the CDN is down for whatever reason. :-)

R0ss:
http://requirejs.org/

```requirejs.config({
    enforceDefine: true,
    paths: {
        jquery: [
            '//ajax.aspnetcdn.com/ajax/jquery/jquery-2.0.0.min',
            //If the CDN location fails, load from this location
            'js/jquery-2.0.0.min'
        ]
    }
});
 
//Later
require(['jquery'], function ($) {
});```

PR #798, opened 2016-03-10
Grid: Highlighting for selected courses

cchens:
Some leftover code from last term that I found in the depths of my git stash. Currently uses highlighting to indicate courses from the search list that have been selected, but there was discussion to use a star icon instead.

### DIFF
--- a/public/js/common/objects/react_course.js
+++ b/public/js/common/objects/react_course.js
@@ -1,0 +1,100 @@
+var Course = React.createClass({
+
+    getDefaultProps: function() {
+        var course = getCourse("CSC108H1");
+
+        return {
+            F: course.fallSession,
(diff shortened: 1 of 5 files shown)


public/js/common/objects/react_course.js
```diff
@@ -0,0 +1,100 @@
+var Course = React.createClass({
+
+    getDefaultProps: function() {
+        var course = getCourse("CSC108H1");
```
cchens: I originally passed the course name in from outside the component, but now it seems like `this.props` doesn't work in this function

david-yz-liu:
You can make a new branch which is reset to the previous commit point, and push that new branch to make a pull request. I can close this pull request for you.

PR #801, opened 2016-03-21
Lookup graph by name

R0ss:


hs/Database/CourseQueries.hs
```diff
@@ -127,10 +128,19 @@ buildSession lecs tuts =
 
 -- ** Other queries
 
+-- | Gets Graph Id using graph title
+getGraphId :: String -> IO Int64
```
david-yz-liu: I don't think `getGraphJSON` is used anywhere else; if it isn't, you can change it directly to take in a string rather than Int64.

hs/Database/CourseQueries.hs
```diff
@@ -127,10 +127,15 @@ buildSession lecs tuts =
...
--- | Gets Shape, Text and Path elements for rendering graph returned as JSON
-getGraphJSON :: Int64 -> IO Response
-getGraphJSON gId =
+-- | Looks up a graph using its title then gets the Shape, Text and Path elements 
+-- for rendering graph (returned as JSON).
+getGraphJSON :: String -> IO Response
+getGraphJSON graphName =
     runSqlite databasePath $ do
+        graphEnt :: (Maybe (Entity Graph)) <- selectFirst [GraphTitle ==. graphName] []
+        gId <- case graphEnt of
```
david-yz-liu: Don't use `-1` as a placeholder. Instead, make the `case` expression the outermost.

```haskell
graphEnt :: ...
case graphEnt of
    Nothing -> createJSONResponse ...
    Just gId ->
        sqlTexts :: ...
```

hs/Database/CourseQueries.hs
```diff
@@ -127,44 +127,52 @@ buildSession lecs tuts =
...
-                                    texts
-                                    
-            result = createJSONResponse ["texts" .= (texts ++ regionTexts), 
-                                         "shapes" .= (rects ++ ellipses), 
-                                         "paths" .= (paths ++ regions ++ edges)]
-                                         
-        return result
+        graphEnt :: (Maybe (Entity Graph)) <- selectFirst [GraphTitle ==. graphName] []
+        case graphEnt of
+                Nothing -> return $ createJSONResponse ["texts" .= (),
```
david-yz-liu: 1. The indentation for this entire block is off. Remember to use 4 spaces, and no tabs.
2. Don't use `()`. You can use `[]` if this is supposed to be a list of texts, which I think it is.

R0ss: I tried using `[ ]` initially, but I got an error that the type of `[ ]` was ambiguous. When I tried defining it using: `[] :: Shape` I got other errors, so `( )` was just a lot easier to deal with.

david-yz-liu: `()` may be easier to deal with, but it is not correct. Please try harder to find the right type.

david-yz-liu:
Good job, @R0ss!

PR #790, opened 2016-02-15
Added 2 new server requests and some Ajax

R0ss:
- Created a server response: '/get-json-data', which takes a gID (using a GET: "?gid=1") and returns SVG data corresponding to that ID. 

- Created a second server response: '/draw-json', which is a duplicate of the '/draw' response, except that it accepts user input. It uses user input in order to retrieve SVG data corresponding to an area of study (using an Ajax request to '/get-json-data').

PR #802, opened 2016-03-21
Save drawing

R0ss:
Still working on having the new graphs be updated with a title that include the graph's Id number...


dependencies/HaXml-1.25.3/src/Text/XML/HaXml/Wrappers.hs
```diff
@@ -31,7 +31,7 @@ fix2Args :: IO (String,String)
 fix2Args = do
   args <- getArgs
   when ("--version" `elem` args) $ do
-      putStrLn $ "part of HaXml-"++ VERSION
+      putStrLn $ "part of HaXml-"
```
Ian-Stewart-Binks: Why did you modify HaXml?

hs/Database/Tables.hs
```diff
@@ -140,6 +141,13 @@ FacebookTest
 
 -- ** TODO: Remove these extra types and class instances
 
+-- | JSON SVG data
+data SvgJSON =
+    SvgJSON { texts :: [Text],
```
Ian-Stewart-Binks: See `TODO` above. Is it possible to modify the database schema instead of creating the extra type?

public/js/draw/draw.js
```diff
@@ -389,3 +390,98 @@ function regionClicked(e) {
...
+    var shapes = [];
+    var paths = [];
+    var svgElements = document.getElementById('mySVG').children;
+    for (var i = 0; i < svgElements.length; i++) {
+        if (svgElements[i].tagName == "g" && svgElements[i].id != "regions") {
+            var textObj = "";
+            var gText = svgElements[i].querySelector("text");
+            if (gText !== null) {
+                textObj = {
+                    "graph"     : parseInt(gId),
```
Ian-Stewart-Binks: `'`

public/js/draw/draw.js
```diff
@@ -389,3 +390,98 @@ function regionClicked(e) {
...
+                    "target"    : getClosestText(pathCoords[1], shapes)
+            });
+        }
+    }
+    return JSON.stringify({texts:texts, shapes:shapes, paths:paths});
+}
+
+
+/**
+ * getPathCoords(pathCoords)
```
Ian-Stewart-Binks: Putting the function name in the JSdoc is not only redundant, but also bad from a maintainability and consistency standpoint. When the function is renamed in the future, the JSdoc might not be changed.

hs/Database/CourseInsertion.hs
```diff
@@ -11,16 +11,40 @@ into the database. These functions are used as helpers for the WebParsing module
...
+-- | Insert SVG graph and return its id
+insertGraph :: IO Response
+insertGraph = do
+       gId <- runSqlite databasePath $ insert $ Graph "User generated graph" 256 256
+       return $ createJSONResponse gId
+
+-- | Inserts SVG graph data into Texts, Shapes, and Paths tables
+saveGraphJSON :: String -> IO Response
+saveGraphJSON jsonStr = do
+  print jsonStr
```
Ian-Stewart-Binks: Why is it printed?

dependencies/HaXml-1.25.3/src/Text/XML/HaXml/Wrappers.hs
R0ss: These changes allow `stack repl` to work, because `stack` had a problem with `VERSION`

hs/Database/CourseInsertion.hs
R0ss: Argg, it was debugging code...

public/js/draw/draw.js
R0ss: Good point. I'll fix that.

hs/Database/Tables.hs
R0ss: It might be possible to delete it entirely. I only used it to use it with: `Just (SvgJSON texts shapes paths) -> do 
`

hs/Database/CourseInsertion.hs
```diff
@@ -11,16 +11,40 @@ into the database. These functions are used as helpers for the WebParsing module
...
+import Util.Happstack (createJSONResponse)
+import Database.Persist.Sqlite (fromSqlKey, update, insertMany_, insert_, insert, SqlBackend, (=.), (==.), updateWhere, runSqlite)
 import Database.Tables
+import Data.Aeson
+
+-- | Insert SVG graph and return its id
+insertGraph :: IO Response
+insertGraph = runSqlite databasePath $ do
+    gId <- insert $ Graph "" 256 256
+    update gId [GraphTitle =. ((++) "User generated graph " $ show $ fromSqlKey gId)]
```
R0ss: A new graph is inserted, and persistent generates a new `GraphId` for it, which is used to update the graph's title.

public/js/draw/draw.js
R0ss: ?

Ian-Stewart-Binks: https://github.com/Courseography/courseography/pull/790#discussion_r53579355

R0ss: Ah, I see.

dependencies/HaXml-1.25.3/src/Text/XML/HaXml/Wrappers.hs
david-yz-liu: @R0ss These changes don't belong in this pull request; if you want, you can submit them separately.

hs/Response/Post.hs
```diff
@@ -115,18 +115,14 @@ checkPost =
...
                 H.div ! A.id "specextra" ! A.class_ "more-info" $ do
                     H.input ! A.type_ "text"
                     H.input ! A.type_ "text"
                     H.input ! A.type_ "text"
                     H.input ! A.type_ "text"
             H.div ! A.id "spec_misc" $ do
-                H.p ! A.class_ "code" $ "Any from this list: CSC301H, CSC318H, CSC404H, CSC411H, CSC418H, CSC420H, \
-                    \CSC428H, CSC454H, CSC485H, CSC490H, CSC491H, CSC494H, or PEY (0.5 FCEs) \
-                    \ ** Note: Type 'PEY' for Check my POSt to recognize it **"
+                H.p ! A.class_ "code" $ "Any from this list: CSC301H, CSC318H, CSC404H, CSC411H, CSC418H, CSC420H,  CSC428H, CSC454H, CSC485H, CSC490H, CSC491H, CSC494H, or PEY (0.5 FCEs)   ** Note: Type 'PEY' for Check my POSt to recognize it **"
```
david-yz-liu: Same comment as HaXml - please don't include these changes in this pull request.

public/js/draw/draw.js
```diff
@@ -389,3 +390,95 @@ function regionClicked(e) {
...
+ */
+function convertSvgToJson(gId) {
+    'use strict';
+
+    var texts = [];
+    var shapes = [];
+    var paths = [];
+    var svgElements = document.getElementById('mySVG').children;
+    for (var i = 0; i < svgElements.length; i++) {
+        if (svgElements[i].tagName == 'g' && svgElements[i].id != 'regions') {
```
david-yz-liu: Use `===` and `!==`

public/js/draw/setup.js
```diff
@@ -86,6 +86,30 @@ $('#finish-region').click(function () {
     finishRegion();
     });
 
+$('#save-graph').click(function () {
+        $.ajax({
```
david-yz-liu: Let's indent this by 4 spaces.

hs/Database/CourseInsertion.hs
```diff
@@ -11,16 +11,46 @@ into the database. These functions are used as helpers for the WebParsing module
...
+insertGraph :: String -> IO Response
+insertGraph nameStr = runSqlite databasePath $ do
+    nameExists <- selectFirst [GraphTitle ==. nameStr] []
+    case nameExists of
+      Nothing -> do
+        gId <- insert $ Graph nameStr 256 256
+        return $ createJSONResponse gId
+      Just _ -> do
+        gId <- insert $ Graph "" 256 256
+        update gId [GraphTitle =. ((++) "User generated graph " $ show $ fromSqlKey gId)]
```
david-yz-liu: For now, let's not insert anything in the case that the graph name is already taken.

hs/Database/Tables.hs
david-yz-liu: I think we need a separate type to create a `FromJSON` instance.

hs/DevelopmentConfig.hs
```diff
@@ -123,4 +123,4 @@ holidays = ["20151012T", "20151109T", "20151110T",
 
 -- | Enable Facebook integration. Should only be true on the production server.
 enableFb :: Bool
-enableFb = False
+enableFb = True
```
david-yz-liu: Don't change this.

public/js/draw/setup.js
```diff
@@ -86,19 +86,47 @@ $('#finish-region').click(function () {
     finishRegion();
     });
 
+$('#save-graph').click(function () {
```
david-yz-liu: There should just be one AJAX request here, not two.

public/js/draw/setup.js
```diff
@@ -86,19 +86,47 @@ $('#finish-region').click(function () {
...
-            data: {gid : $('#area-of-study').val()},
-            dataType: 'json', 
-            success: function(data) {
-                $('#json-data').html('<pre>' + JSON.stringify(data) + '<pre>');
-            },
-            error: function(xhr, status, err) {
-                console.error('graphs', status, err.toString());
+$('#save-graph').click(function () {
+    $.ajax({
+        url: 'insert-graph',
```
david-yz-liu: You should be able to do this in one AJAX request, not two. You'll need to set graph IDs on the server, not here.

R0ss: The only reason I separate it into two Ajax requests is because of the situation where the user enters an invalid graph name multiple times. I don't want to send the us in data until a valid name is entered.

R0ss: The json data*

david-yz-liu: That's not a bad thought, but the way this is set up now, it is possible for a graph to be inserted while its corresponding elements are not. E.g., if the JSON data is not parsed properly, which actually also happened to me while I was testing this.

It will be possible in the future to check for duplicate names through some other method.

david-yz-liu:
Also make sure to do a pull from Courseography's master branch so that you don't have any merge conflicts, and test your code thoroughly.

R0ss:
Ah, I see.

public/js/draw/draw.js
```diff
@@ -389,3 +390,107 @@ function regionClicked(e) {
...
+        return a[1] === b[1] ? 0 : a[1] < b[1] ? -1: 1
+    })[0][0];
+}
+
+
+/**
+ * saveGraph(jsonData)
+ * Insert JSON data into Persistent graph table.
+ */
+function saveGraph(jsonData) {
```
david-yz-liu: I think you can remove this function.

public/js/draw/draw.js
```diff
@@ -389,3 +390,107 @@ function regionClicked(e) {
...
+                    'graph'     : parseInt(gId),
+                    'id_'       : '',
+                    'pos'       : [ parseFloat(gRect.getAttribute('x')),
+                                    parseFloat(gRect.getAttribute('y'))
+                                    ],
+                    'width'     : parseFloat(gRect.getAttribute('width')),
+                    'height'    : parseFloat(gRect.getAttribute('height')),
+                    'fill'      : 'none',
+                    'stroke'    : '',
+                    'text'      : [textObj],
```
david-yz-liu: Use your `texts` variable from above.

public/js/draw/draw.js
```diff
@@ -389,3 +390,107 @@ function regionClicked(e) {
...
+    var texts = [];
+    var shapes = [];
+    var paths = [];
+    var svgElements = document.getElementById('mySVG').children;
+    for (var i = 0; i < svgElements.length; i++) {
+        if (svgElements[i].tagName === 'g' && svgElements[i].id !== 'regions') {
+            var textObj = '';
+            var gText = svgElements[i].querySelector('text');
+            if (gText !== null) {
+                textObj = {
```
david-yz-liu: You don't need a temporary variable at all - just push the object literal.

public/js/draw/draw.js
```diff
@@ -389,3 +390,107 @@ function regionClicked(e) {
...
+        } else if (svgElements[i].tagName === 'path') {
+            var pathCoords = getPathCoords(svgElements[i].getAttribute('d'));
+            paths.push({
+                    'graph'     : parseInt(gId),
+                    'id_'       : '',
+                    'points'    : pathCoords,
+                    'fill'      : 'none',
+                    'stroke'    : '',
+                    'isRegion'  : false,
+                    'source'    : getClosestText(pathCoords[0], shapes),
```
david-yz-liu: It may be that `getClosestText` returns undefined, when the source/target node doesn't have a label.

This is a larger problem to deal with later, but for now please just set this function to return the empty string rather than undefined (this will fix another parsing problem).

app/Database/CourseInsertion.hs
```diff
@@ -11,16 +11,35 @@ into the database. These functions are used as helpers for the WebParsing module
...
+import Data.Aeson
+
+-- | Inserts SVG graph data into Texts, Shapes, and Paths tables
+saveGraphJSON :: String -> String -> IO Response
+saveGraphJSON jsonStr nameStr = do
+    let jsonObj = decode $ BSL.pack jsonStr
+    case jsonObj of
+        Nothing -> return $ toResponse ("Error" :: String)
+        Just (SvgJSON texts shapes paths) -> do
+            response <- runSqlite databasePath $ do insertMany_ texts
```
david-yz-liu: style comment: start the first action after the `do` on a new line, so you only indent 4 spaces

app/Database/CourseInsertion.hs
```diff
@@ -11,16 +11,35 @@ into the database. These functions are used as helpers for the WebParsing module
...
+                                                    insertMany_ shapes
+                                                    insertMany_ paths
+                                                    gId <- insert $ Graph nameStr 256 256
```
david-yz-liu: You're on the right track, but you should really use record updating.

If you have a text `t` and you want to change the graph field, keep in mind that the actual name of the field is `textGraph`, so you can do something like `t { textGraph = gId }`. Do that for all of `texts`, `shapes`, and `paths` before inserting into the database, so you won't have to later do an update.

PR #803, opened 2016-03-24
Check My POSt Info Modals

christinem:
This PR changes Check My POSt so that when clicking an individual course, it now opens up the course modal that opens up when clicking a tool tip in the graph. This is in attempt to prevent all the GET requests to gather each course title before,


david-yz-liu:
@christinem Thanks - please don't forget about the course name-specific links for the modals.

PR #806, opened 2016-03-25
Removed trailing whitespace from Loading.hs

R0ss:


R0ss:
I set `"ensure_newline_at_eof_on_save": true,` in Sublime, so this shouldn't happen in the future.

PR #813, opened 2016-04-06
Check My POSt Course Links

christinem:
Makes each course an independent link that opens its own course modal.


david-yz-liu:
Thanks, @christinem!

PR #817, opened 2016-04-20
Created colour wheel for draw tool

R0ss:


david-yz-liu:
Instead of the color wheel image storing data in css, why don't we use something a little more programmatic like http://jscolor.com/

app/Css/Draw.hs
```diff
@@ -26,6 +28,20 @@ drawStyles = do
...
+    "#colour-table" ? do
+        height (px 40)
+        width  (px 200)
+        forM_ (zipWith3 (\func xyPos color -> func xyPos color)
+                        (Prelude.repeat (\[x, y] rgb -> tr # nthChild x ** td # nthChild y ? do background rgb))
+                        [Prelude.map (\z -> T.pack $ show z) [x, y] | x <- [1 .. 2], y <- [1 .. 5]]
+                        -- [2x5] Colour Pallet --
+                        [crimson,       sienna,    tomato,     navy,       lightskyblue,
+                         paleturquoise, wheat,     seagreen,   lightcoral, moccasin    ])
+              (\css -> css)
```
david-yz-liu: This is the `id` function

app/Css/Draw.hs
```diff
@@ -26,6 +28,20 @@ drawStyles = do
...
     finishRegionCSS
 
+{- The colour table. -}
+colourTableCSS :: Css
+colourTableCSS =
+    "#colour-table" ? do
+        height (px 40)
+        width  (px 200)
+        forM_ (zipWith3 (\func xyPos color -> func xyPos color)
+                        (Prelude.repeat (\[x, y] rgb -> tr # nthChild x ** td # nthChild y ? do background rgb))
```
david-yz-liu: You shouldn't need a `do` if there's only one expression

app/Css/Draw.hs
```diff
@@ -26,6 +28,20 @@ drawStyles = do
...
+                        [Prelude.map (\z -> T.pack $ show z) [x, y] | x <- [1 .. 2], y <- [1 .. 5]]
```
david-yz-liu: You can use function composition to simplify the lambda - `(T.pack . show)`

app/Response/Draw.hs
```diff
@@ -37,15 +38,17 @@ modePanel = H.div ! A.id "side-panel-wrap" $ do
...
     H.div ! A.id "path-mode" ! A.class_ "mode" $ "PATH (p)"
     H.div ! A.id "region-mode" ! A.class_ "mode" $ "REGION (r)"
     H.div ! A.id "finish-region" ! A.class_ "button" $ "finish (f)"
     H.div ! A.id "change-mode" ! A.class_ "mode" $ "SELECT/MOVE (m)"
     H.div ! A.id "erase-mode" ! A.class_ "mode" $ "ERASE (e)"
+    H.input ! A.id "select-colour"
+            ! A.class_ "jscolor"
+            ! A.value "ab2567"
+            ! A.size "15"
+    H.table !A.id "colour-table" $ forM_ ([replicate 5 "", replicate 5 ""] :: [[String]])
```
david-yz-liu: Missing a space after `!`

david-yz-liu: Hmm string literals can be interpreted as `Html` type, so I wonder if you can do `[[Html]]` so that you don't need to call `H.toHtml` below

public/js/draw/jscolor.min.js
```diff
@@ -0,0 +1,10 @@
+/**
```
david-yz-liu: Make sure to put all external libraries in `public/js/vendor`

david-yz-liu:
@R0ss please do a merge from master. You'll need to resolve conflicts from your other pull requests, and there seems to be a new use of `nodeColourId`.

app/Css/Draw.hs
R0ss: Cool!

public/js/draw/draw.js
```diff
@@ -87,7 +87,7 @@ function makeNode(posX, posY, jsonObj) {
     } else {
         var nodeWidth = 40;
         var nodeHeight = 32;
-        var nodeFill = nodeColourId;
+        var nodeFill = $('#select-colour').css('background-color');
```
david-yz-liu: It seems like here you want to use the colour picker's `value` attribute (that's typically what we do for inputs).

public/js/draw/draw.js
```diff
@@ -109,6 +109,8 @@ function makeNode(posX, posY, jsonObj) {
         node.setAttribute('id', nodeId_);
         node.setAttribute('width', nodeWidth);
         node.setAttribute('height', nodeHeight);
+        node.setAttribute('class', 'node');
+        node.setAttribute('fill', $('#select-colour').css('background-color'));
```
david-yz-liu: You may notice that the `fill` is set above, and with the proper variable.

public/js/draw/draw.js
```diff
@@ -432,7 +434,7 @@ function finishRegion() {
 
     if (mode === 'region-mode' && curPath !== null && curPath.elbows.length >= 3) {
         curPath.setAttributeNS(null, 'd', curPath.getAttribute('d') + 'Z');
-        curPath.setAttributeNS(null, 'data-group', nodeColourId);
```
david-yz-liu: I didn't notice this earlier, but I believe this old attribute was used in the CSS to actually fill in the regions with different colours. Look inside the css files to see if there are any rules that can be removed.

R0ss: All the files within the CSS folder are clean.

david-yz-liu:
Good job, @R0ss!

PR #822, opened 2016-04-26
React Graph Modal Fix

christinem:
This PR fixes the short cut used to get the modals on the graph page to work using React, and now works within the require module system.


david-yz-liu:
Thanks, @christinem!

PR #823, opened 2016-04-26
React post

david-yz-liu:
Merges the work of @christinem in converting the Check My POSt view to use React.


david-yz-liu:
@christinem some comments

1. Need to change the path for `require.js`. Notice that it's been moved in master (you may not have deleted the file properly yourself).
2. There's a requiest for file `graph/objects/node.js` that doesn't exist anymore.
3. I just tried it out, and it looks like the "Inquiry" field doesn't work quite right. First, you can't type PEY into it (disabled). Second, if just CSC301 is selected in the graph, it doesn't cause the category to highlight.

PR #824, opened 2016-05-04
React Post fixes

christinem:
This pull request fixes the comments by @david-yz-liu on PR #823.


public/js/post/course_components.js
```diff
@@ -252,7 +252,8 @@ export var InquiryCategory = React.createClass({
     getInitialState: function() {
         return {
             completed: false,
-            infoOpened: false
+            infoOpened: false,
```
david-yz-liu: Should this really be false?

PR #830, opened 2016-05-11
Course Table metadata retrieved from JSON object.

R0ss:


david-yz-liu:
@R0ss This ran fine on my machine, so it's probably a file system issue on yours. Try deleting the database file, re-running `stack exec courseography graphs`, and then try again.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+
+{- URL to UofT courses (stored as JSON string) -}
+jsonURL :: String
+jsonURL = "https://timetable.iit.artsci.utoronto.ca/api/courses?org=&code=csc108"
+
+{- Decode JSON string to JSON object -}
+getJSON :: IO (Maybe Object)
+getJSON = do
+    resp <- simpleHttp jsonURL
+    return $ (decode resp :: Maybe Object)
```
david-yz-liu: If you use parentheses, you don't need the `$`.

I'm surprised you need a type annotation here - do you get an error if you omit it?

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+
+{- Extract course metadata (values) from JSON object -}
+getMetadata :: IO (Maybe (M.HashMap k v)) -> IO [v]
+getMetadata = fmap (\(Just v) -> M.elems v)
```
david-yz-liu: Couple things: 

1. You can use `$` after `fmap`.
2. Generally, if a function's type is `IO a -> IO b` and its top level is `fmap`, then I'd rather see the function defined as `a -> b` and then applied using `fmap`.
3. The pattern matching isn't great. If you really want to return a list, probably use the `maybe` function to either get `v` or an empty hash map.
But

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+import Data.List
+import Database.Tables
+import WebParsing.PrerequisiteParsing
+import Database.Persist.Sqlite (runSqlite, insertMany_)
+import qualified Data.HashMap.Lazy as M
+import qualified Data.Text as T
+import Config (databasePath)
+import Network.HTTP.Conduit (simpleHttp)
+
+{- URL to UofT courses (stored as JSON string) -}
```
david-yz-liu: Our documentation style isn't great, but please use the Haddock syntax `-- | My comment` instead.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+        -- [Maybe (Value a)] -> [Maybe a]
+        return $ fmap (\(Just (String v)) -> Just v) searchResults
+
+{- Retrieve all Course Table metadata -}
+getAllCourses :: IO ()
+getAllCourses = do
+    print ("parsing JSON data from: " ++ jsonURL)
+    allMetadata <- getMetadata getJSON
+    let codes'                    = lookupField "code"                     allMetadata
+        codes                     = map (\(Just c) -> c) codes'
```
david-yz-liu: Again, not great use of pattern matching. I would look at this page for ideas: https://hackage.haskell.org/package/base-4.8.0.0/docs/Data-Maybe.html

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+    let courseTitles              = lookupField "courseTitle"              allMetadata
+        courseDescriptions        = lookupField "courseDescription"        allMetadata
+        prereqs                   = fmap        parsePrerequisites         prereqStrings
+        exclusions                = lookupField "exclusion"                allMetadata
+        breadths                  = lookupField "breadthCategories"        allMetadata
+        distributions             = lookupField "distributionCategories"   allMetadata
+        prereqStrings             = lookupField "prerequisite"             allMetadata
+        coreqs                    = lookupField "corequisite"              allMetadata
+    let manualTutorialEnrolments  = repeat      $ Just False
```
david-yz-liu: You can only properly set these if you parse the sections associated with the course

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+                                                    Courses c (lst !! 0) (lst !! 1) mTutEnrl mPratEnrl (lst !! 2)
+                                                            (lst !! 3) (lst !! 4) (lst !! 5) (lst !! 6) corqs vUrl)
+                                                    codes
+                                                    lst7
+                                                    coreqs
+                                                    manualTutorialEnrolments
+                                                    manualPracticalEnrolments
+                                                    videoUrls
+
+    print ("All Courses have been successfully inserted")
```
david-yz-liu: no need for the parentheses

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,80 @@
...
+    let lst7 = zipWith7 (\a b c d e f g -> a:b:c:d:e:f:g:[])
+                         courseTitles
+                         courseDescriptions
+                         prereqs
+                         exclusions
+                         breadths
+                         distributions
+                         prereqStrings
+
+    print ("inserting " ++ (show $ length lst7) ++ " courses into database")
```
david-yz-liu: Probably better to use `$`

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,157 @@
...
+import Database.Persist.Sqlite (runSqlite, insertMany_)
+import qualified Data.HashMap.Lazy as M
+import qualified Data.Text as T
+import Config (databasePath)
+import Network.HTTP.Conduit (simpleHttp)
+
+-- | Converts 24-hour time into a double
+-- | Assumes times are rounded to the nearest hour
+getHourVal :: String -> Double
+getHourVal time = (read $ take 2 time :: Double)
```
david-yz-liu: I suppose you'll modify this to handle half-hours, but also you don't need parentheses around the body of a function.

I also doubt you need an explicit type annotation, because there's already one in the function annotation.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,157 @@
...
+import WebParsing.PrerequisiteParsing
+import Database.Persist.Sqlite (runSqlite, insertMany_)
+import qualified Data.HashMap.Lazy as M
+import qualified Data.Text as T
+import Config (databasePath)
+import Network.HTTP.Conduit (simpleHttp)
+
+-- | Converts 24-hour time into a double
+-- | Assumes times are rounded to the nearest hour
+getHourVal :: String -> Double
```
david-yz-liu: In general, use "to___" for naming conversion functions.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,157 @@
...
+-- | Takes a day and start/end times then generates a set of 30-minute timeslots
+getTimeSlots :: String -> String -> String -> [Time]
+getTimeSlots day start end = do
+    let dayDbl = getDayVal day
+        startDbl = getHourVal start
+        endDbl = getHourVal end
+    [Time [dayDbl, timeDbl] | timeDbl <- [startDbl, (startDbl + 0.5) .. (endDbl - 0.5)]]
+
+getNestedTimeSlots :: [Maybe T.Text] -> [Maybe T.Text] -> [Maybe T.Text] -> [[Time]]
+getNestedTimeSlots [Nothing] [Nothing] [Nothing] = [[]]
```
david-yz-liu: It seems this function assumes the input lists are just a single `Nothing`, or a list of `Just`s. If that's the case, then the proper input type isn't `[Maybe T.Text]`, but instead `Maybe [T.Text]`.

On the other hand, perhaps the lists could contain a mixture of Justs and Nothings, which might mean there shouldn't be two pattern-matching rules, but instead a more robust function passed to `zipWith3`.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,157 @@
...
+        return $ fmap (\(Just (Object v)) -> Just v) searchResults
+
+ -- | Retrieve all Course Table metadata
+getAllCourses :: IO ()
+getAllCourses = do
+    print ("parsing JSON data from: " ++ jsonURL)
+    allMetadata <- getMetadata getJSON
+
+    let codes'                    = lookupField "code"                                       allMetadata
+        codes                     = map (\(Just c) -> c)                                     codes'
```
david-yz-liu: Look up `fromJust` (and other helpful functions in Data.Maybe)

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+{-
```
david-yz-liu: Put this back in.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+module WebParsing.UtsgJsonParser
+     (getJSON) where
+-}
+
+import Data.Aeson
```
david-yz-liu: Please only import the functions/types you need (other than for `Database.Tables`).

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+import Database.Persist.Sqlite (runSqlite, insertMany_, insert_)
+import qualified Data.HashMap.Lazy as M
+import qualified Data.Text as T
+import Config (databasePath)
+import Network.HTTP.Conduit (simpleHttp)
+
+-- | Converts 24-hour time into a double
+-- | Assumes times are rounded to the nearest hour
+getHourVal :: String -> Double
+getHourVal time = (read $ take 2 time :: Double) + (/) (read $ drop 3 time :: Double) 60
```
david-yz-liu: I'm not sure why you didn't use `/` inline.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+
+-- | Converts a weekday into a double
+-- | Monday to Friday becomes 0.0 to 4.0
+getDayVal :: String -> Double
+getDayVal day = case day of
```
david-yz-liu: Use pattern-matching on the function definition.

```
getDayVal "MO" = ...
getDayVal "TU" = ...
...
```

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+jsonURL = "https://timetable.iit.artsci.utoronto.ca/api/courses?org=&code=PRT100Y"
+
+ -- | Decode JSON string to JSON object
+getJSON :: IO (Maybe Object)
+getJSON = do
+    resp <- simpleHttp jsonURL
+    return $ decode resp
+
+-- | Extract course metadata (values) from JSON object
+-- | getMetadata :: IO (Maybe (M.HashMap k v)) -> IO [v]
```
david-yz-liu: This should be a pure function (no need for `IO` in the type).

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+getMetadata = fmap (\(Just v) -> M.elems v)
```
david-yz-liu: Also, pattern-matching on just a "Just" is unsafe. I'm guessing you want to return an empty list in the case of `Nothing`.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+
+getValues = fmap (\(Just v) -> M.elems v)
+
+-- | Extract course metadata (keys) from JSON object
+-- | getKeys :: [Maybe Object] -> [T.Text]
+getKeys = fmap (\(Just k) -> M.keys k)
+
+ -- | Lookup field (by name) within metadata and then "expose" results.
```
david-yz-liu: Remove the extra space at the beginning of this line.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+--  | lookupField :: Functor f => Data.Text.Internal.Text -> f Value -> f (Maybe Data.Text.Internal.Text)
+lookupField fieldName = do
+        searchResults <- fmap (\val -> case val of
```
david-yz-liu: Pull out the small anonymous functions like this into helpers; you can put the helpers in a `where` block is they're specific to this function.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -0,0 +1,251 @@
...
+                                                                                            startLst
+                                                                                            endLst)
+                                                                      courseDays
+                                                                      courseStarts
+                                                                      courseEnds)
+                                             meetingDays
+                                             meetingStartTimes
+                                             meetingEndTimes
+
+    let manualTutorialEnrolments  = repeat      $ Just False
```
david-yz-liu: Compute these two *after* computing the lecture objects. The point is to search through the corresponding list of lecture objects, and see if there's a "tutorial" section and a "laboratory" section. You can do this using a `map`.

PR #832, opened 2016-05-11
Post Schema Initial Setup

christinem:
Set up initial basic tables for Posts and Post Categories. Also set up initial parsing overhead to work when running `stack exec courseography database`. At the current moment, all it does is run two function calls, one that inserts a new post and one that inserts a new post category. Tested it by commenting out all function calls in `parseAll` except for `parsePosts` (since the rest of the parsing is currently broken with the new timetable format), and ran the above stack command, and can confirm it properly creates the new tables and adds the new values.

### DIFF
--- a/app/Database/PostInsertion.hs
+++ b/app/Database/PostInsertion.hs
@@ -19,7 +19,6 @@ import Data.Aeson
 insertPost :: T.Text -> T.Text -> T.Text -> IO ()
 insertPost departmentName postName postCode =
     runSqlite databasePath $ do
-        runMigration migrateAll
         insert_ $ Post postName departmentName postCode
 
 -- | Insert a new post category into the database
--- a/app/Database/PostInsertion.hs
+++ b/app/Database/PostInsertion.hs
@@ -5,15 +5,9 @@ module Database.PostInsertion(
     insertPost, insertPostCategory) where
 
 import qualified Data.Text as T
-import qualified Data.ByteString.Lazy.Char8 as BSL
-import Happstack.Server.SimpleHTTP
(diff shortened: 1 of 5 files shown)


app/Database/PostInsertion.hs
```diff
@@ -0,0 +1,31 @@
...
+import Config (databasePath)
+import Database.Persist.Sqlite (selectFirst, fromSqlKey, toSqlKey, insertMany_, insert_, insert, SqlBackend, (=.), (==.), updateWhere, runSqlite, runMigration)
+import Database.Tables
+import Data.Aeson
+
+-- | Insert a new post into the database
+insertPost :: T.Text -> T.Text -> T.Text -> IO ()
+insertPost departmentName postName postCode =
+    runSqlite databasePath $ do
+        runMigration migrateAll
```
david-yz-liu: You don't need to run the migration every time; you can remove this line.

app/WebParsing/PostParser.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE OverloadedStrings #-}
+module WebParsing.PostParser
+    (parsePosts) where
+
+import Network.HTTP
```
david-yz-liu: Look carefully at what dependencies are actually needed. It's a good practice to start from scratch and just add the dependencies as you need them. (So remove the ones that aren't necessary.)

david-yz-liu:
@christinem I'd like to merge this in, if you have time to clean it up a little first. The basic functionality is acceptable.

christinem:
@david-yz-liu sorry for the lateness, but I've removed the unnecessary inputs and the unnecessary migration line

PR #833, opened 2016-05-11
Reactgraphjs [Issue 812]

klmerryn:
Refactor repetitive node lookups into helper function.


public/js/graph/react_graph.js
```diff
@@ -1,5 +1,22 @@
...
+/**
+ * 
+ * Search for target node in list of nodes,
+ * or if node not found search through list of bools.
+ * @param {Node} targetNode
+ * @returns {Node}
+ *
+ */
+
+function refLookUp(targetNode) {
```
david-yz-liu: Generally try not to use `this` when there isn't a clear class.

In this particular case, you can probably just pass in `svg` as an argument.

david-yz-liu:
@klmerryn You have the right idea, but I suggested a different signature to simplify the code.

public/js/graph/react_graph.js
```diff
@@ -1,5 +1,22 @@
...
+ * 
+ * Search for target node in list of nodes,
+ * or if node not found search through list of bools.
+ * @param {React.PropTypes.node} targetNode
+ * @param {React.PropTypes.element} elem
+ * @returns {React.PropTypes.Node} 
+ *
+ */
+function refLookUp(targetNode, elem) {
+    var svg = elem;
```
david-yz-liu: This line is unnecessary; just name the parameter `svg`.

public/js/graph/react_graph.js
```diff
@@ -579,8 +595,7 @@ var Node = React.createClass({
             this.setState({status: newState}, function () {
                 setCookie(nodeId, newState);
                 this.props.childs.forEach(function (node) {
-                    var currentNode = svg.refs['nodes'].refs[node] ||
-                                      svg.refs['bools'].refs[node];
+                    var currentNode = refLookUp(node, this.props.svg);
```
david-yz-liu: @klmerryn I should have caught this sooner, but this doesn't work. Use of `this` inside an anonymous function no longer refers to the "outer this" (in this case, the Node instance). Note that the original code did not refer to `this` at all.

Also, ask Christine/Ryan about this problem, as they've come across it before.

Note: this happens in multiple places.

PR #834, opened 2016-05-11
React-grid: fetch course information

cchens:
Uses an ajax call to fetch course information. Currently displays lecture information in the left panel.

### DIFF
--- a/public/js/common/objects/react_course.js
+++ b/public/js/common/objects/react_course.js
@@ -1,48 +1,5 @@
 var Course = React.createClass({
 
-    getDefaultProps: function() {
-        var course = getCourse("CSC108H1");
-
-        return {
-            F: course.fallSession,
-            S: course.springSession,
-            Y: course.yearSession,
-            name: course.name,
-            title: course.title,
-            prereqs: course.prereqs,
-            prereqString: course.prereqString,
-            coreqs: course.coreqs,
-            breadth: course.breadth,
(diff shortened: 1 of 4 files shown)


public/js/common/objects/react_course.js
```diff
@@ -0,0 +1,92 @@
...
+
+    $.ajax({
+        url: 'course',
+        dataType: 'json',
+        data: {name : courseName},
+        async: false,
+
+        success: function (data) {
+            React.render(
+                <Course F={data.fallSession}
```
david-yz-liu: I'm surprised you didn't just pass in the `data` as props, like you do for the `section` component. Seems a lot easier.

david-yz-liu:
@cchens thanks for sharing the code. You're definitely on the right track - keep it up.

I think it's an easy win if you are able to use the classes which already exist so that the section names are styled nicely - I don't think it should be hard to do, and should be quite satisfying.

public/js/common/objects/react_course.js
```diff
@@ -0,0 +1,77 @@
...
+
+    'use strict';
+
+    var course;
+
+    $.ajax({
+        url: 'course',
+        dataType: 'json',
+        data: {name : courseName},
+        async: false,
```
david-yz-liu: You can remove this line (we should have removed it a long time ago).

PR #836, opened 2016-05-18
Post Parsing

christinem:
This PR builds on #832. It is an initial start to parsing the programs in each department page. Can currently parse from the tag where the programs on the page start and print out the resulting list of tags. Work in Progress.


david-yz-liu:
@christinem You can do a pull from master, as I have merged in your earlier pull request.

You should remove all the duplicate code `PostParser.hs`. Instead, just define a function which you can import into `ArtSciParser.hs`, adding to the bottom of the function `parseArtSci`.

app/WebParsing/PostParser.hs
```diff
@@ -1,11 +1,49 @@
...
+        isNotComment (TagComment _) = False
+        isNotComment _ = True
+        secondH2 tags =
+            let sect = sections (isTagOpenName "h2") tags
+            in
+                if null sect
+                then
+                    []
+                else
+                    takeWhile (~/= ("<a name=courses>" :: String)) $ sect !! 1
```
david-yz-liu: `!!` is an unsafe function, and should be used with caution (I know we covered it in CSC324, but still). It raises an exception when it fails.

Instead if checking if `sect` is empty, why not check if its length is less than 2?

david-yz-liu: Look at [tagOpenAttrLit](https://hackage.haskell.org/package/tagsoup-0.13.8/docs/Text-HTML-TagSoup-Match.html#v:tagOpenAttrLit).

app/WebParsing/PostParser.hs
```diff
@@ -1,11 +1,49 @@
...
+        isPostName (TagOpen _ attrs) = any (\x -> fst x == "name" && T.length (snd x) == 9) attrs
```
david-yz-liu: I think tagOpenAttrNameLit will work instead of defining your own function (check docs from above).

david-yz-liu:
@christinem Okay, I'd like you to make the changes I've suggested, and then I will merge this in.

The parsing is still not correct, but I don't want you to go further with this approach. So, I will make an issue about incorrect parsing, and allow you to continue working on just that portion.

I want you to use the *parsing combinators* defined in [Parsec](https://wiki.haskell.org/Parsec). It's extremely popular, and you can find lots of good documentation online.

christinem:
@david-yz-liu Thanks David. I made the changes based on your comments so you can merge this in. I'll start using the new approach you suggested starting now.

david-yz-liu:
Thanks, @christinem!

PR #837, opened 2016-05-18
Graph controls

bell-kelly:
Updating react graph with some zoom functionality.


app/Css/Graph.hs
```diff
@@ -223,6 +223,22 @@ graphContainer = do
         "shape-rendering" -: "geometricPrecision"
     ".highlight-nodes" ? do
         backgroundColor grey
+    "#zoom-in-button" ? do
+        cursor pointer
```
david-yz-liu: Create a `graph-button` class to factor out the common styles.

public/js/graph/react_graph.js
```diff
@@ -125,7 +125,9 @@ var Graph = React.createClass({
             boolsJSON: [],
             edgesJSON: [],
             highlightedNodes: [],
-            fceCount: 0
+            fceCount: 0,
+            zoomFactor: 1
+
```
david-yz-liu: Remove this blank line.

public/js/graph/react_graph.js
```diff
@@ -303,11 +306,41 @@ var Graph = React.createClass({
...
+            preserveAspectRatio: 'xMinYMin'
+        };
 
         return (
+            <div>
+            <Button
+                id='zoom-in-button'
+                sourceImg="static/res/ico/in.png"
+                mouseDown={this.incrementZoom}
+                mouseDownArg={true}/>
```
david-yz-liu: You usually don't want to pass in a separate prop for an argument. You can change the mouseDown to be an anonymous function (`() => ...`) whose body calls `this.incrementZoom` on the correct argument.

public/js/graph/react_graph.js
```diff
@@ -331,11 +364,28 @@ var Graph = React.createClass({
...
+    handleMouseDown: function() {
+        this.props.mouseDown(this.props.mouseDownArg);
+    },
+
+    render: function() {
+
+        return (
+         <div id={this.props.id}>
+        <img id={this.props.id} onMouseDown={this.handleMouseDown}
+        src={this.props.sourceImg}/>
```
david-yz-liu: You need to provide an `alt` attribute; this can be the id for now.

public/js/graph/react_graph.js
```diff
@@ -331,11 +364,28 @@ var Graph = React.createClass({
...
+
+    handleMouseDown: function() {
+        this.props.mouseDown(this.props.mouseDownArg);
+    },
+
+    render: function() {
+
+        return (
+         <div id={this.props.id}>
+        <img id={this.props.id} onMouseDown={this.handleMouseDown}
```
david-yz-liu: You should never duplicate ids. So you'll need to figure out which one you want to have the id, I'm assuming for styling purposes.

public/js/graph/react_graph.js
```diff
@@ -331,11 +364,28 @@ var Graph = React.createClass({
...
-            </svg>
+            </svg></div>
+
         );
     }
 });
 
+var Button = React.createClass({
+
+    handleMouseDown: function() {
```
david-yz-liu: If you do what I said above, you'll no longer need this function.

public/js/graph/react_graph.js
david-yz-liu: Also the indentation is off here.

david-yz-liu:
@bell-kelly this is quite good, well done. I will merge this in after you have made the changes I suggested. After that, you can work on the panning buttons.

bell-kelly:
Thanks for the feedback, David. I won't have much time to make changes until Sunday, but at that point I should be able to implement all of these changes and work on adding the panning functionality as well. I will update the pull request then.

public/js/graph/react_graph.js
bell-kelly: I haven't been able to remove this function successfully yet, but I will keep trying.

david-yz-liu: You'll need to change the `mouseDown` prop before removing the function (see above comment).

bell-kelly: Thanks! I finally figured out what I was doing wrong.

david-yz-liu:
@bell-kelly You should do a pull from master; I have updated the width and height now.

You should incorporate this new state into your existing code, and set limits to the panning. This should be your top priority; I will merge it in after that last piece is done. Then you can create a new branch to continue working on the aesthetic improvements for the buttons.

public/js/graph/react_graph.js
```diff
@@ -303,11 +307,75 @@ var Graph = React.createClass({
...
+        }
+    },
+
     render: function () {
         // not all of these properties are supported in React
-        var svgAttrs = {width: this.props.width, height: this.props.height};
+        var svgAttrs = {
+            width: '100%',
+            height: '100%',
+            viewBox: (this.state.horizontalPanFactor + ' ' +
```
david-yz-liu: I believe the outer parentheses are not necessary.

public/js/graph/react_graph.js
```diff
@@ -303,11 +307,75 @@ var Graph = React.createClass({
...
+                      (this.props.width * this.state.zoomFactor) + ' ' +
+                      (this.props.height * this.state.zoomFactor)),
+            preserveAspectRatio: 'xMinYMin'
+        };
 
         return (
+            <div>
+            <Button
+                divId='zoom-in-button'
+                imgId='zoom-in'
```
david-yz-liu: The images don't need ids.

public/js/graph/react_graph.js
```diff
@@ -331,11 +399,24 @@ var Graph = React.createClass({
                     edgesJSON={this.state.edgesJSON}
                     svg={this}/>
                 <EdgeGroup svg={this} ref='edges' edgesJSON={this.state.edgesJSON}/>
-            </svg>
+            </svg></div>
```
david-yz-liu: Put the `</div>` on a new line. You don't need a blank line, though.

public/js/graph/react_graph.js
```diff
@@ -331,11 +399,24 @@ var Graph = React.createClass({
...
+
         );
     }
 });
 
+var Button = React.createClass({
```
david-yz-liu: Look up "pure React classes."

public/js/graph/react_graph.js
```diff
@@ -331,11 +399,24 @@ var Graph = React.createClass({
...
+
+    render: function() {
+        return (
+            <div id={this.props.divId} className='graph-control-button'>
+            <img id={this.props.imgId} alt={this.props.imgId}
```
david-yz-liu: See above comment about `img` id here.

bell-kelly:
I have added the reset button (it wasn't a priority, but it didn't take long and it helped with debugging) and I have mostly completed panning restrictions, but they are not working quite how I want them to. At some zoom levels, I can still pan too far down and to the right and at some zoom levels I am not able to pan far enough to get to the right and bottom edges. The left and top edges of the graph are easy to identify, but I am having a hard time pinning down the right and bottom edges. I believe there are exactly three things I need to consider: graph width/height, zoom factor, and the resize factor caused by the graph automatically fitting into the container. I've accounted for these three things, but maybe I'm doing something wrong in my calculations, or maybe I've forgotten to consider something else that is affecting the graph.

david-yz-liu:
@bell-kelly please do another merge from master and update this pull request - the diff isn't showing properly, meaning the commit history is a bit wonky.

Also, the reset button image file is missing from the commit.

david-yz-liu:
About the panning limits: right now the calculation for `heightToContainerRatio` and `widthToContainerRatio` doesn't take into account the zoom factor, and so are in fact constant. Seems like this is not what you actually want.

bell-kelly:
I'm not sure that I understand how to incorporate the zoomFactor into this ratio calculation. I think that the only time the graph resizes to fit the container is when the zoom factor is 1, so I think that it might be ok that this ratio is constant.

I do think I found a different solution, though. First of all, I wasn't accounting for when the graph increases in size to fit the container. Additionally, the pan factors are affected by the zoom and I was not accounting for this. I'm going to push my current solution because it seems to be mostly working now. I will continue to test a few more cases to make sure. 

One more thing to point out, there is one main issue with the current calculation that I have noticed. In the Statistics and Cell Bio graphs the category lists don't seem to be included in my right edge calculation. It's definitely possible that there's stills something wrong with my calculation, but I'm hoping this issue is a result of the category list not being included in the width of the graph for some reason.

public/js/graph/react_graph.js
```diff
@@ -305,11 +311,111 @@ var Graph = React.createClass({
...
+        var topEdge = -(this.state.verticalPanFactor);
+        var leftEdge = -(this.state.horizontalPanFactor);
+        var bottomEdge = (this.state.height - this.state.verticalPanFactor) / this.state.zoomFactor;
+        var rightEdge = (this.state.width - this.state.horizontalPanFactor) / this.state.zoomFactor;
+
+        // size of container
+        var containerWidth = document.getElementById("react-graph").clientWidth;
+        var containerHeight = document.getElementById("react-graph").clientHeight;
+        
+        // if the graph does not fit in it's container, it is resized by the inverse factor
```
david-yz-liu: wrong `it's`

public/js/graph/react_graph.js
```diff
@@ -305,11 +311,111 @@ var Graph = React.createClass({
...
+        // of the greater of these two ratios.
+        var autoResizeFactor;
+        var heightToContainerRatio = this.state.height / containerHeight;
+        var widthToContainerRatio = this.state.width /containerWidth;
```
david-yz-liu: Put a space after the `/`

public/js/graph/react_graph.js
```diff
@@ -305,11 +311,111 @@ var Graph = React.createClass({
...
+        autoResizeFactor = Math.max(heightToContainerRatio, widthToContainerRatio);
+        bottomEdge /= autoResizeFactor;
+        rightEdge /= autoResizeFactor;
+
+        if (direction === 'up' && topEdge < 0) {
+            this.setState({verticalPanFactor: this.state.verticalPanFactor -= 10});
```
david-yz-liu: Don't use `-=`, just use `-`. Don't use mutating operations on state directly.

public/js/graph/react_graph.js
```diff
@@ -305,11 +311,111 @@ var Graph = React.createClass({
...
+
+        } else if (direction === 'left' && leftEdge < 0) {
+            this.setState({horizontalPanFactor: this.state.horizontalPanFactor -= 10});
+
+        } else if (direction ==='down' && bottomEdge > (containerHeight)) {
```
david-yz-liu: No need for the blank line above, nor the parentheses around `containerHeight`

public/js/graph/react_graph.js
```diff
@@ -305,11 +311,111 @@ var Graph = React.createClass({
...
+            this.setState({verticalPanFactor: this.state.verticalPanFactor += 10});
+
+        } else if (direction === 'right' && rightEdge > (containerWidth)) {
+            this.setState({horizontalPanFactor: this.state.horizontalPanFactor += 10});
+        }
+    },
+
+    resetZoomAndPan: function() {
+        this.setState({
+            zoomFactor:1,
```
david-yz-liu: Put a space after the `:`

david-yz-liu:
@bell-kelly Okay, I guess I'm not sure what the `autoResizeFactor` is supposed to do. You can explain this more at the meeting today. I've made a few minor style comments.

Don't worry about the cut-off text - it's probably a width issue in the SVG itself.

bell-kelly:
Thanks David. I made those changes.

PR #840, opened 2016-05-25
issue-815: Render overridden (red) edges properly [WIP]

nzliu:
attempted to render missing edges separately from the rest of the edges in the edgegroup class (WIP).


public/js/graph/react_graph.js
```diff
@@ -934,24 +940,26 @@ var EdgeGroup = React.createClass({
...
-			if (edges[i].id_ in missEdgesDict) {
-				var edgeID = edges[i].id_;
-	      		if (missEdgesDict[edgeID] !== true) {
-	      			otherEdges.push(edges[i]);
-	      		} else {
-	     			missingEdges.push(edges[i]);
-	     		}
+			var edgeID = edges[i].id_;
+			// console.log(this.state);
+			if (edgeID in this.state) {
```
david-yz-liu: I don't think you ever remove edges from the state. So you should be checking not only for whether the state has the edgeID, but whether `this.state[edgeID]` is missing (or something like that.

david-yz-liu:
@nzliu progress has indeed been made.

I've been playing around with your pull request. It seems that when an edge switches from one partition to the other (`missingEdges` vs. `otherEdges`), it gets recreated, and its status gets reset to `inactive`, so the edge doesn't highlight.

So, new plan. Rather than creating a partition, first create a copy of the edge json array, and then call Javascript's `sort` function, passing in a custom compare function to put the missing edges after the non-missing edges. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Note: even with this change I haven't gotten this fully working, but am continuing to work on it.

david-yz-liu:
Oh, and please configure your text editor to use spaces instead of tabs, and to convert all tabs in your current file to spaces. Use 4 spaces for indentation.

public/js/graph/react_graph.js
```diff
@@ -927,7 +989,16 @@ var Edge = React.createClass({
         }
     },
 
+    componentDidUpdate : function() {
+    	if (this.state.status === 'missing') {
```
david-yz-liu: Make this more general: call `this.props.updateEdgeStatus` if `this.state.status` is different from the *old* state (add two parameters to the function - see documentation for `componentDidUpdate`

david-yz-liu:
Okay, last comment: you'll need to update the `updateStatus` method of `Edge` so that it checks if the target's status is "missing", and if so, sets the edge's status to missing as well. Right now, it only ever sets status as "inactive", "takeable", or "active".

But the good news is, if you do that (and everything else I said) then the status of the edge should stay "missing," and you'll be done. [The problem was the `componentDidUpdate` method of EdgeGroup was calling `updateStatus` on each edge, which overrode the "missing" status.]

public/js/graph/react_graph.js
```diff
@@ -895,19 +932,73 @@ var EdgeGroup = React.createClass({
...
+        var edges = this.props.edgesJSON;
+        var edgesCopy = $.extend([], edges);
+        var state = this.state;
+        edgesCopy.sort(function(a, b) {
+            var aID = a.id_;
+            var bID = b.id_;
+            var aMiss = false;
+            var bMiss = false;
+            if (aID in state) {
+                if (state.aID) {
```
david-yz-liu: 1. You should combine this check with the previous one with an AND
2. Same comment as before. You need to check whether `state.aID` is missing.

public/js/graph/react_graph.js
```diff
@@ -918,7 +1009,9 @@ var Edge = React.createClass({
                      this.props.svg.refs.bools.refs[this.props.source];
         var target = this.props.svg.refs.nodes.refs[this.props.target] ||
                      this.props.svg.refs.bools.refs[this.props.target];
-        if (!source.isSelected()) {
+        if (target.state.status === 'missing') {
```
david-yz-liu: This isn't bad, but will fail if this edge is takeable, but the target has another prerequisite which is missing.

You need to check both the state of the target as well as whether the source is not selected.

public/js/graph/react_graph.js
```diff
@@ -1,7 +1,7 @@
 import * as tooltip from 'es6!graph/tooltip';
 
 /**
- *
+ *
```
david-yz-liu: @nzliu please revert this whitespace change (and the similar ones below).

public/js/graph/react_graph.js
```diff
@@ -612,6 +615,7 @@ var Node = React.createClass({
                     var sourceNode = svg.refs['nodes'].refs[currentEdge.props.source] ||
                                      svg.refs['bools'].refs[currentEdge.props.source];
                     if (!sourceNode.isSelected()) {
+                        // currentEdge.setState({status: 'missing'}, () => svg.refs['edges'].render());
```
david-yz-liu: Please delete this (and remove other commented-out blocks that are no longer necessary)

public/js/graph/react_graph.js
```diff
@@ -875,6 +881,26 @@ var Bool = React.createClass({
 
 
 var EdgeGroup = React.createClass({
+    // EdgeGroup's state is used to keep track of the edgeIDs of
+    // edges that are missing. Void is just a placeholder state so 
+    // we can declare an initial state; it does nothing. 
+    getInitialState: function() {
+        return {void: null};
```
david-yz-liu: I'm still not sure you can't just do `return {}` here

public/js/graph/react_graph.js
```diff
@@ -875,6 +881,26 @@ var Bool = React.createClass({
...
+    // When an edge's state changes and the edge is not undefined, 
+    // it will call updateEdgeStatus and update EdgeGroup's state with its
+    // edgeID and status. This function is passed as a props to Edge.
+    updateEdgeStatus: function(edgeID, state) {
+        var isMissing = true;
+        if (state !== 'missing') {
+            isMissing = false;
+        }
+        var newState = {};
+        newState[edgeID] = isMissing;
```
david-yz-liu: It seems like this variable is unnecessary

public/js/graph/react_graph.js
```diff
@@ -875,6 +881,26 @@ var Bool = React.createClass({
...
+    // we can declare an initial state; it does nothing. 
+    getInitialState: function() {
+        return {void: null};
+    },
+
+    // When an edge's state changes and the edge is not undefined, 
+    // it will call updateEdgeStatus and update EdgeGroup's state with its
+    // edgeID and status. This function is passed as a props to Edge.
+    updateEdgeStatus: function(edgeID, state) {
+        var isMissing = true;
```
david-yz-liu: Please simplify the code to `isMissing = state !== 'missing'`

public/js/graph/react_graph.js
```diff
@@ -895,19 +921,55 @@ var EdgeGroup = React.createClass({
...
+        var state = this.state;
+        edgesCopy.sort(function(a, b) {
+            // If an edge is missing, its edgeID should be in EdgeGroup's
+            // state and its value should be true.
+            var aID = a.id_;
+            var bID = b.id_;
+            var aMiss = false;
+            var bMiss = false;
+            if (aID in state) {
+                if (state[aID]) {
```
david-yz-liu: I understand now that this is a boolean. You should combine this check with the previous one (`aID in state && state[aID]`), and also use it to set `aMiss` directly, rather than using an `if` (similar to above comment). Same with `bMiss`.

public/js/graph/react_graph.js
```diff
@@ -895,19 +921,55 @@ var EdgeGroup = React.createClass({
...
             <g id='edges'>
-                {this.props.edgesJSON.map(this.generateEdge)}
+                {edgesCopy.map(this.generateEdge)}
             </g>
         );
     }
 });
 
-
+
```
david-yz-liu: Please remove this whitespace.

public/js/graph/react_graph.js
```diff
@@ -918,7 +980,29 @@ var Edge = React.createClass({
                      this.props.svg.refs.bools.refs[this.props.source];
         var target = this.props.svg.refs.nodes.refs[this.props.target] ||
                      this.props.svg.refs.bools.refs[this.props.target];
-        if (!source.isSelected()) {
+        // if (target.state.status === 'missing' && !source.isSelected()) {
```
david-yz-liu: Reminder to remove this commented block.

david-yz-liu:
@nzliu it seems to me like this is working properly now! You should do a merge from master (there's a small merge conflict, but I'm sure you can sort it out). I've left a few minor comments, and that's about it.

PR #849, opened 2016-06-08
Adding more functionality to graph control buttons

bell-kelly:


public/js/graph/react_graph.js
```diff
@@ -438,6 +439,39 @@ var Graph = React.createClass({
         });
     },
 
+    // Zoom and pan function calls work once, but they are not repeated while mouse is down, as wanted.
+    onButtonPress: function(zoomOrPanFunction, direction) {
+        this.setState({mouseDown: setInterval(zoomOrPanFunction(direction), 50)});
```
david-yz-liu: You need to pass in a function, not a function call (which doesn't return anything in this case). So probably use an anonymous function.

david-yz-liu:
@bell-kelly I'm not sure why a lot of the code is commented/deleted, but if things are working properly, I would like to merge this in.

bell-kelly:
Earlier today, the graph wasn't rendering and I was getting a weird console error. Commenting out that code had fixed things at the time. 

I just went an uncommented everything and added the code for the Modal (on line 490) and the InfoBox back in (on line 514) to try to replicate the error to show you, but everything seems to be working fine now. So I'm not sure what my issue was earlier, but now it doesn't seem to be a problem. Should I update the pull request with the code un-commented?

public/js/graph/react_graph.js
```diff
@@ -456,40 +486,47 @@ var Graph = React.createClass({
                     divId='zoom-in-button'
                     altId='zoom-in'
                     sourceImg="static/res/ico/in.png"
-                    mouseDown={() => this.incrementZoom(true)}/>
+                    mouseDown={() => this.onButtonPress(this.incrementZoom, true)}
+                    mouseUp={() => this.onButtonRelease()}/>
```
david-yz-liu: You should be able to simplify this: `() => f()` is equivalent to `f`.

public/js/graph/react_graph.js
```diff
@@ -438,6 +440,34 @@ var Graph = React.createClass({
         });
     },
 
+    onButtonPress: function(zoomOrPanFunction, direction) {
+        this.setState({mouseDown: setInterval(() => zoomOrPanFunction(direction), 100)});
```
david-yz-liu: Increase the interval time to 500. 100 is too short, and the `setState` is a bit of a hack right now (interval time doesn't correspond to re-rendering time).

david-yz-liu: Also, before this line, you should call `zoomOrPanFunction(direction)` once so that if a user clicks (but doesn't hold) the button, it'll still work.

bell-kelly: I'm about to update the pull request. I think I fixed the hack that you mentioned, but I just wanted to confirm that I understood the problem correctly (and fixed it correctly). 

Is it the case that the old code (in the snippet above) sets the state over and over at each interval? If so, my solution (included in my update) is to use a variable to get the return value from `setInterval` and then just pass the variable to set the state of `mouseDown`. 

If this isn't the correct approach, then my feeling is that maybe I need to find a different way to pass the return value from `setInterval` to `clearInterval` rather than using state to do this.

Also, now controlling the graph with the buttons is quite slow (by changing the interval to 500) and using the mouse wheel and arrow buttons are quite fast. Should I try to find a way to slow down the mouse wheel zoom and the arrow button panning?

david-yz-liu: Sort of. The problem is `setState` is asynchronous, so even if `setState` is called at each interval, that's not necessarily going to change the state right away, nor is it going to re-render the svg right away (that's why the zooming is a little choppy, but it's less noticeable when there's more of a delay). I'm not really sure what your solution is, since it seems like the existing code already puts the return value of `setInterval` as the `mouseDown` state.

However, I did some research and perhaps I was wrong about this being a "hack" at all. E.g., this is basically the approach used in https://facebook.github.io/react/docs/reusable-components.html#mixins. Maybe the only issue is when the period is too short. So this might not be worth worrying about.

As for the arrow/scroll events... don't worry about that yet (even though it is a little annoying). I think the best thing to do in that case is use an external library for help. You might be interested in checking out https://lodash.com/docs#throttle

public/js/graph/react_graph.js
```diff
@@ -132,12 +132,15 @@ var Graph = React.createClass({
...
-            verticalPanFactor: 0
+            verticalPanFactor: 0,
+            mouseDown: false
         };
     },
 
     componentDidMount: function () {
         this.getGraph();
+        document.body.addEventListener('keydown', this.onKeyDown);
+        document.body.addEventListener('wheel', this.onWheel);
```
arkon: Remember to remove these event listeners in `componentWillUnmount`.

bell-kelly: Good point. Thanks for noticing that!

public/js/graph/react_graph.js
bell-kelly: Thanks for the feedback and those resources. throttle seems interesting and could probably work here. 

I will continue to work on ways to improve the current implementation.

public/js/graph/react_graph.js
```diff
@@ -139,8 +139,13 @@ var Graph = React.createClass({
 
     componentDidMount: function () {
         this.getGraph();
-        document.body.addEventListener('keydown', this.onKeyDown);
-        document.body.addEventListener('wheel', this.onWheel);
+        document.getElementById('react-graph').addEventListener('keydown', this.onKeyDown);
+        document.getElementById('react-graph').addEventListener('wheel', this.onWheel);
+    },
+
+    componentDidUnmount: function () {
```
arkon: This should be `componentWillUnmount`. Obviously if it's already unmounted, there'd be no way for the component to really know that... :)

 Check out the [lifecycle docs](https://facebook.github.io/react/docs/component-specs.html) for more info.

bell-kelly: Thanks! That was a silly mistake seeing as componentDidUnmount doesn't even exist. Giving that document a read, I think I now better understand why exactly that is.

public/js/graph/react_graph.js
```diff
@@ -139,12 +151,13 @@ var Graph = React.createClass({
...
-        document.getElementById('react-graph').addEventListener('keydown', this.onKeyDown);
+        // can't detect keydown event when adding event listener to react-graph
+        document.body.addEventListener('keydown', this.onKeyDown);
         document.getElementById('react-graph').addEventListener('wheel', this.onWheel);
     },
 
-    componentDidUnmount: function () {
-        document.getElementById('react-graph').addEventListener('keydown', this.onKeyDown);
+    componentWillUnmount: function () {
+        document.body.addEventListener('keydown', this.onKeyDown);
```
arkon: These should be `removeEventListener`.

bell-kelly: The dangers of copy and paste...Thank you!

bell-kelly:
I'm trying something new. I ran into a lot of problems while trying to change the image of a button that was pushed. If you think CSS was the wrong way to go and that I should keep trying to change the image, I can discard these changes and continue to look into different ways to approach the problem while using images for the buttons.

bell-kelly:
I will clean up all the weird spacing issues in the morning. Already shut off my computer and making the changes on my iPad isn't working out super well.

public/js/graph/react_graph.js
```diff
@@ -450,46 +536,59 @@ var Graph = React.createClass({
             preserveAspectRatio: 'xMinYMin'
         };
 
+        var zoomInDisabled = (this.state.zoomFactor <= 0.5) ? true : false;
```
david-yz-liu: These expressions can all be simplified - `zoomInDisabled = this.state.zoomFactor <= 0.5`

public/js/graph/react_graph.js
```diff
@@ -450,46 +536,59 @@ var Graph = React.createClass({
...
+                    text="←"
+                    mouseDown={() => this.onButtonPress(this.panDirection, 'left', 10)}
+                    mouseUp={this.onButtonRelease}
+                    disabled={panLeftDisabled}/>
+                <Button
                     divId='reset-button'
-                    altId='reset'
-                    sourceImg="static/res/ico/reset.png"
-                    mouseDown={() => this.resetZoomAndPan()}/>
+                    text="Reset"
```
david-yz-liu: In general use single-quotes

david-yz-liu:
@bell-kelly great work. I made two minor comments; after you address these, I will merge this in.

bell-kelly:
Thanks David. I made those changes you suggested, as well as some changes that were brought up in the meeting. The graph buttons are now a darker grey when they are disabled and the reset button will now also be disabled when resetting will have no effect (ie. when the zoom and pan factors are at their initial state).

david-yz-liu:
Thanks, @bell-kelly!

PR #851, opened 2016-06-11
Use new Modal component in Check My POSt

david-yz-liu:
@ryanfan @christinem I did this one myself because I wanted it on CDF today.

Ryan, I noticed that there are several redundant requests for course info when generating the modals. In particular, it seems like your code is making one request to populate the outer `Modal` component, but then the inner `Description` component does it again. (And there are others on top of that.)

### DIFF
--- a/public/js/common/react_modal.js
+++ b/public/js/common/react_modal.js
@@ -51,11 +51,24 @@ var Description = React.createClass({
         };
     },
 
-    componentDidMount: function(){
-        //This loads the course json
+    componentDidMount: function() {
+        this.refresh();
+    },
+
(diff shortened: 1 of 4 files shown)


public/js/common/react_modal.js
```diff
@@ -51,11 +51,24 @@ var Description = React.createClass({
         };
     },
 
-    componentDidMount: function(){
-        //This loads the course json
+    componentDidMount: function() {
+        this.refresh();
+    },
+
+    componentWillUpdate: function (newProps, newState) {
```
david-yz-liu: @ryanfan @christinem we should probably be using life cycle hooks like this one more.

PR #853, opened 2016-06-22
Removing duplicates from course insertion

klmerryn:
Minor change to prevent duplicate Course entities from being inserted into the databse. Duplicates are assessed based on course code, now the primary key in Courses table (table.hs).


david-yz-liu:
@klmerryn beautiful.

PR #855, opened 2016-06-24
[Issue 828] Color Palette

hermish:
Changed the color palette to include a nice selection of pastel colors, all of which are light enough to allow superimposed text to be readable.


david-yz-liu:
@hermish looked great, thanks!

PR #857, opened 2016-07-06
[PARTIAL- to review] Refactor repeated queries to database-CourseQueries.hs

klmerryn:
In progress snippet with a possible implementation for removing repeated db queries, replacing them with one query + helper methods.


app/Database/InProgCourseQueries.hs
```diff
@@ -0,0 +1,70 @@
+--| UberGoal: go from T.Text -> [Entity Lecture] / [Entity Tutorial] without multiple DB queries.
+
+import Data.Tuple.Select as Tup -- selecting first, second and third elements of tuples (fall, spring, year)
+
+-- | Step 1: single query for lecture or tutorial. Produces list of all 
+-- | lectures/tutorials respectively, whcih are not separated by session
+
+-- | Queries db for all matching records with lecture or tutorial code of this course
+lectureQuery :: T.Text <- [Entity Lecture]
```
david-yz-liu: This type isn't right. `selectList` is a database action, which has type `MonadIO => ReaderT SqlBackend m a`, where `a` is the "returned" type. In this case, this would be `[Entity Lecture]`, for a better type of `ReaderT SqlBackend m a`.

The `T.Text` as the argument type of the function is correct. So the full type is something like `T.Text -> ReaderT ...`, plus the `MonadIO` constraint.

app/Database/InProgCourseQueries.hs
```diff
@@ -0,0 +1,70 @@
...
+
+-- | Step 2: find a way to split these into separate lists.
+
+-- | goal: refactor both below to something like
+-- | splitSessions :: Entity => [a] -> ([a], [a], [a]) 
+
+-- | Returns a 3-element tuple representing Fall, Spring, and Yearlong entity lists.
+-- | problem: Data.Text vs String
+splitSessionsL :: [Entity Lecture] -> ([Entity Lecture], [Entity Lecture], [Entity Lecture])
+splitSessionsL lecturesList = do
```
david-yz-liu: Only use `do` notation for monads - the only monads you care about right now are `IO` and `ReaderT SqlBackend`.

In this case, you can just use a simple `let` to bind `fallLec = getBySessionL lecturesList "F"`, etc.

app/Database/InProgCourseQueries.hs
```diff
@@ -0,0 +1,70 @@
...
+-- | However, this is inefficient because it requires 3 separate calls to the [Entity] list.
+-- | There is probably a way to go directly to a tuple of lists.
+
+-- | goal: refactor both below to something like
+-- | getBySession :: [EntityField a] -> [Entity a] -> [Entity a]
+
+-- | Returns the matching lecture entities if sessions correspond with session code.
+  -- bb: need Data.Text not string!
+getBySessionL :: String -> [Entity Lecture] -> [Entity Lecture]
+getBySessionL xs sessionStr = [ x | x <- xs, (LectureSession x) == sessionStr ]
```
david-yz-liu: I recommend using the `filter` function, as it was designed specifically for this purpose.

app/Database/InProgCourseQueries.hs
david-yz-liu: Also, you may want to look at how we're converting `Entity Lecture -> Lecture` (there's an `entityVal` helper function defined somewhere).

app/Database/InProgCourseQueries.hs
```diff
@@ -0,0 +1,70 @@
...
+getBySessionT :: String -> [Entity Tutorial] -> [Entity Tutorial]
+getBySessionT xs sessionStr = [ x | x <- xs, (TutorialSession x) == sessionStr ]
+
+
+-- | Queries the database for all information about @course@,
+-- constructs and returns a Course value.
+returnCourse :: T.Text -> IO Course
+returnCourse lowerStr = runSqlite databasePath $ do
+    let courseStr = T.toUpper lowerStr
+    sqlCourse :: [Entity Courses] <- selectList [CoursesCode ==. courseStr] []
```
david-yz-liu: Use `selectFirst` instead of `selectList` - the latter will return a `Maybe` instead of a list, which is probably better for your purposes.

app/Database/InProgCourseQueries.hs
```diff
@@ -0,0 +1,70 @@
...
+    -- Query once each for lec and tut, then build session of each 
+    allLectures <- splitSessionsL $ lectureQuery courseStr
```
david-yz-liu: It's not necessary to use a separate library here, you should be able to make use of pattern-matching to get extract the components of the tuple.

```haskell
(fall, winter, spring) <- splitSessionsL ...
```

app/Database/CourseQueries.hs
```diff
@@ -36,7 +37,51 @@ import Data.Int (Int64)
 import Database.DataType
 import Svg.Builder
 
--- ** Querying a single course
+-- | Queries db for all matching records with lecture or tutorial code of this course
+lectureQuery :: MonadIO m => T.Text -> ReaderT SqlBackend m a
```
david-yz-liu: @klmerryn Did you see my comment for you on #859? This type can be simplified.

david-yz-liu: Oh also the last part of the type, the `a`, should really be the type of value which results from the query. In this case it would be something like `[Entity Lecture]`.

app/WebParsing/ArtSciParser.hs
```diff
@@ -46,7 +46,10 @@ getCalendar str = do
         coursesSoup = lastH2 tags
         course = map (processCourseToData . filter isTagText) $ partitions isCourseTitle coursesSoup
     print $ "parsing " ++ str
+    --runSqlite databasePath $
```
david-yz-liu: Right, I looked into this even more, but still was not able to come up with a solution. I don't think we should worry about this for now, so you can revert the changes in this file.

app/Database/CourseQueries.hs
```diff
@@ -36,7 +37,54 @@ import Data.Int (Int64)
 import Database.DataType
 import Svg.Builder
 
--- ** Querying a single course
+---- | Queries db for all matching records with lecture or tutorial code of this course
+lectureQuery :: MonadIO m => T.Text -> ReaderT SqlBackend m [Entity Lecture]
```
david-yz-liu: @klmerryn I mentioned in #859 how you can simplify this type.

app/Database/CourseQueries.hs
```diff
@@ -36,7 +37,54 @@ import Data.Int (Int64)
...
+        tutorialsList :: [Entity Tutorial] <- tutorialQuery courseStr
+        let (fallLec, springLec, yearLec) = splitSessionsL lecturesList
+            (fallTut, springTut, yearTut) = splitSessionsT tutorialsList
+            fallSession = buildSession fallLec fallTut
+            springSession = buildSession springLec springTut
+            yearSession = buildSession yearLec yearTut
+        return (buildCourse fallSession springSession yearSession (entityVal course))
+
+        
+--buildAllSessions :: T.Text -> ([Maybe Tables.Session], [Maybe Tables.Session], [Maybe Tables.Session])
```
david-yz-liu: You can remove this part now.

klmerryn:
Hey David -- sorry, I messed something up with my second-last commit, I was checking out an older version to compare my changes and shouldn't have pushed anything. I think I have fixed it. I also added a text file to show you where I'm at... hope that all makes sense. 
The SqlPersistM thing didn't totally work out, even when importing the module. will show you tomorrow.

app/WebParsing/ArtSciParser.hs
david-yz-liu: @klmerryn please make sure to revert these changes.

app/Database/16-07-27-weekly-summary.txt
```diff
@@ -0,0 +1,22 @@
+16-07-27 Weekly summary (brief).
```
david-yz-liu: @klmerryn you shouldn't use a pull request to share status updates.

It wouldn't be so bad, but in particular, the contents of this document have nothing to do with this pull request. This pull request is in the process of being wrapped up, and which I hope to merge in soon. But I can't do so with this file in here.

klmerryn: sorry. I thought all of this was supposed to be part of the same refactoring, and you wanted it all to be one issue. I guess it wasn't a very smart place for an update .. I will take it out.

PR #858, opened 2016-07-07
react-draw: use react to draw nodes [WIP]

nzliu:
Use react to draw nodes on a mouse on click event

### DIFF
--- a/public/js/draw.js
+++ b/public/js/draw.js
@@ -12,7 +12,6 @@ requirejs([
         var graphComponent = reactGraph.renderReactGraph('react-graph', true, true);
 
         svgDoc = $('#react-graph svg')[0];
-        // svgDoc.addEventListener('mousedown', draw.makeNodePath, false);
         svgDoc.addEventListener('mousemove', draw.moveNodeElbow, false);
         svgDoc.addEventListener('mouseup', draw.unclickAll, false);
     })
--- a/public/js/draw.js
+++ b/public/js/draw.js
@@ -12,7 +12,7 @@ requirejs([
         var graphComponent = reactGraph.renderReactGraph('react-graph', true, true);
 
         svgDoc = $('#react-graph svg')[0];
-        svgDoc.addEventListener('mousedown', draw.makeNodePath, false);
+        // svgDoc.addEventListener('mousedown', draw.makeNodePath, false);
         svgDoc.addEventListener('mousemove', draw.moveNodeElbow, false);
         svgDoc.addEventListener('mouseup', draw.unclickAll, false);
     })
--- a/public/js/graph/react_graph.js
+++ b/public/js/graph/react_graph.js
@@ -217,6 +217,8 @@ var Graph = React.createClass({
             }
         }
 
+        graphName = graphName.replace('-', ' ');
+
         $.ajax({
             dataType: 'json',
             url: 'get-json-data',
--- a/public/js/graph/react_graph.js
+++ b/public/js/graph/react_graph.js
@@ -112,6 +112,8 @@ function Button(props) {
         <button id={props.divId} className='graph-control-button'
         onMouseDown={props.mouseDown}
         onMouseUp={props.mouseUp}
+        onMouseEnter={props.onMouseEnter}
+        onMouseLeave={props.onMouseLeave}
         disabled={props.disabled}>{props.text}</button>
     );
 }
@@ -119,15 +121,25 @@ function Button(props) {
 
 function renderReactGraph(graph_container_id, start_blank, edit) {
     'use strict';
+
+    var initialGraphMode;
+
     if (start_blank === undefined) {
         start_blank = false;
     }
+    // if edit is NOT undefined, then the user is on the draw page
     if (edit === undefined) {
         edit = false;
+        initialGraphMode = 'graph';
+    } else {
+        initialGraphMode = 'draw-node';
     }
 
     return ReactDOM.render(
-        <Graph start_blank={start_blank} edit={edit} />,
+        <Graph 
+            start_blank={start_blank}
+            edit={edit}
+            initialGraphMode={initialGraphMode}/>,
         document.getElementById(graph_container_id)
     );
 }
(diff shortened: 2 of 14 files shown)


public/js/graph/react_graph.js
```diff
@@ -554,6 +554,11 @@ var Graph = React.createClass({
         }
     },
 
+    drawNode: function(nodeJSON) {
+        var newNodesJSON = $.extend([], this.state.nodesJSON).push(nodeJSON);
+        this.setState({nodesJSON: newNodesJSON});
+    },
+
```
nzliu: not sure if i will need to manually create a nodeJSON object here

david-yz-liu: Yeah, you'll need to extract data from the mouse event object. Take a look at how nodes are currently added.

public/js/graph/react_graph.js
```diff
@@ -1194,8 +1199,7 @@ var Bool = React.createClass({
 
 var EdgeGroup = React.createClass({
     // EdgeGroup's state is used to keep track of the edgeIDs of
-    // edges that are missing. Void is just a placeholder state so
-    // we can declare an initial state; it does nothing.
+    // edges that are missing.
```
david-yz-liu: This particular change should be done in a separate pull request to `master`, not here. It doesn't relate to the Draw tool.

public/js/graph/react_graph.js
```diff
@@ -554,6 +556,96 @@ var Graph = React.createClass({
         }
     },
 
+    /**
+     * Calculates the position of the click in relation to the page.
+     * @param {object} e The click event.
+     * @param {HTMLElement} elem The target of the click event.
+     * @return {object} The position of the click.
+     */
+    getClickPosition: function(e, elem) {
```
david-yz-liu: Because these functions don't need `this`, you can make them helpers which are outside of the class. (Put them at the bottom of the file for now.)

public/js/graph/react_graph.js
```diff
@@ -554,6 +556,96 @@ var Graph = React.createClass({
...
+            xPosition += (elem.offsetLeft || 0) - elem.scrollLeft + elem.clientLeft;
+            yPosition += (elem.offsetTop || 0) - elem.scrollTop + elem.clientTop;
+            elem = elem.parentElement; // offsetParent undefined in mozilla
+        }
+        return { x: xPosition, y: yPosition };
+    },
+
+    makeNode: function(xPos, yPos) {
+        var textsList = [];
+        // create text jsons, move to helper function?
```
david-yz-liu: This isn't a bad idea, but because it'll always start as `[]`, I wouldn't worry about it right now.

public/js/graph/react_graph.js
```diff
@@ -554,6 +556,96 @@ var Graph = React.createClass({
...
+        this.setState({nodesJSON: newNodesJSON, drawNodeID: newDrawNodeID});
+    },
+
+    /**
+    * In node-mode creates a new node at the position of the click event on the SVG canvas.
+    * In path-mode creates an elbow at the position of the click event on the SVG canvas,
+      if the startNode is defined.
+    * @param {object} e The mousedown event.
+    */
+    makeNodePath: function(e) {
```
david-yz-liu: Why don't we change this name to `makeGraphObject` or something?

public/js/graph/react_graph.js
```diff
@@ -554,6 +556,96 @@ var Graph = React.createClass({
...
+        var position = getClickPosition(e, e.currentTarget);
```
david-yz-liu: Seems redundant to pass in both `e` and `e.currentTarget`.

david-yz-liu:
@nzliu okay, so it appears that you did not actually undo the merge, or at least not fully. Your diff shows some changes from the master branch.

I've updated the `react-draw` branch with the latest version of master. Once you do a `git pull upstream react-draw`, the changes will be absorbed, so you shouldn't have to worry about that.

In the future, remember that you're only working with `upstream react-draw`, not `upstream master` for this project.

public/js/draw.js
```diff
@@ -12,7 +12,7 @@ requirejs([
         var graphComponent = reactGraph.renderReactGraph('react-graph', true, true);
 
         svgDoc = $('#react-graph svg')[0];
-        svgDoc.addEventListener('mousedown', draw.makeNodePath, false);
+        // svgDoc.addEventListener('mousedown', draw.makeNodePath, false);
```
david-yz-liu: You can delete this line.

public/js/graph/react_graph.js
```diff
@@ -150,10 +150,21 @@ var Graph = React.createClass({
             zoomFactor: 1,
             horizontalPanFactor: 0,
             verticalPanFactor: 0,
-            mouseDown: false
+            mouseDown: false,
+            graphMode: 'graph',
```
david-yz-liu: You can use a prop to set the initial value rather than doing what you did below. See https://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html for details.

public/js/graph/react_graph.js
```diff
@@ -150,10 +150,21 @@ var Graph = React.createClass({
...
+            drawNodeID: 0
         };
     },
 
+    componentWillMount: function () {
+        if (location.href.endsWith('draw')) {
+            this.setState({graphMode: 'draw-node'});
+            document.body.addEventListener('mousedown', this.drawGraphObject, false);
```
david-yz-liu: You don't want to set the callback here; instead, it should be directly tied to the `svg` element that's rendered for this component.

public/js/graph/react_graph.js
```diff
@@ -556,6 +567,64 @@ var Graph = React.createClass({
...
+        // text position uses node position for now
+        var textJSON = {
+            'align': 'begin',
+            'fill': '',
+            'graph': 0,
+            'pos': [xPos, yPos+20],
+            'rId': 'text' + this.state.drawNodeID,
+            'text': 'la'
+        }
+        var texts = [textJSON];
```
david-yz-liu: You don't need this temporary variable.

public/js/graph/react_graph.js
```diff
@@ -556,6 +567,64 @@ var Graph = React.createClass({
...
+            'pos': [xPos, yPos],
+            'stroke': '',
+            'text': texts,
+            'tolerance': 9,
+            'type_': 'Node'   
+        };
+        var newNodesJSON = $.extend([], this.state.nodesJSON);
+        newNodesJSON.push(nodeJSON);
+        console.log(newNodesJSON);
+        var newDrawNodeID = this.state.drawNodeID + 1;
```
david-yz-liu: You don't need a temporary variable for this one

public/js/graph/react_graph.js
```diff
@@ -556,6 +567,64 @@ var Graph = React.createClass({
...
+
+    /**
+    * In draw-node creates a new node at the position of the click event on the SVG canvas.
+    * In path-mode creates an elbow at the position of the click event on the SVG canvas,
+      if the startNode is defined.
+    * @param {object} e The mousedown event.
+    */
+    drawGraphObject: function(e) {
+        var position = getClickPosition(e, e.currentTarget);
+        if (this.state.graphMode === 'draw-node') {
```
david-yz-liu: Make this a top-level check (i.e., swap it with the previous line).

PR #860, opened 2016-07-07
Adding a response for the path /export-pdf to display a PDF

bell-kelly:
… pdf on a blank page.


app/Response/Export.hs
```diff
@@ -8,14 +8,7 @@ import qualified Text.Blaze.Html5 as H
...
-    ok $ toResponse $
-        displayPDF
-
-
-{- Insert a large loading icon into the page -}
-displayPDF :: H.Html
-displayPDF =
-    H.div ! A.id "exportDiv" $
-        H.img ! A.id "export-pdf" ! A.src "../CSgraph.pdf"
+    serveDirectory DisableBrowsing ["CSGraph.pdf"] "."
```
david-yz-liu: Better, although it seems like there should be a `serveFile` function to use instead

bell-kelly: OK thanks! Changed it!

I'm not really sure what I need to be focusing on as a next step now.

david-yz-liu:
@bell-kelly Okay cool, the next step is to write a Hello World Latex `.tex` file, and then spawn a subprocess from the server to run `pdflatex` on it to generate a PDF, and then render that PDF instead.

Note that there is already an example of running a subprocess in the Haskell code which runs ImageMagick, so if you do some digging you can find that.

bell-kelly:
Ok I'll give that shot! Thank you!

app/Response/Export.hs
```diff
@@ -0,0 +1,54 @@
...
+								CreatePipe
+								CreatePipe
+								False
+								False
+								False
+
+-- | Opens a new process to create a .pdf file from a .tex file and waits for that process to terminate.
+createPDF :: String -> String -> IO ()
+createPDF texName pdfName =
+	do
```
david-yz-liu: In general, put the `do` on the same line as the function header, so that everything after the `do` is indented only once.

Also, you seem to be using tabs instead of spaces.

app/Server.hs
```diff
@@ -62,6 +62,7 @@ runServer = do
               dir "get-json-data" $ look "graphName" >>= \graphName -> liftIO $ getGraphJSON graphName,
               dir "loading" $ look "size" >>= loadingResponse,
               dir "save-json" $ look "jsonData" >>= \jsonStr -> look "nameData" >>= \nameStr -> liftIO $ saveGraphJSON jsonStr nameStr,
+              dir "export-pdf" $ exportResponse "HelloWorld",
```
david-yz-liu: You can start by making this use the same image from the `image` route.

In order to do the `timetable-image` route, you'll need to add `look` calls to access the "courses" and "session" query parameters.

bell-kelly: So, this week I've been quite confused about how this will all fit together. I've been trying to work it out on my own, but I might just be making this more complicated in my head than it needs to be.

In order to make "this use the same `image` from the image route", should the `export-pdf` route first call the `graphImageResponse` function and then use the resulting image data sent as a response to the `exportResponse`. Following that, would I have to work on finding a way to insert that data into a TEX file which will be converted to a pdf by `exportResponse`? I did some searching and it might be possible to use the filecontents package in Latex to include a base64 encoded PDF in a TEX file.

Or does this mean, that `exportResponse` should retrieve the graph id on the current page in the same way that `graphImageResponse` does, and then use that info to generate a PDF from the SVG which will then somehow be instead into the TEX file?

I actually have a feeling I'm just completely missing the point and the correct approach is something else entirely, but I haven't been able figure it out.

david-yz-liu: 1. You should extract the necessary code to generate an image into a helper function which is used by both `image` and `export-pdf` routes separately.

2. Getting the filename into the tex document is harder. A first approach would be to hardcode this name.
A second, better approach would be to capture the randomly-generated filename, and then pass it as a commandline argument to the tex file using the technique from here: http://tex.stackexchange.com/questions/243854/possible-to-compile-a-tex-file-with-command-line-arguments-passed-to-the-tex-fil.

HelloWorld.tex
```diff
@@ -0,0 +1,4 @@
+\documentclass{article}
```
david-yz-liu: @bell-kelly It seems you didn't update the tex in this file.

app/Response/Export.hs
```diff
@@ -0,0 +1,54 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Response.Export
+    (exportResponse) where
+
+import           Text.Blaze ((!))
```
david-yz-liu: Double-check the imports here to remove unnecessary ones.

bell-kelly:
I'm sorry it's been so long since I've updated this pull request. I've been working through countless type errors as I try to learn more about Haskell. 

I believe I've addressed all of your feedback in this commit (using do on the same line as the header, removing unused imports and I added my new .tex file). 

This is still a work in progress as I try to learn better ways of completing some of the functions and extracting helper functions. I know the code is a bit clumsy.

app/ImageConversion.hs
```diff
@@ -1,8 +1,81 @@
...
+    let (rand, _) = next gen
+        svgFilename = show rand ++ ".svg"
+        imageFilename = show rand ++ ".png"
+    renderTable svgFilename courses session
+    createImageFile svgFilename imageFilename
+    return (svgFilename, imageFilename)
+
+-- | Opens a new process to create a .pdf file from a .tex file and waits for that process to terminate.
+createPDF :: String -> String -> IO ()
+createPDF texName imgName = do
```
david-yz-liu: The second parameter name should be changed

app/ImageConversion.hs
```diff
@@ -1,8 +1,81 @@
...
+-- | Create a process to use the pdflatex program to create a .pdf file from a .tex file using a supplied image
+convertTexToPDF :: String -> String -> IO
+                                   (Maybe Handle,
+                                    Maybe Handle,
+                                    Maybe Handle,
+                                    ProcessHandle)
+
+convertTexToPDF texName imgName = createProcess $ CreateProcess
+                      (ShellCommand $ "pdflatex " ++
+                              "--jobname=" ++
```
david-yz-liu: What is the purpose of this argument?

app/ImageConversion.hs
```diff
@@ -1,8 +1,81 @@
...
+                      Nothing
+                      Nothing
+                      CreatePipe
+                      CreatePipe
+                      CreatePipe
+                      False
+                      False
+                      False
+
+-- Note: hGetContents can be used to read Handles. Useful when trying to read from
```
david-yz-liu: This comment should be moved.

app/ImageConversion.hs
```diff
@@ -1,8 +1,81 @@
...
 
 import System.Process
 import GHC.IO.Handle.Types
+import TimetableImageCreator (renderTable)
+import qualified Data.Map as M
+import System.Random
+import Svg.Generator
+import Database.Tables (GraphId)
+
+-- | Creates an image, and returns the name of the svg used to create the
```
david-yz-liu: You should move these to a separate file, since they are helper functions that are shared between multiple actions.

app/Response/Export.hs
```diff
@@ -0,0 +1,42 @@
...
+import ImageConversion
+import Database.Persist.Sql (toSqlKey)
+import Data.Int(Int64)
+
+-- | Serves a pdf file of a graph image
+exportGraphResponse :: ServerPart Response
+exportGraphResponse = do
+    req <- askRq
+    let cookies = M.fromList $ rqCookies req
+        gId = maybe "1" cookieValue (M.lookup "active-graph" cookies)
```
david-yz-liu: This is out of date. Make sure you have the latest version of master, and look at the `graphImageResponse` function in `Image.hs`.

app/Response/Export.hs
```diff
@@ -0,0 +1,42 @@
...
+        graphKey = read gId :: Int64
+    (svgFilename, imageFilename) <- liftIO $ getGraphImage (toSqlKey graphKey) (M.map cookieValue cookies)
+    pdfName <- liftIO $ returnPDF svgFilename imageFilename
+    serveFile (asContentType "application/pdf") pdfName
+
+-- | Serves a pdf file of a timetable requested by user
+-- Note: Not completely working yet! Sending ByteString I think
+exportTimetableResponse :: String -> String -> ServerPart Response
```
david-yz-liu: You don't need a separate response here. In fact, you will ultimately generate one PDF file that has both sets of images.

courseography.sublime-project
```diff
@@ -0,0 +1,8 @@
+{
```
david-yz-liu: You can remove this file.

app/Server.hs
```diff
@@ -81,4 +83,4 @@ runServer = do
         cwd <- getCurrentDirectory
         --let parentDir = Path.parent $ Path.decodeString cwd
         --return $ Path.encodeString $ Path.append parentDir $ fromString "public/"
-        return $ Path.encodeString $ Path.append (Path.decodeString cwd) $ fromString "public/"
+        return $ Path.encodeString $ Path.append (Path.decodeString cwd) $ fromString "public/"
```
david-yz-liu: Please revert this change.

david-yz-liu:
@bell-kelly this is good progress. I've left another round of comments - and in particular you can remove the timetable image stuff (but you can leave it until the next meeting if you want to demo it). After this latest round, I'll probably be able to merge this in.

app/ImageConversion.hs
bell-kelly: The `pdflatex` arguments I had previously been using resulted in a .pdf with the name "export.pdf" (when using the .tex file "export.tex"). The `--jobname` argument forces the resulting .pdf to have the same name as the image it contains (a random number). I learned about this by searching how to influence the name of the .pdf resulting from a call to `pdflatex` and found [this](http://tex.stackexchange.com/questions/19182/how-to-influence-the-name-of-the-pdf-file-created-with-pdflatex-from-within-the). I thought having a random name as opposed to a constant name might be the desired result, but maybe I was wrong?

bell-kelly:
Thanks for the feedback, David. I made most of those changes. I will leave the timetable image stuff for now will work on removing it in the coming week.

david-yz-liu:
@bell-kelly I see, I didn't realize that `jobname` was how you specify an output file name. That's fine.

david-yz-liu:
@bell-kelly Okay I merged this in. A couple of comments from my testing:

1. The images render fine for me. Delete your database and recreate the graphs, and see if the problem still comes up.
2. Don't use the `incgraph` library. Just use `graphicx`; see https://en.wikibooks.org/wiki/LaTeX/Importing_Graphics.
3. You have some whitespace issues that you really need to fix. I don't want to have to keep bringing this up. It seems like you need to change your text editor's configuration.
4. You need to actually make use of the helper functions you've created.

bell-kelly:
Thanks for the feedback, David. 
1. I reinstalled `imagemagick`. I also updated my fork, deleted the database, and ran the `build`, `graph` and `database` commands to see if this would fix the problem. However, the images still render as they did at Thursday's meeting. 
2. I originally used `graphicx`, but when I found `incgraph` I thought it was pretty cool. I can easily go back to using `graphicx` though. 
3. I'm sorry about the whitespace issues. I've spent some time going through the settings of my text editor and will monitor this closely to make sure the problem was resolved. 
4. I hope to have this done before the next meeting.
Thanks again!

Issue #710, opened 2015-09-12
Create a real export modal.


david-yz-liu:
Should have different tabs for the graph and grid, and allow both the timetable image and iCal formats.

Best to integrate the Facebook posting directly into the modal, rather than have a separate button in the main nav.

Ian-Stewart-Binks:
I think that this is (partially) blocked by #709, as we need some way to get the timetable data from the context of the graph.

david-yz-liu:
@hermish will work on taking the existing modal and converting it into a React component, before making any further UI changes.

PR #862, opened 2016-07-15
Fix export

david-yz-liu:
This fixes the export modals on the Graph and Grid pages.

This is a hack, and doesn't fix the Facebook modals, which I am going to disable on CDF. I'll consider fixing those as part of #710.


david-yz-liu:
@bell-kelly After I merge this in, you'll be able to use the images generated by the server in your PDFs.

PR #864, opened 2016-08-03
[In progress] Foreign key constraints

klmerryn:
Introducing foreign key constraints to database by first using Breadth and Distribution table keys as foreign keys in Courses table. (In progress).

### DIFF
--- a/app/Database/CourseInsertion.hs
+++ b/app/Database/CourseInsertion.hs
@@ -48,7 +48,6 @@ saveGraphJSON jsonStr nameStr = do
 getDistributionKey :: Maybe T.Text -> SqlPersistM (Maybe (Key Distribution))
 getDistributionKey Nothing = return Nothing 
 getDistributionKey (Just description) = do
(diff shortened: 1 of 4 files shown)


klmerryn:
Still working on resolving two type errors, one re:Data.Text vs Strings and one re: Maybe [Key Breadth/Distribution]

david-yz-liu:
@klmerryn 

1. Use Data.Text.pack/unpack to convert between these two types.
2.  Use pattern matching to define `getDistributionKey` (and the breadth one too, probably). Remember that you can define a function by pattern matching on constructors used for its arguments:

```
getDistributionKey Nothing = Nothing
getDistributionKey (Just x) = something with x...
```

3. `getDistributionKey` involves a database action, and so must itself be a database action. You'll need to specify this in the type! (`SqlPersistM ...`)

david-yz-liu:
@klmerryn I cleaned things up a little bit myself and merged this in, but there's still some stuff to do (e.g., with the name shadowing warnings).

PR #865, opened 2016-08-03
Export graph and timetable

bell-kelly:
Working on including both the active graph and timetable query results into a pdf image.


bell-kelly:
Sorry it has been a while since I was able to contribute, but I finally had time over the past couple days to work through some of the HaTex documentation. I modified some of the code so that now LaTeX files are created dynamically and so that they will be able to support the addition of multiple graphs and timetables. However, the export response still only generates one graph and one timetable image based on the query parameters.

PR #877, opened 2016-09-18
Issue 844 Fix

PkHutch:
Fixes the German graph not being parsed.
Changes the name of some files that were incorrectly named.
Changes the order in which the graphs are parsed for readability.

Note: The German graph does appear to have some visual bugs, opening issue about it.


david-yz-liu:
@PkHutch based on what we discussed, please make a new pull request which just adds the german graph.

PR #879, opened 2016-09-18
Issue 870 Partially Fixed

lanaelsanyoura:
The scroll bar resets to the top of the info box after the user chooses to scroll.

### DIFF
--- a/public/js/common/react_modal.js
+++ b/public/js/common/react_modal.js
@@ -20,7 +20,7 @@ export var Modal = React.createClass({
     render: function () {
         if (this.state.courseId) {
             return (
-                <div className='modal fade' >
+                <div className='modal fade'>
                     <div className='modal-dialog'>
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@lanaelsanyoura this seems plausible, though I haven't tried it out yet. What are you observing on your end when you open and close the modal?

lanaelsanyoura:
When the scroll bar reaches the bottom of the modal, it resets to the top. If the scroll bar is in the middle of the screen, whenever the user chooses to scroll, it will reset to the top and he would have to scroll down from there. Also, when a new info box is opened, the scroll bar is in the saved position. However, as soon as the user chooses to scroll, it goes to the top. I've been trying to use "scrollIntoView" in another function so that it resets only when the modal opens but I've yet to succeed .I've tried putting the line inside the "if" statement of the 'willUpdate' function but it doesn't work.

PR #890, opened 2016-09-22
CourseQueries warning fix and add WebParsing.ParsecCombinators to cabal

AilsaF:


app/Database/CourseQueries.hs
```diff
@@ -68,17 +67,17 @@ returnCourse lowerStr = runSqlite databasePath $ do
       Just course -> do
         lecturesList :: [Entity Lecture] <- lectureQuery courseStr
         tutorialsList :: [Entity Tutorial] <- tutorialQuery courseStr
-        let (fallSession, springSession, yearSession) = buildAllSessions lecturesList tutorialsList
-        buildCourse fallSession springSession yearSession (entityVal course)
+        let (fallsession, springsession, yearsession) = buildAllSessions lecturesList tutorialsList
```
david-yz-liu: @AilsaF we want to always use camelCase for variable names. Why don't we just name these "fall" "winter" and "year". Similar comment below.

app/Database/CourseQueries.hs
```diff
@@ -186,31 +185,31 @@ getGraphJSON graphName =
...
                                              (map keyAsInt sqlPaths)
-                    (regions, edges) = partition pathIsRegion paths
+                    (regions, _)   = partition pathIsRegion graphpaths
                     regionTexts    = filter (not .
                                              intersectsWithShape (rects ++ ellipses))
-                                            texts
+                                            graphtexts
 
-                    result = createJSONResponse $
+                    responseresult = createJSONResponse $
```
david-yz-liu: You can just call this `response`

PR #893, opened 2016-10-10
Export timetable

AilsaF:
Finish using cookie to get image and save images to local!!
However, when I used "returnPDF" (line 39 in Image.hs, detail code in Export.hs), I couldn't get a pdf. I am curious there may be some problem with returnPDF function, cuz I failed to export graph image as well. 
And also curious why I am able to export graph pdf in courseography.cdf.utoronto.edu while NOT able in localhost.


AilsaF:
Oh and I'll work on solving schedule conflict situation later.

david-yz-liu:
For the exporting problem, you might not have the additional software installed and available on your computer's PATH. Exporting requires both imagemagick and LaTeX (specifically, the executable pdflatex).

app/Database/CourseQueries.hs
```diff
@@ -271,3 +273,22 @@ queryGraphs =
         do graphs :: [Entity Graph] <- selectList [] []
            return $ createJSONResponse graphs
 
+-- ========================================================
+
+getLectureTime :: [String] -> IO (Maybe [Time])
+getLectureTime [code, section, session] = runSqlite databasePath $ do
```
david-yz-liu: Make these three strings as separate arguments; lists should only be used if the input can truly be an arbitrary length.

Also, don't have a separate `runSqlite` for each of these, and `getTutorialTime`.  Instead, you want to just define the database action (`SqlPersistM`), and then surround the whole thing in a `runSqlite` where you use it in `getActiveTimetable`.

david-yz-liu: Ah, I see what you're doing below. In this case, please have the type of the parameter be `(String, String, String)`.

app/Export/GetImages.hs
```diff
@@ -1,25 +1,98 @@
...
+    renderTableHelper springsvgFilename (zipWith (:) times spring_schedule') "Spring"
+    createImageFile fallsvgFilename fallimageFilename
+    createImageFile springsvgFilename springimageFilename
+    return (fallsvgFilename, fallimageFilename, springsvgFilename, springimageFilename)
+    where isFall (c, t) = last c == "F"
+    
+
+
+-- "CSC148H1-L5101-S_CSC148H1-T0501-S_STA355H1-L0101-F_CSC108H1-L0102-F"
+-- ([["CSC148H1","L5101","S"],["STA355H1","L0101","F"],["CSC108H1","L0102","F"]],  [["CSC148H1","T0501","S"]])
```
david-yz-liu: I recommend replacing "L" with "LEC" and "T" with "TUT" (and "P" with "PRA") here, to clean up the database lookup after.

app/Export/GetImages.hs
```diff
@@ -1,25 +1,98 @@
...
+parseCourseCookie :: String -> ([[String]], [[String]])
+parseCourseCookie s = let lecAndTut = map (splitOn "-") $ splitOn "_" s
+                          lectures = filter isLec lecAndTut
```
david-yz-liu: This can be simplified using the `partition` function.

app/Database/CourseQueries.hs
david-yz-liu: Also keep the return type as `[Time]` by returning an empty list if you get `Nothing`.

AilsaF: Do you mean using (String, String, String) to filter out cases not having 3 Strings (i.e. parsing wrongly)?

AilsaF: So I don't need to handle wrong cases, just let complier check and give error msg in getActiveTimetable?

david-yz-liu: Sure. You could just ignore this part of the string from the cookie.

AilsaF:
There is no style signature for getLetureTime and getTutorialTime for weird error. 
Couldn't match expected type ‘IO (Maybe (Entity Lecture))’
                with actual type ‘Control.Monad.Trans.Reader.ReaderT
                                    SqlBackend m0 (Maybe (Entity Lecture))’ 
Will keep doing some research on it. Please ignore it for the moment THX.

AilsaF:
Already had a router for generating PDF and put 3 images in PDF.

**I am leaving this note because I created CourseInfo data structure to wrap { code, session, section, time_list}. I figure this can make code more readable, but I am not sure. Thus, I also left original code there. Can you help me to take a look and see which version is better if you have time? Thanks a lot for taking time from busy midterm and assignment period :)**

There are also some minor problems. I will fix them and make another pull request then.

app/Database/CourseQueries.hs
```diff
@@ -269,3 +274,53 @@ queryGraphs =
...
+--                          HasPersistBackend env SqlBackend) =>
+--                         CourseInfo -> m CourseInfo
+getLectureTime courseInfo = do
+    maybeEntityLectures <- selectFirst [LectureCode ==. (T.pack . code $ courseInfo),
+                                        LectureSection ==. (T.pack . section $ courseInfo),
+                                        LectureSession ==. (T.pack . session $ courseInfo)]
+                                       []
+    -- liftIO $ print maybeEntityLectures
+    case maybeEntityLectures of
+        Nothing -> return CourseInfo {code = code courseInfo, section = section courseInfo, session = session courseInfo, time = []}
```
david-yz-liu: Use a `let time = ...` to first extract the time, and then you can remove the code duplication in the `return`

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
 import Happstack.Server (Request, rqCookies, cookieValue)
 import Data.List.Utils (replace)
+import Data.List.Split (splitOn)
+import Database.CourseQueries (getLectureTime, getTutorialTime)
+import Database.Tables as Tables
+import Control.Monad.IO.Class  (liftIO)
+import Data.List (partition)
+import Database.Persist.Sqlite (runSqlite, runMigration)
+import Config (databasePath)
+import Data.Fixed
```
david-yz-liu: Please specify what is being imported from here.

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
+-- [["8:00","","","","",""],["9:00","","","","",""],["10:00","CSC108 (L)","","CSC108 (L)","","CSC108 (L)"],["11:00","","","","",""],["12:00","","","","",""],["1:00","","","","",""],["2:00","STA355 (L)","","STA355 (L)","",""],["3:00","STA355 (L)","","","",""],["4:00","","","","",""],["5:00","","","","",""],["6:00","","","","",""],["7:00","","","","",""],["8:00","","","","",""]]
+getActiveTimetable :: Request -> String -> IO (String, String)
+getActiveTimetable req session = do
+    -- get cookie value of "selected-lectures" from browser
+    let cookies = M.fromList $ rqCookies req  --  Map String Cookie
+        coursecookie = maybe "" cookieValue $ M.lookup "selected-lectures" cookies
+        (lecture, tutorial) = parseCourseCookie coursecookie session  -- ([CourseInfo], [CourseInfo])
+    (lecTimes, tutTimes) <- getTimes (lecture, tutorial)
+    let schedule' = getScheduleByTime lecTimes tutTimes
+    (svgFilename, imageFilename) <- generateTimetableImg schedule' session
```
david-yz-liu: The last two lines here are redundant. You can simply put `generateTimetableImg schedule' session` as the last line.

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
+-- [["8:00","","","","",""],["9:00","","","","",""],["10:00","CSC108 (L)","","CSC108 (L)","","CSC108 (L)"],["11:00","","","","",""],["12:00","","","","",""],["1:00","","","","",""],["2:00","STA355 (L)","","STA355 (L)","",""],["3:00","STA355 (L)","","","",""],["4:00","","","","",""],["5:00","","","","",""],["6:00","","","","",""],["7:00","","","","",""],["8:00","","","","",""]]
+getActiveTimetable :: Request -> String -> IO (String, String)
+getActiveTimetable req session = do
+    -- get cookie value of "selected-lectures" from browser
+    let cookies = M.fromList $ rqCookies req  --  Map String Cookie
```
david-yz-liu: This type comment would be more appropriately written as `let cookies :: Map String Cookie = ...`

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
+-- parseCourseCookie s session = let lecAndTut = map (splitOn "-") $ splitOn "_" s
+--                                   (lecture, tutorial) = partition isLec lecAndTut
+--                                   [lectureOfsession, tutorialOfsession] = map (filter (\x -> or ([(x !! 2 !! 0) == (head session), (x !! 2 !! 0) == 'Y']))) [lecture, tutorial]
+--                                   lecture' = map (convertLec . list2tuple) lectureOfsession
+--                                   tutorial' = map (convertTut . list2tuple) tutorialOfsession
+--                               in (lecture', tutorial')
+--                               where isLec x = (x !! 1 !! 0) == 'L'
+
+parseCourseCookie :: String -> String -> ([CourseInfo], [CourseInfo])
+parseCourseCookie s session = let lecAndTut = map (splitOn "-") $ splitOn "_" s
```
david-yz-liu: More standard is to start the `let` on a new line (reduces indentation)

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
+convertLec (code, section, session) = CourseInfo {code = code, section = take 1 section ++ "EC-" ++ drop 1 section, session = session, time = [] }
+
+-- ("CSC148H1","T0501","S") -> CourseInfo {code = "CSC148H1", section = "TUT-0501", session = "S", time = []} convert format to match database
+convertTut :: (String, String, String) -> CourseInfo
+convertTut (code, section, session) = CourseInfo {code = code, section = take 1 section ++ "UT-" ++ drop 1 section, session = session, time = [] }
+
+
+-- | Get data from database for selected courses
+getTimes :: ([CourseInfo], [CourseInfo]) -> IO ([CourseInfo], [CourseInfo])
+getTimes (lecture, tutorial) = runSqlite databasePath $ do
```
david-yz-liu: Use plural forms for the names here.

app/Export/GetImages.hs
```diff
@@ -1,25 +1,161 @@
...
+    lecTimes <- mapM getLectureTime lecture  -- [(String, String, String)] -> IO ([[Time]])
+    tutTimes <- mapM getTutorialTime tutorial
+    return (lecTimes, tutTimes)
+
+
+-- | Generate scheduel in special format based on Times
+getScheduleByTime :: [CourseInfo] -> [CourseInfo] -> [[String]]
+getScheduleByTime lecTimes tutTimes = let allTimes = lecTimes ++ tutTimes
+                                          schedule = replicate 13 $ replicate 5 ""
+                                          schedule' = foldl (\acc x -> addCourseToSchedule x acc) schedule allTimes
```
david-yz-liu: You should be able to just do `foldl addCourseToSchedule schedule allTimes`, perhaps if you flip the order of arguments.

app/Response/Export.hs
```diff
@@ -14,18 +14,21 @@ exportGraphResponse :: String -> String -> ServerPart Response
...
+returnPDF :: String -> String -> String -> String -> String -> String -> IO String
+returnPDF graphSvg graphImg fallTimetableSvg fallTimetableImg springTimetableSvg springTimetableImg = do
     rand <- randomName
     let texName = rand ++ ".tex"
         pdfName = rand ++ ".pdf"
-    generateTex [graphImg, timetableImg] texName -- generate a temporary TEX file
-    createPDF texName                            -- create PDF using TEX and delete the TEX file afterwards
-    removeImage (graphSvg ++ " " ++ graphImg ++ " " ++ timetableSvg ++ " " ++ timetableImg)
+    generateTex [graphImg, fallTimetableImg, springTimetableImg] texName -- generate a temporary TEX file
+    createPDF texName                            -- create PDF using TEX and delete the TEX file afterwards
```
david-yz-liu: There are some places with trailing whitespace that you should remove.

app/Routes.hs
```diff
@@ -25,7 +28,7 @@ routes staticDir aboutContents privacyContents = [
     ("course-info", look "dept" >>= courseInfo),
     ("depts", liftIO deptList),
     ("timesearch", searchResponse),
-    ("calendar", lookCookieValue "selected-lectures" >>= calendarResponse),
+    -- ("calendar", lookCookieValue "selected-lectures" >>= calendarResponse),
```
david-yz-liu: Restore this route, and rename your "calendar" route back to "timetable-image". I guess you should add a new link to the export modal.

public/js/common/grid_image.js
```diff
@@ -8,12 +8,12 @@ function getGridImage(session) {
...
     $.ajax({
-        url: 'timetable-image',
-        data: {courses: courses, session: session},
+        url: 'calendar',
+        data: {session: session},
         success: function (data) {
             var contentDiv = $('<div></div>');
             var topContentDiv = $('<div></div>');
-            var calendarOption = $('<a href="calendar">Download ICS</a>');
+            var calendarOption = $('<a href="timetable-pdf" onclick="getPDF()">Download ICS</a>');
```
david-yz-liu: You don't need a new AJAX request here (no `onclick` necessary). You can add a `target="_blank"` to open the PDF in a new tab.

app/Export/GetImages.hs
AilsaF: Sorry, I fail to get it...

PR #896, opened 2016-10-21
React Modal With Course Titles and Descriptions

lanaelsanyoura:
The info modal is now capable of displaying the course information of the node whose info box has been selected.

### DIFF
--- a/public/js/graph/react_graph.js
+++ b/public/js/graph/react_graph.js
@@ -160,7 +160,9 @@ var Graph = React.createClass({
             horizontalPanFactor: 0,
             verticalPanFactor: 0,
             mouseDown: false,
-            modalIsOpen: false,        };
+            modalIsOpen: false,
+            courseId: 'csc108H1'
+            };
     },
 
     componentDidMount: function () {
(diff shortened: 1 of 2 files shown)


public/js/graph/react_graph.js
```diff
@@ -379,6 +389,9 @@ var Graph = React.createClass({
 
     infoBoxMouseClick: function () {
         var infoBox = this.refs.infoBox;
+        this.setState({courseId: infoBox.state.nodeId.substring(0, 6) });
+        this.openModal();
```
david-yz-liu: This results in two separate calls to `setState`, which isn't good. You should modify these two lines to only make one call to `setState`.

public/js/graph/react_graph.js
```diff
@@ -1,5 +1,6 @@
 import * as tooltip from 'es6!graph/tooltip';
 import {Modal} from 'es6!common/react_modal';
```
david-yz-liu: This import can be removed now that this component is no longer used.

public/js/graph/react_graph.js
```diff
@@ -566,12 +578,17 @@ var Graph = React.createClass({
                             this.state.verticalPanFactor == 0;
         return (
             <div>
-                <ReactModal
-                  isOpen={true}
-                >
-                  <h1>Test modal</h1>
-                  <p>Hi Lana!</p>
+                <ReactModal className = 'ModalClass'
```
david-yz-liu: Don't put spaces around `=` for passing props

public/js/graph/react_graph.js
```diff
@@ -1026,7 +1045,11 @@ var Node = React.createClass({
 
 var BoolGroup = React.createClass({
     componentDidMount: function () {
-        for (var ref in this.refs) {
+        console.log(this, this.refs);
```
david-yz-liu: Remove the `console.log` calls, and the commented-out line below.

david-yz-liu:
@lanaelsanyoura I realized that you changed the `application.css` file, but this is not the right way to update the CSS. Instead, you'll need to learn how to use Clay and modify `hs/Css/Graph.hs`.

PR #905, opened 2016-11-10
Wrapper modal

lanaelsanyoura:
Implemented a wrapper class on the React Modal but experienced an " Uncaught Syntax Error: Unexpected Token (1342:12)' when running the program.


public/js/graph/react_graph.js
```diff
@@ -1347,5 +1333,36 @@ var InfoBox = React.createClass({
...
+                    <div className='modal-header'>
+                        {this.state.courseTitle}
+                          </div>
+                        <div className='modal-body'>
+                            <Description course  = {this.state.formatCourse}/> </div>  
+                </ReactModal>
+                </div>
+            );
+        }
+    }
```
david-yz-liu: There is an extra `}` here. You would have caught this error if you fixed the indentation of your code.

david-yz-liu:
@lanaelsanyoura CircleCI reports a compilation error - please see the "Details" of the check for more information, though you should be able to get the same error when you compile your code on your machine.

david-yz-liu:
@lanaelsanyoura I also noticed you seem to have pulled in some changes from the master branch (in general don't do this). But I updated the modal-project branch, so you'll need to do a pull from `upstream modal-project` to update your branch. GitHub reports merge conflicts, but they are likely just because you committed the same changes, so they should be straightforward to resolve.

public/js/common/react_modal.js
```diff
@@ -13,93 +15,80 @@ export var Modal = React.createClass({
...
+            this.setState({modalIsOpen: true, courseId: newCourse, courseTitle: getCourseTitle(newCourse, formatted)});
 
-                        </div>
-                    </div>
-                </div>
-            );
-        } else {
-            return <div className='modal fade'></div>;
+            $.ajax({
+                    url: 'course',
```
david-yz-liu: @lanaelsanyoura you seem to be using tabs instead of spaces (here, and below). Please double-check all of your changed files and convert any tabs to four spaces.

public/js/common/react_modal.js
```diff
@@ -13,93 +15,80 @@ export var Modal = React.createClass({
         return {
             courseId: '',
             course: [],
-            sessions: []
+            sessions: [],
+            modalIsOpen: false,
+            courseTitle : '',
+            formatCourse : '',
```
david-yz-liu: Is the `formatCourse` state actually being used? Please double-check.

public/js/common/react_modal.js
lanaelsanyoura: Thank you professor I'll do so.

public/js/common/react_modal.js
lanaelsanyoura: I'm using formatCourse as a state so that I only format a course once and then pass that as a parameter for the getCourseTitle function (which previously used to call the format course function)

public/js/graph/react_graph.js
```diff
@@ -159,9 +157,8 @@ var Graph = React.createClass({
             horizontalPanFactor: 0,
             verticalPanFactor: 0,
             mouseDown: false,
-            modalIsOpen: false,
-            courseId: 'csc108H1'
-            };
+            courseId: ''
```
lanaelsanyoura: Yes you're right, I don't call it when opening the modal, I use the courseId straight from the info box. I will fix this as well.

public/js/common/react_modal.js
david-yz-liu: While you are doing *something* like that, notice that you never actually access `this.state.formatCourse`.

PR #911, opened 2016-11-15
Issue 892

RyanDJLee:
- Fix implementation of Show for From and Progreq.

### DIFF
--- a/app/Database/Requirement.hs
+++ b/app/Database/Requirement.hs
@@ -24,9 +24,9 @@ instance Show ProgramReq where
 data CourseReq = CRSREQ String Req Req Req
 
 instance Show CourseReq where
-    show (CREQ course creq excl preq) = "Corequisites for " ++ course ++ ":\n" ++ show req ++ "\n"
-                                      ++ "Exclusions for " ++ course ++ ":\n" ++ show req ++ "\n"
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@RyanDJLee I wonder if the reason the compilation error wasn't detected before is that this module isn't listed in `courseography.cabal`. You should add it there and try to run a "stack build".

The module name should be changed to `Database.Requirements` (at the top of the file), and added to the cabal file.

app/Database/Requirement.hs
```diff
@@ -12,28 +12,28 @@ module Requirement
 ) where
 
 import qualified Data.String as S
+import qualified Data.List as L
 
-
-data ProgramReq = PRGREQ String [Req]
```
david-yz-liu: I'm not sure why you changed this; it seems to be the case that a program will have multiple requirements. I guess you could represent this as an implicit "AND", but it will be useful to keep them separate at the top level, I think.

RyanDJLee:
Changed it back! It does make a lot more sense to keep program requirements as a list of requirements.

RyanDJLee:
- Changed value constructor of program requirements back to list of reqs.
- Changed name of Requirement module and added it to list the of dependencies.
- Build worked well and tested with stack repl.

david-yz-liu:
Thanks, @RyanDJLee!

PR #914, opened 2016-12-09
setup check

AilsaF:


AilsaF:
And you said you want to check db after getting server start, so I put the check in server.hs. But it seems ok to check it before server starts, since "db" is a fixed path.

Setup.hs
```diff
@@ -1,2 +1,25 @@
...
+            configExistbool <- doesFileExist "app/Config.hs"
+            case configExistbool of
+                False -> copyFile "app/DevelopmentConfig.hs" "app/Config.hs" >> return emptyHookedBuildInfo
+                _     -> return emptyHookedBuildInfo
+
+        -- | check that Imagemagick and LaTeX are available
+        checkDependencies _ _ _ _ = do
+            mapM_ check ["convert", "pdflatex"]
+        check :: String -> IO ()
+        check depdency = do
```
david-yz-liu: I think you meant to name the parameter "dependency".

app/Server.hs
```diff
@@ -25,6 +25,8 @@ runServer = do
     staticDir <- getStaticDir
     aboutContents <- LazyIO.readFile $ markdownPath ++ "README.md"
     privacyContents <- LazyIO.readFile $ markdownPath ++ "PRIVACY.md"
+    -- Create db folder if it doesn't exist
+    createDirectoryIfMissing False "db"
```
david-yz-liu: This path shouldn't be hard-coded to "db" - it's set in the configuration file, I believe.

david-yz-liu:
@AilsaF I think you're on certainly on the right track. :)

I think the `HookedBuildInfo` is specific to the setup infrastructure, and so using the `Directory` package is certainly a good idea.

AilsaF:
Since lack of db folder would cause error for ```stack exec courseography database```, I figured it suppose to be handled before ```stack exec courseography```. And it seems that Setup.hs cannot import Config.hs. Thus, I put creating db in Database.hs.

app/Database/Database.hs
```diff
@@ -16,18 +16,26 @@ import Database.Tables
...
-                   print "Distribution table set up"
-                   setupBreadthTable
-                   print "breadth table set up"
-                   parseAll
-                   seedVideos
+setupDatabase = do
+    -- Create db folder if it doesn't exist
+    let ind = fromMaybe (T.length databasePath) (T.findIndex (=='/') databasePath)
+        db = T.unpack $ T.take ind databasePath
+    createDirectoryIfMissing False db
```
david-yz-liu: This should probably actually be True, not False. (https://hackage.haskell.org/package/directory-1.3.0.0/docs/System-Directory.html)

AilsaF: Hmm... Not sure I got what you meant. But I tried both False and True, and both of them created db folder under courseography. And both contained ```..```, which means they both created parent successfully.

david-yz-liu: It doesn't matter for this case, but if you change the config file to have the path `db/some/other/folders/` then we want the entire path to be created.

But now I realize that your current algorithm won't actually do that anyway (it seems to look for the first occurrence of a `/` rather than the last), so that should also be updated.

david-yz-liu: Yeah, that's right. (That's what the `True` should do.)

david-yz-liu:
Thanks, @AilsaF!

PR #923, opened 2017-01-16
Li parser

christinem:
This commit includes the code for the "li" parser. It still includes the two problems I mentioned:

- If something has "li"s within other places in the post (example, CS), but it’s not the way the categories are split up, it’ll only parse the "li"s
- If the post uses "li"s before the requirements (example, Psychology), it’ll parse those too

Refactoring will come in another pull request.


david-yz-liu:
Thanks, @christinem. I didn't make any comments here because you'll be working on refactoring anyway, and I gave you plenty of comments last meeting. But let me know if you want to chat more!

PR #931, opened 2017-02-02
Schema design

AilsaF:
So far: 
added postType data type and modify post parsing code
basically done String -> T.Text


app/Database/CourseQueries.hs
```diff
@@ -168,7 +168,7 @@ buildSession lecs tuts =
 getGraphJSON :: String -> IO Response
 getGraphJSON graphName =
     runSqlite databasePath $ do
-        graphEnt :: (Maybe (Entity Graph)) <- selectFirst [GraphTitle ==. graphName] []
+        graphEnt :: (Maybe (Entity Graph)) <- selectFirst [GraphTitle ==. T.pack graphName] []
```
david-yz-liu: Don't change this; instead, change the type of the `graphName` parameter to `T.Text`, and in the Routes.hs file, use `lookText'` (see [here](http://hackage.haskell.org/package/happstack-server-7.4.6.2/docs/Happstack-Server-RqData.html#v:lookText-39-)) instead of `look` to get the query parameter.

app/Database/DataType.hs
```diff
@@ -17,3 +17,17 @@ derivePersistField "ShapeType"
 
 instance ToJSON ShapeType
 instance FromJSON ShapeType
+
+-- data AlignType = Begin | Start | Middle | End
```
david-yz-liu: Please remove this from this branch (of course this is a good change to make, but should be done on a separate branch).

app/Export/GetImages.hs
```diff
@@ -26,7 +27,7 @@ getActiveGraphImage req = do
         graphName =
             replace "-" " " $
                 maybe "Computer-Science" cookieValue (M.lookup "active-graph" cookies)
-    getGraphImage graphName (M.map cookieValue cookies)
+    getGraphImage graphName (M.mapKeys T.pack $ M.map cookieValue cookies)
```
david-yz-liu: No need to use `T.pack` here - use `readCookieValue` instead: http://hackage.haskell.org/package/happstack-server-7.4.6.2/docs/Happstack-Server-RqData.html#v:lookCookieValue

app/Svg/Builder.hs
```diff
@@ -66,10 +67,10 @@ buildRect texts entity elementId =
                             0  -- no tolerance for text intersection
                             . textPos
                             ) texts
-        textString = concatMap textText rectTexts
+        textString = T.concat $ map textText rectTexts
```
david-yz-liu: It appears there's a `T.concatMap`?

app/Svg/Parser.hs
```diff
@@ -159,8 +160,8 @@ parseTextHelper key styles' trans textTags =
         currTrans = getTransform $ head textTags
         newTrans = addTuples trans currTrans
         alignAttr = styleVal "text-anchor" newStyle
-        align = if null alignAttr
-                then "begin"
+        align = if T.null alignAttr
+                then T.pack "begin"
```
david-yz-liu: Don't call `T.pack` on string literals. Instead, we use the GHC extension `OverloadedStrings` (this automatically converts a string literal like `"begin"` into the correct type. This is already present at the top of this file, so it should just work.

AilsaF:
I have done with what could be modified so far. However, it still includes some issues.

- Haven't modified so far. 
  - I didn't modify *Requirements.hs* and *ReqParser.hs*, since I suppose Ryan will keep work on Parser as we have discussed in last meeting. 
  - And also left *ParseCombinator.hs and PostParser.hs*, as noticing Christine working on it. (p.s. I commended out line97 in *ArtSciParser.hs*. Will change it back once I am able to harmonize it with *PostParser.hs*)

- Prefer leaving as String
  - Couldn't find similar function working on Text. (e.g. *Tables.hs* line 336 `Text` is not an instance of `Read`; cmd)
  - Strings interacting with tons of `++`, which is better in term of efficiency, I suppose. (e.g. filenames). But I was terribly indecisive on whether modifying *Calendar* Time part... 😣

- For *PrerequisiteParsing.hs*, I felt it did same thing as Ryan's Parser, not sure whether it will be kept in the future. Please let me know if you want to revise it. (BTW, there is something wrong with *function isntDelim* line 29, it returns True though str is "(STA247, STA248)"  )

What a long list, phew!

david-yz-liu:
Thanks, @AilsaF! I'm fine with all the decisions you made. I'm surprised that `Text` is no longer an instance of `Read` (I think this must have changed in a newer version of the library...).

PR #935, opened 2017-02-20
Issue933: fix Wunused-imports and  Wtype-defaults stack build errors

frankjohnlee:
Request to pull to fix Wunused-imports and  Wtype-defaults stack build errors which appear upon running Stack Build.


app/WebParsing/PostParser.hs
```diff
@@ -9,14 +9,17 @@ import Data.List
...
 
 fasCalendarURL :: String
 fasCalendarURL = "http://calendar.artsci.utoronto.ca/"
 
+failedString :: String
+failedString = "Failed."
+
 getPost :: String -> IO ()
-getPost str = do
+    getPost str = do
```
david-yz-liu: This shouldn't be indented.

app/Svg/Generator.hs
```diff
@@ -44,6 +44,10 @@ import qualified Data.Map.Strict as M
 import Data.Monoid (mempty, mappend, mconcat)
 import Config (databasePath)
 
+hybridFontSize :: Integer
```
david-yz-liu: Don't change this here; change it in `Css.Constants` if you need to.

frankjohnlee: Okay I will take a look at this thanks David.

app/Css/Constants.hs
```diff
@@ -455,13 +455,14 @@ nodeFontSize :: Num a => a
 nodeFontSize = 12
 
 -- |Defines hybrid font size, 7 in pixels.
-hybridFontSize :: Num a => a
+hybridFontSize :: Integer a => a
```
david-yz-liu: This isn't correct - you're mixing up a type class (`Num`) with a concrete type (`Integer`). Use `hybridFontSize :: Integer`. Also, this change wouldn't have compiled, so this is indicating that you aren't compiling your code before submitting a pull request, which you should make sure to always do.

frankjohnlee: When I write in the following code:
hybridFontSize :: Num a => a
hybridFontSize = 7
hybridFontSize :: Integer

I get the error:
C:\Users\Frank2\Desktop\Courseography\courseography\app\Css\Constants.hs:460:1: error:
    Duplicate type signatures for `hybridFontSize'
    at app\Css\Constants.hs:458:1-14
       app\Css\Constants.hs:460:1-14

Am I doing this incorrectly David?

PR #937, opened 2017-02-20
Issue936: Fix Wname-shadowing Warnings

frankjohnlee:


app/Database/PostInsertion.hs
```diff
@@ -7,18 +7,17 @@ module Database.PostInsertion(
 import qualified Data.Text as T
 import Config (databasePath)
 import Database.Persist.Sqlite (insert_, runSqlite, runMigration, SqlPersistM)
-import Database.Tables
+import Database.Tables hiding (postCode, description, postCategoryName)
 
 -- | Insert a new post into the database
 insertPost :: T.Text -> T.Text -> T.Text -> T.Text -> IO ()
-insertPost departmentName postName postCode description =
+insertPost departmentName1 postName0 postCode description =
```
david-yz-liu: In general, don't append a number; instead, append an underscore (e.g., `departmentName_`).

frankjohnlee: Oh okay I will fix these and let you know David.

app/Database/CourseInsertion.hs
```diff
@@ -46,17 +46,17 @@ saveGraphJSON jsonStr nameStr = do
...
     -- option: keyListDistribution :: [DistributionId] <- selectKeysList [ DistributionDescription `contains'` description] []
     return $ case keyListDistribution of
         [] -> Nothing
         _ -> Just (head keyListDistribution)
 
 getBreadthKey :: Maybe T.Text -> SqlPersistM (Maybe (Key Breadth))
 getBreadthKey Nothing = return Nothing
-getBreadthKey (Just description) = do
-    keyListBreadth :: [Key Breadth] <- selectKeysList [ BreadthDescription ==. (T.unpack description) ] []
+getBreadthKey (Just description__) = do
```
david-yz-liu: The `description_` above is a parameter of `getDistributionKey`, so it's local to that function. You can use the same parameter name here in `getBreadthKey` without getting the shadowing warning.

PR #945, opened 2017-02-20
Merge ltp

AilsaF:
WebParsing.UtsgJsonParser  and  Response.Calendar still have `Either`


david-yz-liu:
@AilsaF beautiful work. Thanks very much!

PR #948, opened 2017-02-28
Issue938 ([-Wunused-do-bind])

frankjohnlee:
Fixed all Wunused-do-bind warnings. 
Ran compiler on each fix to make sure none of my change broke any code.

### DIFF
--- a/app/MasterTemplate.hs
+++ b/app/MasterTemplate.hs
@@ -11,8 +11,6 @@ import Config (enableFb, enableCdn)
 import Util.Blaze
 import Scripts (globalScripts)
 
-
-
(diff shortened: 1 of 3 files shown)


app/MasterTemplate.hs
```diff
@@ -11,6 +11,8 @@ import Config (enableFb, enableCdn)
 import Util.Blaze
 import Scripts (globalScripts)
 
+
```
david-yz-liu: @frankjohnlee please delete these two blank lines (and in general review your commits so that you don't include extra whitespace).

frankjohnlee: Hey David. I'll make these changed. 

I just wanted to note that I didn't get an email alert for this comment. I think it's because I won't get alerted if you comment on you're review?

PR #953, opened 2017-03-07
Fixed errors Pattern matched are not exhaustive. Though I left parser…

frankjohnlee:
Fixed errors Pattern matched are not exhaustive. Though I left parser…


.gitignore
```diff
@@ -28,3 +28,4 @@ cabal.sandbox.config
 
 # IDE project files
 .idea
+app/Svg/Generator.hs
```
david-yz-liu: @frankjohnlee please revert this change. This is not appropriate (either for this specific pull request, or in general).

app/Svg/Parser.hs
```diff
@@ -362,6 +362,7 @@ updateShape fill r =
                               Hybrid   -> Hybrid
                               BoolNode -> BoolNode
                               Node     -> Node
+                              _ -> ""
```
david-yz-liu: @frankjohnlee This is the right idea, but contains a type error. You can use `undefined` instead of `""` here.

frankjohnlee: Thanks David. I have made this change.

PR #964, opened 2017-03-14
Issue 960

RyanDJLee:
- Run test suites using cabal test suite interface exitcode-stdio-1.0 and test-framework library.
- Update test cases for all parsers.
- Instead of using QuickCheck, manually scraping for test cases from ArtSci webpage.
- Next step is to get parsers to exhibit expected behaviour and then potentially create more test cases, repeat.


david-yz-liu:
@RyanDJLee I want to see this work split up into separate branches. On your `issue-930` branch, you should put in the work for setting up the unit testing, since I asked you to do that on that branch. On a new branch, you can start adding new tests. However, these should be separated. Please resolve this before adding any new tests.

PR #966, opened 2017-03-14
Issue965 [-Wunused-imports] Round 2

frankjohnlee:


app/Export/TimetableImageCreator.hs
```diff
@@ -11,7 +11,7 @@ import Diagrams.Prelude
 import Diagrams.Backend.SVG.CmdLine
 import Diagrams.Backend.SVG
 import Data.List (intersperse)
-import Data.List.Split (splitOn)
+import Data.List.Split()
```
david-yz-liu: @frankjohnlee If this is due to an unused import warning, you can remove this line. Same comment for `app/Svg/Generator` below.

frankjohnlee: Thank you David I removed both and will push.

david-yz-liu:
@frankjohnlee I believe you have some extra files in this pull request that you should remove.

PR #977, opened 2017-04-19
upgrade-js-workflow: Incorporate yarn, webpack, and babel

david-yz-liu:


app/Server.hs
```diff
@@ -8,26 +10,40 @@ module Server
...
 import qualified Data.Text.Lazy.IO as LazyIO
 import Routes (routes)
+import Data.List.Utils (replace)
+import GHC.IO.Handle (hClose)
+
+webpackScript :: Path.FilePath
+#ifdef mingw32_HOST_OS
+webpackScript = Path.concat ["node_modules", ".bin", "webpack.cmd"]
+#else
+webpackScript = Path.concat ["node_modules", ".bin", "webpack.cmd"]
```
arkon: This probably shouldn't have `.cmd`.

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "courseography",
+  "version": "0.2.0.0",
+  "main": "index.js",
```
arkon: This can probably be removed since there's no need for a Node module entry point.

package.json
```diff
@@ -0,0 +1,19 @@
...
+  "repository": "git@github.com:Courseography/courseography.git",
+  "author": "David Liu <david@cs.toronto.edu>",
+  "license": "MIT",
```
arkon: It's actually GPLv3.

webpack.config.js
```diff
@@ -0,0 +1,18 @@
...
+    'js/search/timetable': './js/timetable.js.jsx'
+  },
+  output: {
+    path: path.resolve(__dirname, 'public'),
+    publicPath: './public',
+    filename: '[name].js'
+  },
+  module: {
+    loaders: [
+      { test: /\.js$/, loader: 'babel-loader', exclude: /node_modules/ },
```
arkon: You could just do a single loader with the test being `/\.jsx?$/`.

david-yz-liu: I think I'm going to leave this one.

david-yz-liu:
Thanks for the feedback, @arkon. I've been meaning to do this for a while, and so I took some time before the summer term.

circle.yml
```diff
@@ -1,3 +1,7 @@
+machine:
+  node:
+    version: 6.1.0
```
arkon: `setup_linux` seems to use v7.x of Node, while CircleCI will be using v6.x. It might be a good idea to make it consistent?

package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "courseography",
+  "version": "0.2.0.0",
+  "repository": "git@github.com:Courseography/courseography.git",
+  "author": "David Liu <david@cs.toronto.edu>",
+  "license": "GPLv3",
+  "dependencies": {
+    "path": "^0.12.7",
```
arkon: I feel like it might make more sense for these to be in `devDependencies`, while things like React and ReactDOM would be in here.

webpack.config.js
```diff
@@ -0,0 +1,18 @@
+const path = require('path');
+
+module.exports = {
+  entry: {
+    'js/search/timetable': './js/timetable.js.jsx'
```
arkon: I might be missing something, but isn't this file actually in `./public/js/...`?

package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "courseography",
+  "version": "0.2.0.0",
```
arkon: Traditional semver would just be "0.2.0"

.gitignore
```diff
@@ -31,3 +31,9 @@ cabal.sandbox.config
 
 # IDE project files
 .idea
+
+# Node (Javascript) modules
+/node_modules
+
+# TODO: eventually remove all public/js files from tracking
+/public/js/search
```
arkon: So is this where the transpiled output files would go? It seems a bit confusing as to where the source and built files go.

circle.yml
david-yz-liu: I imagine that by the time this matters for us, CircleCI will support v7 of Node.

package.json
david-yz-liu: This is the version in our cabal file right now.

package.json
david-yz-liu: I'll take a look at this later.

webpack.config.js
david-yz-liu: I've moved the raw file out of `public/`. Eventually I'll do the same for the other files.

.gitignore
christinem: It looks like he's only transpiling the search page for now based on the webpack.config file and their respective /public/js/* folder is where they would go. So I assume eventually as more pages get added, more will get added to the gitignore.

scripts/setup_osx
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+% Install software dependencies from homebrew
```
christinem: The % character needs to be a # character to make it a comment.

PR #989, opened 2017-05-18
Hook Post Parsing back in

christinem:
This PR hooks the post parsing back in based on the new updated calendar.

A couple of notes
* There's a little hack for Post parsing failures and Focuses (since Focuses now follow the same structure and are at the same level as Posts, I need to remove those still), it gets inserted into the Post table with a Post Type of "Other" and everything else blank
* Since the unique post codes are no longer embedded within the HTML like they were before, I've left the Post Code column of the Post Category table blank. Only thing is we've lost the connections between the categories and what Post they belong to. We could use a Post ID instead to recreate that connection for now, but I haven't added that in yet.
* Need to fix the new warnings I introduced too.

Otherwise, the parsing now works properly again.

### DIFF
--- a/app/WebParsing/ArtSciParser.hs
+++ b/app/WebParsing/ArtSciParser.hs
@@ -21,6 +21,7 @@ import Database.CourseInsertion (insertCourse)
 import Database.Tables (Courses(..), Department(..))
 import WebParsing.ReqParser (parseReqs)
 import Config (databasePath)
+import WebParsing.PostParser (addPostToDatabase)
 
 
 -- | The URLs of the Faculty of Arts & Science calendar.
@@ -81,20 +82,10 @@ parseDepartment (relativeURL, _) = do
 parsePrograms :: [Tag T.Text] -> SqlPersistM ()
 parsePrograms programs = do
     let elems = TS.partitions isPost programs 
-    mapM_ parsePost elems
+    mapM_ addPostToDatabase elems
     return ()
     where
          isPost tag = tagOpenAttrNameLit "h3" "class" (T.isInfixOf "programs_view") tag
-
--- | Parse a particular post in a program section
-parsePost :: [Tag T.Text] -> SqlPersistM ()
-parsePost programElements = do
-    -- TODO: Remove Focuses from programElements
-    -- TODDO: Store name of post before we lose that information in the next line
-    let requirements = map TS.innerText $ TS.sections isRequirementSection programElements
-    liftIO (print requirements)
-    where
-        isRequirementSection element = tagOpenAttrLit "div" ("class", "field-content") element
 
 -- | Parse the section of the course calendar listing the courses offered by a department.
 parseCourses :: [Tag T.Text] -> [(Courses, T.Text, T.Text)]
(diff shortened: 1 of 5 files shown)


christinem:
Not sure why the tests failed, it's compiling and running fine on my computer.

app/WebParsing/ArtSciParser.hs
```diff
@@ -39,7 +40,7 @@ parseArtSci = do
         liftIO $ putStrLn "Inserting departments"
         insertDepts $ map snd deptInfo
         mapM_ parseDepartment deptInfo
-
+        --parseDepartment ("/section/Computer-Science","Computer Science")
```
david-yz-liu: You can remove this line.

app/WebParsing/ArtSciParser.hs
```diff
@@ -78,10 +79,12 @@ parseDepartment (relativeURL, _) = do
 
 -- | Parse the section of the course calendar listing the programs offered by a department.
 parsePrograms :: [Tag T.Text] -> SqlPersistM ()
-parsePrograms _ = do
-    -- let elems = TS.partitions (TS.isTagOpenName "h3") _ -- TODO: complete this function
+parsePrograms programs = do
+    let elems = TS.partitions isPost programs 
+    mapM_ addPostToDatabase elems
     return ()
```
david-yz-liu: You can delete this line

app/WebParsing/ParsecCombinators.hs
```diff
@@ -26,23 +26,32 @@ getCourseFromTag courseTag =
...
-postInfoParser :: Maybe T.Text -> T.Text -> Parser Post
-postInfoParser firstCourse postCode = do
+postInfoParser :: T.Text -> Maybe T.Text -> Parser Post
+postInfoParser fullPostName firstCourse = do
+    let parsed = P.parse getDeptNameAndPostType "(source)" fullPostName
+    case parsed of
+        Right (deptName, postType) -> do
+            programDescription <- getRequirements firstCourse
+            return $ Post (read $ T.unpack postType) deptName (T.pack " ") programDescription
+        Left _ -> return $ Post (read "Other") (T.pack " ") (T.pack " ") (T.pack " ")
```
david-yz-liu: For now why don't you put in `fullPostName` for the post name in this case.
Also if you import the `Other` constructor, it's better to use that than `read "Other"`

app/WebParsing/PostParser.hs
```diff
@@ -20,46 +20,23 @@ fasCalendarURL = "http://calendar.artsci.utoronto.ca/"
...
-
-addPostToDatabase :: [Tag String] -> SqlPersistM ()
-addPostToDatabase tags = do
-    let postCode_ = T.pack (fromAttrib "name" ((take 1 $ filter (isTagOpenName "a") tags) !! 0))
-        liPartitions = partitions isLiTag tags
-        programPrereqs = map getCourseFromTag $ map (T.pack . fromAttrib "href") $ filter isCourseTag tags
+addPostToDatabase :: [Tag T.Text] -> SqlPersistM ()
+addPostToDatabase programElements = do
+    -- TODO: Remove Focuses from programElements
+    let fullPostName = innerText $ take 1 $ filter isTagText programElements
```
david-yz-liu: Use `head` instead of `take 1`

christinem: `innerText` needs a list of Tags, which is what `take 1` gives you, but `head` only gives the element itself.

PR #991, opened 2017-05-18
Minfan_dynamicgraph_graphgenerator

minfanzhang:
These are some reasonable trials I have made, aiming to use Graphviz in Haskell to create images, but none of them worked. The details are as follows:

1.A tutorial showing some samples, claiming to create some images as required(http://haroldcarr.com/posts/2014-02-28-using-graphviz-via-haskell.html). But the code it provided does not work. For instance, in "ex3" of the sample, when I run his code, I got "  14:11: error: parse error on input ‘[’  " in the line "  graph [rankdir=LR];  ", it seems that graph is not admitted as a valid variable..

[10:44] 
2.the "dataflow" package of Haskell with its doc saying it creates outputs of graph and we can write just .gv like codes to run in Haskell, but I lacks enough documentation and any sample so I did not figure out how to use it.

[10:48] 
3.A sample from stackoverflow(http://stackoverflow.com/questions/20849893/how-to-plot-a-graph-using-haskell-graphviz)but it only tells how to convert a graph in Data.Graph.Inductive.Example to a dot excuteble program, which is the only one worked up to now. Unfortunately, it gives nothing about writing code for a general graph and generate the image with Graphviz..


minfanzhang:
Now able to generate any tree-like graph according to the course structure by writing the corresonding codes.

Now, I am trying to embed this as a whole program that can take inputs and generate images automatically instead of writing the codes by hand..

Currently it is not standalone excuteble yet..I am trying to create a relative complete work before making a pull. It seems that this requires more knowledge of Haskell and I am learning it now.

minfanzhang:
I revised the .cabal file and rebuild the Courseograph project. The Courseograph worked but the graph is not created in the "tmp" folder in the folder of my module, is it because I did not do some very fundamental step in order to let my script run or it is a real bug?

minfanzhang:
Now eliminated all warnings and the dynamic graph part can be run as a toy piece

PR #993, opened 2017-06-08
change selected-lecture to local storage

AilsaF:


app/Export/GetImages.hs
```diff
@@ -15,43 +15,44 @@ import qualified Data.Text as T
...
-    let cookies = M.fromList $ rqCookies req
-        graphName =
-            T.pack $
-            replace "-" " " $
-                maybe "Computer-Science" cookieValue (M.lookup "active-graph" cookies)
-    getGraphImage graphName (M.mapKeys T.pack $ M.map (T.pack . cookieValue) cookies)
+getActiveGraphImage :: String -> IO (String, String)
+getActiveGraphImage jsonLocalStorage = do
+    let localStorage = fromJust $ decode $ fromStrict $ BC.pack jsonLocalStorage :: M.Map T.Text T.Text
+        graphName = T.replace "-" " " $ fromMaybe "Computer-Science" $ M.lookup "active-graph" localStorage
```
david-yz-liu: When we used to use cookies, we had trouble getting spaces in the cookies to be parsed. I'm wondering if this is no longer true. Can you please try to store the graph names with spaces instead of hyphens (e.g., "Computer Science") and see if the loading works properly?

app/Export/GetImages.hs
```diff
@@ -15,43 +15,44 @@ import qualified Data.Text as T
...
-getActiveGraphImage :: Request -> IO (String, String)
-getActiveGraphImage req = do
-    let cookies = M.fromList $ rqCookies req
-        graphName =
-            T.pack $
-            replace "-" " " $
-                maybe "Computer-Science" cookieValue (M.lookup "active-graph" cookies)
-    getGraphImage graphName (M.mapKeys T.pack $ M.map (T.pack . cookieValue) cookies)
+getActiveGraphImage :: String -> IO (String, String)
+getActiveGraphImage jsonLocalStorage = do
```
david-yz-liu: We shouldn't refer to "local storage" in variable names in the Haskell code - from the server's point of view, it got the request data from somewhere, but it doesn't need to know where. Please choose more general names throughout your changes.

public/js/common/export/export.js
```diff
@@ -37,8 +37,10 @@ export var ExportModal = React.createClass({
     },
 
     getGraphImage: function() {
+        var JsonLocalStorageObj = JSON.stringify(localStorage);
```
david-yz-liu: It isn't good to send the *entire* local storage as JSON to the server. Instead, for all these requests you should only extract the key-value pairs that are necessary for that particular request. This should have the added benefit of making the server code a little simpler.

public/js/common/export/export.js
```diff
@@ -108,7 +149,7 @@ var GraphImage = function (props) {
             Export
         </div>
         <div className='modal-body'>
-            <a href="timetable-pdf" target="_blank">Download PDF</a>
+            <a href="#" onClick={getPDF}>Download PDF</a>
```
david-yz-liu: You can just remove the `href` attribute altogether.

public/js/common/export/export.js
AilsaF: Yeah I totally agree with you and actually I thought about when writing this code. But I didn't see how to filter out necessary courses for current graph... Or maybe I can delete some LS like 'bool1' (they are really minority of LS though)?

david-yz-liu: This is a good point. I believe local storage values can themselves be dictionaries, so it would be better even to just put all the course info under a nested "courses" dictionary in the local storage. (You could create/modify a separate helper for accessing them.)

PR #997, opened 2017-07-20
Database fixes

christinem:
This includes database fixes, including:

- Connecting post categories back to their posts in the database
-  Fixing the order of parsers (numbered and list)
- Fixing college pages and some special case pages (ex: Classics) not being parsed properly

Missing: 
- Checking for any remaining ad-hoc cases 
- Dealing with college pages being parsed multiple times, causing duplicates in posts and courses. Either we can fix this when collecting all the page links to parse, and not to parse the same college page more than once, or find some new primary key so that duplicates can't be entered into the database. Maybe (name, department) can be a primary key - them together should be unique.


app/WebParsing/ArtSciParser.hs
```diff
@@ -69,25 +69,31 @@ parseDepartment (relativeURL, _) = do
...
     where
-         isPost tag = tagOpenAttrNameLit "h3" "class" (T.isInfixOf "programs_view") tag
+         isPost tag = tagOpenAttrNameLit "h3" "class" isProgramsView tag
+         isProgramsView currentTag = or [(T.isInfixOf "programs_view") currentTag, (T.isInfixOf "_programs") currentTag]
 
 -- | Parse the section of the course calendar listing the courses offered by a department.
 parseCourses :: [Tag T.Text] -> [(Courses, T.Text, T.Text)]
 parseCourses tags =
-    let elems = drop 1 $ TS.partitions (TS.isTagOpenName "h3") tags -- Remove the first one, which is the header
+    let elems = TS.partitions (tagOpenAttrNameLit "h3" "class" (T.isInfixOf "views-accordion")) tags
```
christinem: I removed the `drop 1` because we were actually just excluding the first course of each page all the time

PR #999, opened 2017-07-22
react-grid

PkHutch:


public/js/grid/react_util.js
```diff
@@ -39,29 +47,50 @@ var CourseList = React.createClass({
...
+                    <form>
+                        <input id="course-filter" className="form-control" placeholder="Enter a course!" autoComplete="off" type="text" />
+                    </form>
+                </div>
+                <div id="search-container">
+                    {/* onKeyUp is a prop, so the only way that I know how to
+                        set a prop is below like I have done with courses, but
+                        I also don't know how to access enableSearch in
+                        CourseList so I can set it to onKeyUp.*/}
+                    <CourseList courses={this.state.courseList} onKeyUp={CourseList.enableSearch}/>
```
david-yz-liu: Okay, so a couple of things.

1. The `onKeyUp` should be put onto the `input` above, not `CourseList`. That's the element where the user will do the typing.
2. Set a `ref` for this component. https://facebook.github.io/react/docs/refs-and-the-dom.html
3. Then above, you can properly refer to calling `enableSearch` in this component by passing something like `{this.refs.myRef.enableSearch}`.

PR #1006, opened 2017-08-28
Dynamic graph

minfanzhang:
8.28, made the pull request for the review of the summer


app/DynamicGraphs/WriteRunDot.hs
```diff
@@ -12,5 +12,8 @@ doDots cases = do
 createImage :: PrintDotRepr dg n => (FilePath, dg n) -> IO FilePath
 createImage (n, g) = createImageInDir (normalise "graphs/gen") n Png g
 
+-- (question)did not find fuction addExtension under Data.GraphViz
```
david-yz-liu: `addExtension` is actually found under [Data.GraphViz.Commands](https://hackage.haskell.org/package/graphviz-2999.19.0.0/docs/Data-GraphViz-Commands.html#v:addExtension), but this is *re-exported* by `Data.GraphViz` so you only need to import the latter.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,110 @@ module DynamicGraphs.GraphGenerator
...
+--        pack_string2 = pack string2
+--        gradeedges = [DE $ DotEdge pack_string2 pack_string1 [], DE $ DotEdge pack_string1 name []]
+
+--reqToStmts (name, RAW string1) stmtcounter = connectLst rawedges (concatMap reqToStmts [(name, NONE), (pack_string1, NONE)])
+--  where pack_string1 = pack string1
+--        rawedges = [DE $ DotEdge pack_string1 name []]
+
+--reqToStmts (_, _) stmtcounter = []
+
+data StmtsWithCounter = StmtsWithCounter{ stmtslst :: [DotStatement Text]
```
david-yz-liu: I believe I mentioned this in your last presentation, but your code will be a lot simpler if you use a type synonym like `type StmtsWithCounter = ([DotStatement Text], Int)`. You'll be able to use short pattern matching to extract the two parts of the tuple to use in your code.

Don't worry about extensibility for now -- after this is up and running, we can use some more advanced techniques to carry around mutable state later.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,110 @@ module DynamicGraphs.GraphGenerator
...
+                                        , counter   :: Int }
+
+foldUpReqLst :: [(Text, Req)] -> Int -> StmtsWithCounter
+foldUpReqLst reqlst count = foldl(\acc x -> reqToStmts x acc) StmtsWithCounter{stmtslst = [], counter = count} reqlst
+
+
+
+
+decompJString :: [Req] -> [(Text, Req)]
```
david-yz-liu: Define `decompJString`, `createAndEdge`, and `createOrEdge` in terms of `foldl`. (But you can define a helper function to pass into `foldl`.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,110 @@ module DynamicGraphs.GraphGenerator
...
+
+createOrEdge :: [Text] -> [DotStatement Text]
+createOrEdge [] = []
+createOrEdge (x:xs) = connectLst [DE $ DotEdge x "or" []] (createOrEdge xs)
+
+dragNodeIdInReq :: [Req] -> [Text]
+dragNodeIdInReq [] = []
+dragNodeIdInReq ((J text1):xs) = (pack text1):(dragNodeIdInReq xs)
+
+connectLst :: [a] -> [a] -> [a]
```
david-yz-liu: This function is redundant; you can concatenate two lists by doing `xs ++ ys`.
While this function seems to reverse the first list, I'm not sure that's intentional or necessary.
Please remove this function.

app/DynamicGraphs/GraphGenerator.hs
david-yz-liu: Oh, and also you'll want to put the definition of this type near the top of the file, and certainly above `reqToStmts`.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,110 @@ module DynamicGraphs.GraphGenerator
...
+--reqToStmts (_, _) stmtcounter = []
+
+data StmtsWithCounter = StmtsWithCounter{ stmtslst :: [DotStatement Text]
+                                        , counter   :: Int }
+
+foldUpReqLst :: [(Text, Req)] -> Int -> StmtsWithCounter
+foldUpReqLst reqlst count = foldl(\acc x -> reqToStmts x acc) StmtsWithCounter{stmtslst = [], counter = count} reqlst
+
+
+
```
david-yz-liu: Be careful with blank lines -- generally, there should only be *one* blank line between consecutive definitions.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,110 @@ module DynamicGraphs.GraphGenerator
...
+    let stmts = stmtslst (foldUpReqLst reqs 0)
     in
         buildGraph stmts
 
-reqToStmts :: (Text, Req) -> [DotStatement Text]
-reqToStmts (name, NONE) = [DN $ DotNode name []]
-reqToStmts (_, _) = []
+-- convert the original requirement data into dot statements that can be used by buildGraph to create the
+-- corresponding DotGraph objects. ([] is the [] after name the optional parameters for DotNode?)
+reqToStmts :: (Text, Req) -> StmtsWithCounter -> StmtsWithCounter
```
david-yz-liu: Since this is now a helper function for `foldl`, you can flip the order of the parameters here. That way, you'll be able to just call `foldl reqToStmts`.

app/DynamicGraphs/GraphGenerator.hs
minfanzhang: Oh, this function actually inserts the first element of xs at the front of ys, and then concatenate the rest of xs along with the new ys, so you still want to delete the function and do the procedures other ways?

app/DynamicGraphs/GraphGenerator.hs
minfanzhang: I am sorry I do not quite get it, would you mind giving some extra advise on how to adjust the design to fit the 
reqToStmts into foldl that way? Thanks!

app/DynamicGraphs/GraphGenerator.hs
minfanzhang: Here, I find it kind of difficult in practice since in many cases, I have to visit the attributes of StmtsWithCounter by its name (like stmtslst or counter), if using short form, it becomes hard or at least very inconvenient to visit them. 

For instance, in line 34 "let stmts = stmtslst (foldUpReqLst reqs 0)", I have to visit the stmtslst attribute of the StmtsWithCounter returned by foldUpReqLst, and I do not know how to use pattern match here. And similar problems in other parts.

app/DynamicGraphs/GraphGenerator.hs
david-yz-liu: Below, you have a call to `foldl` that looks like `foldl (\x y -> f y x)`. If you reorder the parameters of `f`, then instead you could write `foldl (\x y -> f x y)`, which is simplified to `foldl f`.

app/DynamicGraphs/GraphGenerator.hs
david-yz-liu: You can pattern-match on tuples in `let` statements as well. It's actually easier to do so than using records and function calls. For example, you could do

```
let stmts, _ = foldUpReqLst reqs 0
```

if you only wanted the first part of the tuple.

app/DynamicGraphs/GraphGenerator.hs
david-yz-liu: As I said, "While this function seems to reverse the first list, I'm not sure that's intentional or necessary." I have already described what this function is doing and said this is not necessary.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -7,34 +7,73 @@ module DynamicGraphs.GraphGenerator
...
-import Data.Text.Lazy (Text)
+import Data.Text.Lazy (Text, pack)
 
+type StmtsWithCounter = ([DotStatement Text], Int)
 
+-- Serves as a sort of "interface" for the whole part "dynamic graph"
 sampleGraph :: DotGraph Text
 sampleGraph = reqsToGraph [
-    ("CSC108", NONE)
+    ("CSC148", J "CSC108")
```
david-yz-liu: Great change! :)

PR #1014, opened 2017-11-02
New J constructor

RyanDJLee:
- Change the type contract of the J constructor to account for new structures found in Arts & Science website.
BEFORE:
update_j_constructor_after
AFTER:
update_j_constructor

- Refactor singleParser into singleParser for parsing strictly single courses and justParser for parsing text into updated J Req instances.
BEFORE:
singleparser_before
AFTER:
singleparser_after
justparser_after

- Update MinFan's graph generation functions to properly match updated J Reqs and add TODOs where implementations of functions need to be updated.
- Update test cases according to new J Req
- Add more complicated test cases from Christine's output and Arts & Science website.


app/WebParsing/ReqParser.hs
```diff
@@ -135,14 +135,47 @@ singleParser = do
...
+    return $ J (code ++ num ++ sess) ""
+
+-- | Parser for single courses or "atomic" Reqs represented by a J.
+justParser :: Parser Req
+justParser = do
+    code <- Parsec.count 3 Parsec.letter
+    num <- Parsec.count 3 Parsec.digit
+    sess <- Parsec.count 2 Parsec.alphaNum
+    Parsec.spaces
+    lpar <- Parsec.option "" $ Parsec.string "("
```
david-yz-liu: This can be improved if you use the `between` combinator. Something like

`meta <- Parsec.option "" $ Parsec.between lParen rParen markInfoParser`

app/WebParsing/ReqParser.hs
```diff
@@ -135,14 +135,47 @@ singleParser = do
...
+    case tmp of
+        Left mark -> return $ GRADE mark $ J (code ++ num ++ sess) ""
+        Right info -> return $ J (code ++ num ++ sess) info
+    where
+    markInfoParser = do
+        grade <- Parsec.try (percentParser <|> letterParser <|> infoParser)
+        _ <- Parsec.string ")"
+        return grade
+            where
+            percentParser = do
```
david-yz-liu: These are now duplicated from above; so you should make the top-level helper functions.

david-yz-liu:
@RyanDJLee this is good; because I merged in the latest changes from Dyanamic graphs, though, it would be great if you could do another pull and update the code again. (I think you'll have to redo some of your work for that, unfortunately.)

PR #1024, opened 2018-07-09
react-grid: import error when importing Modal from react-modal into course_panel

bettyruntianwang:
This error occurs when importing Modal from react_modal into course_panel
screenshot from 2018-07-09 00-38-20

screenshot from 2018-07-09 11-38-42

In js/common/react_modal.js:
`import * as ReactModal from 'vendor/react-modal'` 
results in an error: Cannot find module "vendor/react-modal". So it is changed to:
`import * as ReactModal from '../vendor/react-modal'`


david-yz-liu:
Thanks, @bettyruntianwang. I pushed some changes so now you should be able to import the modal.

PR #1026, opened 2018-07-31
react-grid: change representation of lecture times in Tables.hs to have more resemblance to the format needed to render the times to the timetable

bettyruntianwang:
A list of Times with the format [day, time], where time is every half hour that a given lecture runs is no longer being generated. The Time data type is now a record with fields weekDay, startHour and endHour.

Because of this, the code to parse lecture times to obtain lecture start and end times after fetching the course data is removed since it is no longer necessary. 
Also removed the creation of Course Objects (has the lectures attribute with a list of lecture objects) when a lecture is selected. Now, just generate the list of lecture objects after in lecture is selected (happens in component Row) without generating the Course object.

TimetableRow is modified to allow for lectures that begin and/or end on a half hour. 

beginendhalf

### DIFF
--- a/js/components/grid/calendar.js.jsx
+++ b/js/components/grid/calendar.js.jsx
@@ -8,10 +8,10 @@ export class Row extends React.Component {
   render() {
     // From a list of Course objects, create a list of Lecture objects
     let courses = this.props.courses;
-    let lectures = courses.map(
-      c => Array.prototype.concat.apply([], Object.values(c.lectures))
-    );
-    lectures = Array.prototype.concat.apply([], lectures);
+    let lectures = [];
+    courses.forEach(course => {
+      lectures = lectures.concat(createNewCourse(course))
+    })
 
     // Organize the structure of the <fallSession> and <springSession> 2-D dictionaries
     let fallSession, springSession;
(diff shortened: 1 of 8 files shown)


js/components/grid/calendar.js.jsx
```diff
@@ -120,23 +120,17 @@ class TimetableBody extends React.Component {
...
-    for (let i = 8; i < 22; i++) {
+    for (let i = 8; i < 22; i+=0.5) {
       rows.push(
         <TimetableRow
           session={this.props.session}
           time={i}
           key={'timetable-row-' + i + this.props.session}
           currentLectures={this.props.lectures[i]}
-          previousLectures={this.props.lectures[i-1]}
+          previousLectures={this.props.lectures[i-0.5]}
```
david-yz-liu: This type of change isn't good; you shouldn't try to index into a list using a fractional value. You need to come up with a formula that maps position in this array (or other arrays) with time of day.

js/components/grid/calendar.js.jsx
```diff
@@ -411,3 +406,28 @@ function storeWidths(session) {
...
+ * Create a list of lecture objects representing each separate time a given lecture section takes place
+ * @param {Object} lectureSection : Represents a lecture section for a course with time periods in which the lecture
+ *                            takes place
+ * @return {array} An array of objects representing each disconnected time period the given
+ *                  lecture occurs in
+*/
+function createNewLectures(lectureSection) {
+    let lectures = [];
+    let days = {0: 'M', 1: 'T', 2: 'W', 3: 'R', 4: 'F'};
+    for (let i = 0; i < lectureSection.times.length; i++) {
```
david-yz-liu: Use `lectureSection.map`

js/components/grid/calendar.js.jsx
david-yz-liu: Hmmm this is okay now then, but still not ideal.

PR #1044, opened 2019-01-21
Graph's components are refactored to have their own file

Fullchee:
I decided to keep renderReactGraph in its own file because
1. it should belong in the graph folder
2. draw's main depends on renderReactGraph

Force-pushed the changes so that the pull request is cleaner


Fullchee:
Force-pushed the changes so that the pull request is cleaner

PR #1046, opened 2019-01-25
Change version from 0.2.0.0 to 0.2.0.

Fullchee:
The npm command can now be used and will no longer complain about.

This will be useful for future students who have a habit of typing "npm"
rather than yarn.

We can also start following semantic versioning if we choose to do so.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "courseography",
-  "version": "0.2.0.0",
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@Fullchee thanks for doing this! I merged it in, but next time please remember to create a separate branch on your fork first.

PR #1047, opened 2019-01-25
Implement hovering feature for lecture sections.

RyanDJLee:


js/components/grid/course_panel.js.jsx
```diff
@@ -257,14 +275,27 @@ class LectureSection extends React.Component {
     return sameLecture.length > 0;
   }
 
+  // Display lecture section currently being hovered over.
+  hoverLecture() {
+    if (!this.clicked && !this.isSelectedLecture()) {
```
david-yz-liu: I've prefer to see this logic directly in the main `Grid` component, since it ultimately has to do with how the different parts of its state interact. Indeed, if you make this change then you won't need to define `hoverLecture` and `unhuverLecture` at all, but instead below pass in lambdas.

PR #1049, opened 2019-02-07
WIP: Snapshot tests

Fullchee:
Here are some of the preliminary snapshot tests.

I console.logged every component's props and copied it into its test file.

I've also removed Graph testing because Graph needs a lot of source code changes to render with the Jest (the test runner)


Fullchee:
Used array spread operator instead of Object.assign(), 

**Why?**
Spread operators will be in the ECMAScript 6 Standard.

(search for Rest/Spread Properties here: https://github.com/tc39/proposals/blob/master/finished-proposals.md)

It can be transpiled by Babel.

They are also less verbose.

Fullchee:
I've intentionally not included a snapshot for the `Graph` component because I will address that with the `integration-tests` pull request.

For example, I have had to copy and paste getURLParameter() into `js/components/graph/Graph.js`

This PR will focus on just adding snapshot tests for every other component.

PR #1052, opened 2019-02-16
Make the meetingId attribute in Times required

bettyruntianwang:
Update the insert and query functions to make the meetingId attribute in the Times table a required attribute.

To do this, the data type Time was added, its attributes mirror the Times attributes, but it does not have the meetingId attribute. Times is now only used as a database table; when working with meeting time related data, Time is used instead of Times.

### DIFF
--- a/app/Database/CourseQueries.hs
+++ b/app/Database/CourseQueries.hs
@@ -17,7 +17,8 @@ module Database.CourseQueries
      deptList,
      returnMeeting,
      getGraphJSON,
-     getMeetingTime) where
+     getMeetingTime,
+     buildTime) where
 
 import Happstack.Server.SimpleHTTP
 import Database.Persist
@@ -34,13 +35,6 @@ import Svg.Builder
 import Data.Maybe
 
 -- | Queries the database for all matching lectures, tutorials,
---   or praticals of this course and construct MeetTimes with the
---   Times corresponding to each lecture, tutorial or practial.
--- meetingQuery :: T.Text -> SqlPersistM [MeetTimes]
--- meetingQuery meetingCode_ = do
---     allMeetings <- selectList [MeetingCode ==. meetingCode_] []
---     mapM buildMeetTimes allMeetings
-
 meetingQuery :: T.Text -> SqlPersistM [MeetTime]
 meetingQuery meetingCode_ = do
     allMeetings <- selectList [MeetingCode ==. meetingCode_] []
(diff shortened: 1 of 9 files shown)


app/Database/CourseQueries.hs
```diff
@@ -75,9 +75,18 @@ returnMeeting lowerStr sect session = do
                                  []
     return $ head entityMeetings
 
+-- | Convert Times into Time
+buildTime :: Times -> SqlPersistM Time
```
david-yz-liu: 1. Make this function have type signature `Times -> Time`, and move it to `Tables.hs`.
2. Make a function that converts the other way, e.g. `buildTimes :: Key Meeting -> Time -> Times`.

(In general, prefer pure functions; also, functions that convert between data types should be located where those data types are defined.)

app/Database/CourseQueries.hs
```diff
@@ -109,11 +118,12 @@ getDescriptionD (Just key) = do
     maybeDistribution <- get key
     return $ fmap distributionDescription maybeDistribution
 
--- | Queries the database for all times corresponding to a given meeting.
-buildMeetTimes :: Entity Meeting -> SqlPersistM Tables.MeetTimes
+-- | Queries the database for all times corresponding to a given meeting and construct a MeetTime
```
david-yz-liu: This documentation change isn't necessary. We rely on the type signature to indicate that a `MeetTime` is returned.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -47,30 +46,30 @@ insertAllMeetings org = do
...
     -- Check that the meeting belongs to a course that exists
-    let code = meetingCode meetData
+    let code = meetingCode meetingData
     courseKey <- selectFirst [ CoursesCode ==. code ] []
     case courseKey of
         Just _ -> do
-          meetingKey <- insert meetData
-          mapM_ (\t -> insert_ $ t {timesMeeting = Just meetingKey}) meetTimes
+          meetingKey <- insert meetingData
+          mapM_ (\t -> insert_ $ Times (weekDay t) (startingTime t) (endingTime t) meetingKey (fstRoom t) (secRoom t)) meetingTime
```
david-yz-liu: Use the conversion function I suggested above to simplify this.

david-yz-liu:
Squashed in 389114f8c9b758d117e9ac379dd9087d88ba72d2.

PR #1054, opened 2019-02-26
Graph can now be initialized with props

Fullchee:


Fullchee:
### Why this PR exists
Normally, the graph is created from data from the server (with the get-json-data call)

We have 2 options
1. Find some other way to initialize the graph (like in this PR)
2. Mock the HTTP request with a new dependency `nock` (or some other way)
  * then the `fetch` call will return some data from `nock`

Fullchee:
I feel like Option 2 is better since we don't have to make an unnecessary change to Graph.

Fullchee:
Closing PR.

Mocking the HTTP Request doesn't need a change to Graph.js

PR #1058, opened 2019-03-01
STUCK: Async Issues

Fullchee:
**Problem**
1. The tests continue to run before the fetch() logic is complete
    * the tests might run on an uninitialized Graph

**Things I've Tried**
1. Delay the first test from running
    * calls graph.update() twice before the first test runs
    * works on my machine, may not on a faster computer
    * in the Pull Request (`Graph.test.js`)

2. Modify `Graph.js` so that `componentDidMount()` returns the fetch() promise
    * I didn't add this to the PR because it doesn't seem to work


Fullchee:
This breaks something (when refreshing, it encounters an error, not critical for presentation on Tuesday)

PR #1060, opened 2019-03-08
Mock get-json-data server API call

Fullchee:
I used the library `async-wait-until` to wait until the graph's render

Graph.js now remains completely unchanged.

I've also copy-pasted Ryan's react-testing-library tests from Tuesday's demo.


Fullchee:
Here's the fake data that I've created.

In order to have an OR node, I invented a CSC111.

I'll add this to the front-end-tests documentation
image

Fullchee:
image

Updated the data to have fake courses (existing course codes could change, resulting in confusion)

Fullchee:
image

Updated the picture so that the tests are more clear.
There's no duplicate "AAA101" and some code to search for the right "AAA101"

Fullchee:
Another update: The previous image doesn't work because a hybrid has to have a portion of the node that it refers to.

image

PR #1061, opened 2019-03-13
react-grid: Add Buildings table to database

bettyruntianwang:
Add Buildings table to database and parse the html uoft buildings list to insert the buildings and their locations into the table: 

http://map.utoronto.ca/c/buildings

Also removed manual practical and tutorial enrolment from Courses and removed unused Room data type.

Note: Will change data type for postal code in Buildings table to be Maybe T.Text as postal code is not available for some buildings


david-yz-liu:
Squashed in a1fe71c312894d4affadb3dc465d7a624cd61643.

PR #1070, opened 2019-05-17
Update node after mounting to initialize CSS classes

Fullchee:
I've added a componentDidMount() so that the CSS classes are initialized.

I've uploaded a before and after that highlights the difference (of the Chrome Dev Tools)
Before
before-componentDidMount

After
componentDidMount


Fullchee:
Only adding the `componentDidMount()` resulted in the old snapshot tests failing so I have added my Node tests to this Pull Request.

PR #1073, opened 2019-05-18
Node mount update

Fullchee:
Summary
1. Added `componentDidMount()` to `Node.js`
2. Modify the snapshot tests
	* necessary to pass the tests because Nodes now have another CSS class
	* I removed the hybrid node snapshot to make this PR simpler

I do have an updated hybrid node snapshot test and I will add it in with the rest of the new Node tests.
It will be the basis of my presentation next week.


Fullchee:
Why does this PR Exist?

The Node isn't initialized with all of its CSS classes.

We want to go from
image

to this:

image

PR #1074, opened 2019-05-18
Convert svg text parser to Parsec for coordinates and size with units

pengmai:
Converted some of the low-level text parsers to use parser combinators via Parsec. This will allow `width` and `height` attributes to accept units (which will currently be ignored).

### DIFF
--- a/app/Svg/Parser.hs
+++ b/app/Svg/Parser.hs
@@ -293,11 +293,10 @@ readAttr attr tag = fromMaybe
 -- Throws an exception on any parse errors.
 parseVal :: Parser a -> T.Text -> a
 parseVal parser input =
-    case P.parse parser "" strippedInput of
+    case P.parse parser "" $ T.unpack input of
         Left err ->
             error ("reading " ++ T.unpack input ++ ":" ++ show err)
         Right val -> val
-    where strippedInput = filter (not . isSpace) (T.unpack input)
 
 -- | Looks up an attribute value using the given parser.
 parseAttr :: Parser a -> T.Text     -- ^ The attribute's name.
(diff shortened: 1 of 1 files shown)


app/Svg/Parser.hs
```diff
@@ -286,6 +289,39 @@ readAttr attr tag = fromMaybe
...
 
+-- | Runs a parser on a text object.
+-- Throws an exception on any parse errors.
+parseVal :: Parser a -> T.Text -> a
+parseVal parser input =
+    case P.parse parser "" strippedInput of
+        Left err ->
+            error ("reading " ++ T.unpack input ++ ":" ++ show err)
+        Right val -> val
+    where strippedInput = filter (not . isSpace) (T.unpack input)
```
david-yz-liu: Don't strip out whitespace here. Instead, modify your parsers to allow optional whitespace (`Text.Parsec.Char` has a `spaces` parser) between the key elements.

app/Svg/Parser.hs
```diff
@@ -286,6 +289,39 @@ readAttr attr tag = fromMaybe
...
+parseAttr parser attr tag = parseVal parser $ fromAttrib attr tag
+
+-- | Parses one or more digit characters.
+digits :: Parser String
+digits = P.many1 P.digit
+
+-- | Parses a double value, ignoring units if present.
+double :: Parser Double
+double = do
+    sign <- P.option "" (P.string "-")
```
david-yz-liu: Use `char` from `Text.Parsec.Char` for single characters (same with the comma below).

pengmai:
Superceded by #1075

PR #1076, opened 2019-05-24
Fix bug in Node constructor

Fullchee:
The CSS classes of Nodes (both types: hybrid and node) weren't being initialized properly.

**Before**
image1

**After**
image2


david-yz-liu:
Thanks, @Fullchee!

PR #1082, opened 2019-05-31
Add libraries leaflet and react-leaflet, and a Map Modal component for the grid page

bettyruntianwang:
The map contained in the added map modal is currently a placeholder.

Note that in order for the leaflet map to be rendered properly, the following must be in place (otherwise the map tiles are scrambled/missing):

1. The leaflet css must be included in the header
2. The leaflet js file must be included after the leaflet css in the header
    (see [https://leafletjs.com/examples/quick-start/])
3. The map container (leaflet-container) must have a defined height. Define a height for ".leaflet-container" in the css.
4. The added loaders in the webpack configurations are needed: file-loader (url-loader will also work instead of file-loader) and style-loader!css-loader
5. The original marker icon image path provided by leaflet can't be found. A custom marker icon is created in the map modal component to replace the original.


js/components/common/react_modal.js.jsx
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import ReactModal from 'react-modal';
-
+import Leaflet from 'leaflet';
+import {CircleMarker, Polygon, Polyline, Map, TileLayer, Marker, Popup, Tooltip, Circle } from "react-leaflet";
```
david-yz-liu: Using single-quotes here.

js/components/common/react_modal.js.jsx
```diff
@@ -1,6 +1,9 @@
...
+import L from 'leaflet'
+import 'leaflet/dist/leaflet.css';
```
david-yz-liu: Remove the css import.

js/components/grid/calendar.js.jsx
```diff
@@ -89,7 +100,11 @@ class TimetableHeader extends React.Component {
         <thead>
           <tr>
             <th className="timetable-dummy-cell"></th>
-            <th className="term-name">Fall</th>
+            <th className="term-name">
+              <MapModal ref={ r => this.modal = r}/>
```
david-yz-liu: This is okay, but we should use the new form of refs now: https://reactjs.org/docs/refs-and-the-dom.html

js/components/grid/calendar.js.jsx
```diff
@@ -99,7 +114,11 @@ class TimetableHeader extends React.Component {
         <thead>
           <tr>
             {dayCells}
-            <th className="term-name">Spring</th>
+            <th className="term-name">
+              Spring
+              <MapModal ref={ r => this.modal = r}/>
```
david-yz-liu: Don't duplicate the `MapModal`. Instead, create just one, and eventually pass in as props the necessary data. Right now, you don't need any props because there's no real data, but you could change the `displayMap` method to take an a semester parameter (e.g., `() => this.displayMap('F')`).

webpack.config.js
```diff
@@ -16,7 +16,9 @@ module.exports = {
   module: {
     loaders: [
       { test: /\.js$/, loader: 'babel-loader', exclude: /node_modules/ },
-      { test: /\.jsx$/, loader: 'babel-loader', exclude: /node_modules/ }
+      { test: /\.jsx$/, loader: 'babel-loader', exclude: /node_modules/ },
+      { test: /\.(gif|svg|jpg|png)$/, loader: 'file-loader' },
+      { test: /\.css$/, loader: "style-loader!css-loader" }
```
david-yz-liu: Please remove this part for now.

PR #1083, opened 2019-05-31
SVG Parser: path "d" attributes

pengmai:
As of these changes, the parser is able to successfully parse the dynamically generated sample graph. A/B Testing was performed via database dumps with all of the static graphs to ensure that the resulting path values were identical with the new path parser.

Additionally, one of the static graphs contained a single value in scientific notation, so this PR includes support for parsing those values.

### DIFF
--- a/app/Svg/Parser.hs
+++ b/app/Svg/Parser.hs
@@ -318,23 +318,29 @@ parseAttr parser attr tag = parseVal parser $ fromAttrib attr tag
 digits :: StatefulParser u String
 digits = P.many1 P.digit
 
--- | Parses a double value, ignoring units if present.
+-- | Parses a double literal, optionally in scientific notation.
 double :: StatefulParser u Double
 double = do
     sign <- P.option ' ' (P.char '-')
     whole <- digits
     fractional <- P.option ".0" parseFractional
-    return (read $ sign : whole ++ fractional)
+    magnitude <- P.option "e0" parseMagnitude
+    return (read $ sign : whole ++ fractional ++ magnitude)
     where
         parseFractional = do
             _ <- P.char '.'
             decimals <- digits
             return ("." ++ decimals)
+        parseMagnitude = do
+            ch <- P.char 'e'
+            sign <- P.option "" $ P.string "-"
+            power <- digits
+            return (ch : sign ++ power)
 
 point :: StatefulParser u Point
 point = do
     xPos <- double
-    _ <- P.char ',' <|> P.space
+    _ <- P.char ','
     yPos <- double
     return (xPos, yPos)
 
@@ -374,7 +380,10 @@ parseTransform transform =
         rotate = P.string "rotate(" >> double >> P.char ')' >> P.spaces
         translate = do
             _ <- P.string "translate("
-            point
+            xPos <- double
+            _ <- P.char ',' <|> P.char ' '
+            yPos <- double
+            return (xPos, yPos)
 
 parseCoord :: T.Text -> Point
 parseCoord "" = (0, 0)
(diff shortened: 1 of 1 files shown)


app/Svg/Parser.hs
```diff
@@ -344,82 +367,106 @@ getTransform :: Tag T.Text -> Point
...
--- | Parses a translation String into a tuple of Float.
+-- | Parses a translation String into a tuple of Float, ignoring scaling and
+-- rotation.
 parseTransform :: T.Text -> Point
 parseTransform "" = (0, 0)
 parseTransform transform =
     parseVal parser transform
     where
-        parser = do
+        parser = P.many (scale <|> rotate) >> translate
```
david-yz-liu: Use `P.sepBy` to move the `P.spaces` out of the definition of scale/rotate.

app/Svg/Parser.hs
```diff
@@ -344,82 +367,106 @@ getTransform :: Tag T.Text -> Point
...
-      convertToPoint (x, y) =
-        case (readMaybe (T.unpack x), readMaybe (T.unpack y)) of
-            (Just m, Just n) -> Just (m, n)
-            _ -> Nothing
+        -- Keep track of the (first point, path mode, most recent point)
+        initialState = PathDState { firstPoint = (0, 0)
+                                  , mode = AbsoluteMove
+                                  , currentPoint = (0, 0) }
+
+        parser = do
```
david-yz-liu: Include type signatures for these helpers.

app/Svg/Parser.hs
```diff
@@ -344,82 +367,106 @@ getTransform :: Tag T.Text -> Point
...
+            p <- parseStep
+            -- Record the first point added.
+            _ <- P.modifyState $ \st -> st { firstPoint = p, currentPoint = p }
+            rest <- P.many parseStep
+            return (p : rest)
+        parseStep = do
+            _ <- (P.option () $ bezier <|>
+                absoluteMove <|> relativeMove <|>
+                absHorizontalMove <|> absVerticalMove <|>
+                relHorizontalMove <|> relVerticalMove) >> P.spaces
```
david-yz-liu: In general, put the `P.spaces` on a new line (`_ <- P.spaces`).

app/Svg/Parser.hs
```diff
@@ -344,82 +367,106 @@ getTransform :: Tag T.Text -> Point
...
+                                  , currentPoint = (0, 0) }
+
+        parser = do
+            p <- parseStep
+            -- Record the first point added.
+            _ <- P.modifyState $ \st -> st { firstPoint = p, currentPoint = p }
+            rest <- P.many parseStep
+            return (p : rest)
+        parseStep = do
+            _ <- (P.option () $ bezier <|>
```
david-yz-liu: Use `P.choice` to simplify the `<|>`s.

app/Svg/Parser.hs
```diff
@@ -344,82 +367,106 @@ getTransform :: Tag T.Text -> Point
...
+                absoluteMove <|> relativeMove <|>
+                absHorizontalMove <|> absVerticalMove <|>
+                relHorizontalMove <|> relVerticalMove) >> P.spaces
+            p <- closeLoop <|> moveArg
+            state <- P.getState
+            let (currentX, currentY) = currentPoint state
```
david-yz-liu: I'd like you to move this part into a separate helper of type `Point -> PathDState -> PathDState ` (extract the relevant parts of the current state in the helper). Then you can do `let p' = (yourHelper p)` etc. in the main parser. Also see my above comment about `P.spaces`.

app/Svg/Parser.hs
```diff
@@ -26,13 +26,18 @@ import Database.DataType
...
 import Text.Read (readMaybe)
 import Data.Char (isSpace)
 import qualified Data.Text as T
 import Data.Text.IO as T (readFile)
 import Data.List.Split (splitOn)
+import Data.Functor.Identity (Identity)
 
+-- A Parsec parser with user defined state. This can be made stateless by
+-- assigning () to u.
+type StatefulParser u a = P.ParsecT String u Identity a
```
david-yz-liu: I see what you're trying to do here, but there's the `P.Parsec` type synonym already that we should just use. You can turn the `Parser a` uses below into `P.Parsec String u a`.

app/Svg/Parser.hs
pengmai: Right, that was an attempt to save lines that resulted in making code less readable in a bunch of cases. Does that also apply to instances like `bezier`, where everything ends up getting discarded?

PR #1086, opened 2019-06-07
Add fake aaa course info for AAA100.

Fullchee:
Why this PR
-------------
1. Uses the more recent format (has `allMeetingTimes`) so it works as a mock.
2. Replace the description to custom AAA100 content.
3. more readable (instead of all on one line)  and helps with future diffs.


Fullchee:
Currently, this course info is not used but it will be used for InfoBox tests (tests where you click on "Info")

PR #1097, opened 2019-06-23
Resolve issue #1085 of ternary returning the same value

### LINKED ISSUE #1085: Bool arePrereqsSatisfied() ternary returns the same value
https://github.com/Courseography/courseography/blob/b561190b53b6484f2d38afa233d47d1d59a8bcf6/js/components/graph/Bool.js#L23-L25

Fullchee:
Resolves #1085 

I reverted the behaviour to follow the logic from three years ago from the following commit

7d0bc

Here's the exact snippet.
https://github.com/courseography/courseography/blob/7d0bc75572ed3a43062f8cf01fc41e3a32939369/public/js/graph/react_node.js#L796-L797

### DIFF
--- a/js/components/graph/Bool.js
+++ b/js/components/graph/Bool.js
@@ -22,7 +22,7 @@ export default class Bool extends React.Component {
     function isAllTrue(element) {
       return svg.nodes.current[element]
         ? svg.nodes.current[element].isSelected()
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@Fullchee nice catch, thank you!

PR #1098, opened 2019-06-23
Remove unused util functions

Fullchee:
Why this PR
--------------
Removed all functions in `util.js` that aren't used anywhere in the repo.

Next Steps/Where I'm stuck
-------------------------
There are two files that are also in the `public` folder which use the `util.js`
- [course.js](https://github.com/Courseography/courseography/blob/master/public/js/common/objects/course.js)
- [section.js](https://github.com/Courseography/courseography/tree/master/public/js/common/objects/section.js)

~~The only place where I see any of these files being used is~~
- ~~the method `parseLectures` in [course_panel.js.jsx](https://github.com/Courseography/courseography/blob/master/js/components/grid/course_panel.js.jsx) in the grid component.~~

Edit: parseLectures is defined in course_panel.js.jsx.
* it doesn't use the parseLectures in `course.js`


Fullchee:
*Scripts.hs*
* Remove the requests for files that no longer exist

*react_modal.js.jsx* and *course_panel.js.jsx*
* add the explicit import

Add the getCourse to utils.js

Remove course.js, section.js and util.js

Issue #1087, opened 2019-06-08
Front End Testing: Variables are inserted into the global scope


Fullchee:
Steps to Reproduce
-------------------------
Add the following code in a test file (eg: `courseography/js/components/graph/__tests__/Isolated.test.js`)
There is no error despite how aaa100 is never explicitly defined.

```js
import React from "react";
import setupGraph from "./setupGraph";
test("shouldnt be able to access old variables", async () => {
  const graph = await setupGraph();
   console.log(aaa100);
});
```
Summary
--------------------------------
* Debugging the test shows that all Nodes and Booleans are added to the global scope.
* A search of the repo shows that nothing is adding items to the `global` object (even in react-testing-library)

I added a new node to `js/components/graph/__mocks__/testData.js` and it is not defined anywhere else.
However, it is still added to the global scope.
![image](https://user-images.githubusercontent.com/11246258/59151411-a4248280-8a00-11e9-925e-cfaaf992028f.png)

Next Steps
-----------------------
- [x] how to set proper break points in VSCode that work with async/await
   * in order to further narrow down the issue
- [x] create an issue on `react-testing-library` GitHub page.

Fullchee:
just used the `debugger` statement

between (inclusive) calling `this.setState` in `Graph.js` and the `componentDidUpdate`, the global variables are added

Fullchee:
Found out why this happens.

the `this` seems to be referring to the global scope which is why all the Nodes are added to the global scope.

https://github.com/Courseography/courseography/blob/efb4343dbb4440cb7e543186d60d414c78a3f4af/js/components/graph/NodeGroup.js#L99

There is also something similar for Bool nodes as well
https://github.com/Courseography/courseography/blob/efb4343dbb4440cb7e543186d60d414c78a3f4af/js/components/graph/BoolGroup.js#L46

PR #1099, opened 2019-07-03
Bind ref callbacks

Fullchee:


Fullchee:
Why this PR
-------------------
The official docs on the callback-refs state that
https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs

> If the ref callback is defined as an inline function, it will get called twice during updates

After extracting the inline functions into methods, the global variables are not created anymore.

Resolves #1087

PR #1104, opened 2019-07-16
Infobox tests

Fullchee:


Fullchee:
Removed the `react-testing-library` update.

PR #1105, opened 2019-07-20
Prerequisite graph generation

pengmai:


app/Database/CourseQueries.hs
```diff
@@ -178,6 +179,16 @@ allCourses = do
       return $ T.unlines codes :: SqlPersistM T.Text
   return $ toResponse response
 
+-- | Retrieves the prerequisites for a course (code) as a string.
+prereqsForCourse :: T.Text -> IO (Either String T.Text)
+prereqsForCourse course = runSqlite databasePath $ do
+    courses :: [Entity Courses] <- selectList [CoursesCode ==. course] []
```
david-yz-liu: Use `selectFirst` instead of `selectList` when you only want one record. (It returns a `Maybe` instead of a list.)

app/DynamicGraphs/CourseFinder.hs
```diff
@@ -0,0 +1,41 @@
...
+    Description : Retrieve course information needed to generate graphs.
+
+This module contains the logic requests for information needed to
+generate graphs before retrieving and structuring the necessary data.
+-}
+module DynamicGraphs.CourseFinder (lookupCourses) where
+
+import qualified Data.Text.Lazy as T
+import qualified Data.Map.Strict as Map
+import Control.Monad.Trans.State
```
david-yz-liu: It would be good to explicitly list what you're importing here (`StateT`, `modify`, etc.?)

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -88,25 +101,31 @@ reqToStmts' parentID (RAW rawText) = do
...
+                nodesMap' = Map.insert name node nodesMap
+            State.put (GeneratorState (i + 1) nodesMap')
+            return node
+        Just node -> return node
+
+makeBool :: Text -> State GeneratorState (DotNode Text)
 makeBool text1 = do
-    i <- State.get
-    _ <- State.put (i + 1)
+    GeneratorState i nodesMap <- State.get
```
david-yz-liu: You can use `State.modify` here as well.

app/Main.hs
```diff
@@ -22,31 +22,30 @@ import Svg.Parser (parsePrebuiltSvgs)
...
+    ("docs", const generateDocs),
+    ("generate", \(name : courses) -> generatePrereqsForCourses (name, courses))]
 
 -- | Courseography entry point.
 main :: IO ()
 main = do
     args <- getArgs
     let taskName = if null args then "server" else head args
-    fromMaybe putUsage (Map.lookup taskName taskMap)
+    fromMaybe putUsage (Map.lookup taskName taskMap) $ tail args
```
david-yz-liu: I think the `tail` here fails if `args` is empty? I'm a little surprised you didn't run into this, so maybe I'm mistaken. But if I'm right you could pattern-match on `case args of ...`

app/DynamicGraphs/GraphGenerator.hs
pengmai: `i` actually gets used below on line 121, so I believe I do need to explicitly access the state before updating it.

Issue #656, opened 2015-06-30
Add JavaScript Tests


Ian-Stewart-Binks:

Fullchee:
Added test cases! (with more to come)

Fullchee:
@david-yz-liu Should we mark this issue as closed?

PR #1108, opened 2019-07-29
Add cypress.io

Fullchee:
Key Differences of Cypress vs Selenium
https://docs.cypress.io/guides/overview/key-differences.html

Overview of Cypress vs Selenium
https://applitools.com/blog/cypress-vs-selenium-webdriver-better-or-just-different


Fullchee:
Removed the unnecessary demo command
`cy.contains("aslkjdflksd")`

cypress/integration/sidebar_spec.js
```diff
@@ -0,0 +1,17 @@
...
+  it("Switch to Aboriginal Studies", () => {
+    cy.get("#sidebar-button").click();
+    cy.contains("Aboriginal").click();
+    cy.get("#sidebar-button").click();
+  });
+  it("navigation buttons", () => {
+    cy.get("#zoom-in-button")
+      .click()
+      .click();
+    cy.get("#pan-right-button").click();
```
david-yz-liu: This is a bit funky. If we do `cypress run` (headless mode) this test fails because the button is disabled. Could be because of the width/height dimensions of the SVG vs. the window; not sure how that works in this context. For now you could just disable this check, or you could look into it.

cypress/integration/grid_spec.js
```diff
@@ -0,0 +1,6 @@
+describe("Visit website", function() {
+    it("Find Courses", function() {
+      cy.visit("localhost:8000/grid");
```
david-yz-liu: You can just do `/grid` right?

cypress/integration/timesearch_spec.js
```diff
@@ -0,0 +1,8 @@
+describe("timesearch", () => {
+  it("should be able to finds stuff", () => {
```
david-yz-liu: You should reword this description.

Fullchee:
Still not ready for a PR yet

Fullchee:
Why we can't use any helper functions
https://docs.cypress.io/guides/references/best-practices.html#Assigning-Return-Values

Fullchee:
Resolves #656

PR #1111, opened 2019-08-02
Add VSCode Launch.json

Fullchee:
Why this PR
-----------------
I'm writing the docs and I want instructions to be as helpful as possible if the tests fail.

Setting up the VSCode debugger settings was non-trivial so I'm adding the config.
https://github.com/facebook/create-react-app/issues/5319

The configs might also change in the future so it will be easier to make changes apparent to everyone when it's already in the repo.

New contributors will automatically have this config in the VSCode debugger panel.
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ cabal.sandbox.config
 
 # IDE project files
 .idea
-.vscode
 
 # Node (Javascript) modules
 /node_modules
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@Fullchee I don't want to put this in the repo because I prefer keeping the project editor-agnostic, but this configuration is good to have in the wiki (and I think you've already put it in the wiki).

PR #1112, opened 2019-08-03
Link to JS instead of CSS file

Fullchee:
Why this PR
-------------------
`Uncaught SyntaxError: Unexpected token {`

I get that error when I go to the grid page
https://courseography.cs.toronto.edu/grid

It's trying to read JS from a CSS file.

CDN link from here: https://blog.getbootstrap.com/2014/02/13/bootstrap-3-1-1-released/


david-yz-liu:
Great catch!

PR #1143, opened 2020-09-22
Refactored Sidebar from Haskell to a React Component

ashih2018:
## Motivation and Context
This PR removes the previous `Text.Blaze.Html5 Haskell` code that was written to render the sidebar HTML for Courseography and replaces it with React components.
To do this, @stefurac and I created a `Focus.js` and `Sidebar.js` to replace the Haskell code, as well as added a `Container.js` component which is a parent component of the `Sidebar.js` and `Graph.js` components.

## Your Changes
- Created `Container.js` component
- Created `Sidebar.js` component
- Created `Focus.js` component
- Reduced the code in `main.js`
- Removed `sidebar-divs.js` as the previously existing jQuery methods are now converted to React.
- Removed previously written Haskell code in Graph.hs to render the sidebar

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Test update (change that modifies or updates tests only)
- [ ] Other (please specify): 

#### Setup
- Run `stack build`
- Run `stack exec courseography css` to load new css
- Run `stack exec courseography`
- Navigate to `http://localhost:8000/graph`
#### Manual Testing
Do the following steps for manual testing. Keep the console open the entire time to ensure there are no errors.
1. Click the sidebar and ensure it opens correctly. All the graph buttons in the sidebar should load properly. Also ensure the arrow favicon flips.
2. Click a non-computer science graph. The graph should load properly and the focus button in the sidebar should not be clickable.
3. Refresh the page. The graph you were previously on should show up and the focuses tab should still be disabled.
4. Click the computer science graph. Click on a node. The sidebar should automatically close. Click on a couple more nodes. Then open the sidebar and click reset graph. This should reset the nodes and the FCE count as expected.
5. Click on Focuses. Click on a particular focus. The focuses should be highlighted in the graph and the focus details should expand and be readable. Navigate to different focuses and ensure the focuses in the graph change. Click the same focus to undo the focus effect.
6. Click the export button on the top. This should open and there should be a GET request for an image in the console. However this does not work locally for some reason (tested on master and same error occurs). Ensuring the GET request for the image is sent should be sufficient.

 - We should make sure that the export modal functionality works before deploying.
 - In order for the CSS changes to propagate, you may need to clear browser cache.

## Proposed Future Updates
 - When a focus is selected, we could disable the graphs navigation. Otherwise, the user is prone to clicking on another graph and the graph is unresponsive because the previous focus is still "focused".
 - When a focus is selected, clicking reset graph should unfocus.
 - The graph and focus buttons go off the sidebar "page"/div and into the disclaimer. This happens in the deployed courseography as well. This is especially obvious when I was working with the console up, but it isn't a huge deal.
 - Add animation to the sidebar opening to match previous animation.

## Checklist

- [ ] I have updated the Changelog.md file. 

### Required documentation changes (if applicable)
????

## TODO:
 - [x] Fix failing tests
 - [x] Update reset button to reset focuses
 - [x] Add tests for sidebar


app/Response/Graph.hs
```diff
@@ -16,42 +16,8 @@ graphResponse =
...
-
-sideBar :: H.Html
-sideBar = do
-    H.div ! A.id "sidebar" $ do
-        H.div ! A.id "fce" $ do
-            H.div ! A.id "fcecount" $ "FCE Count: 0.0"
-            H.button ! A.id "reset" $ "Reset Graph"
-        H.nav ! A.id "sidebar-nav" $ H.ul $ do
-            H.li ! A.id "graphs-nav" $
-                H.a ! A.href "" $ "Graphs"
```
ashih2018, line 33: Why is this an <a> tag?

david-yz-liu: I'm not sure if you're asking me, but I think it's because there's supposed to be an onclick callback here.

ashih2018: Right now this is causing the page to refresh when clicked on. Can I change this to a `div`?

js/components/graph/Focus.js
```diff
@@ -0,0 +1,40 @@
...
+    const detailsInfo = this.getDetailsInfo();
+    const divId = this.props.pId + '-details';
+    return (
+      <div onClick={() => this.props.highlightFocus(this.props.pId)}>
+        <p id={this.props.pId} className="focus">{this.props.focusName}</p>
+        <div 
+          id={divId}
+          className="details"
+          style={detailsInfo["detailsStyle"]}
+          dangerouslySetInnerHTML={{__html: detailsInfo["detailsText"]}}
```
ashih2018, line 28: If I don't use `dangerouslySetInnerHTML` then \<br> and \<em> appear in text. I tried using whitespace: `pre-wrap` with CSS but that didn't work, so for now I am using this.

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+
+  render() {
+    const contentHiddenClass = this.state.contentHidden ? "hidden" : "";
+    const focusDisabled = this.state.focusDisabled ? "disabled" : "";
+    const focusActiveClass = this.state.focusActive ? "active" : "";
+    const focusHiddenClass = !this.state.focusActive ? "hidden" : "";
+    const graphActiveClass = this.state.graphActive ? "active" : "";
+    const graphHiddenClass = !this.state.graphActive ? "hidden" : "";
+    const flippedClass = this.state.toggled ? "flip" : "";
+    const sidebarClass = this.state.toggled ? "opened" : "";
```
ashih2018, line 135: Is there a better way to set CSS for multiple elements?

js/components/graph/Container.js
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      graphName = localStorage.getItem("active-graph") || "Computer Science";
+    }
+    this.setState({ graphName: graphName });
+    return graphName;
+  }
+
+  highlightFocus = id => {
+    if (this.graph.current.state.highlightedNodes == focusInfo[id + "FocusList"]) {
+      this.graph.current.highlightFocuses([]);
```
david-yz-liu, line 60: Don't call this here; instead, just `setState` here, and pass that state as a prop to the `Graph` component and have that component call `highlightFocuses`.

js/components/graph/Container.js
```diff
@@ -0,0 +1,100 @@
...
+
+  render() {
+    return (
+      <div>
+        <Graph
+          ref={this.graph}
+          initialDrawMode="draw-node"
+          edit={this.props.edit}
+          start_blank={this.props.start_blank}
+          getLocalGraph={() => this.getLocalGraph()}
```
david-yz-liu, line 80: You can just do `this.getLocalGraph` here

js/components/graph/Container.js
```diff
@@ -0,0 +1,100 @@
...
+          closeSidebar={() => this.sidebar.current.toggleSidebar("graph")}
+        />
+        <Sidebar
+          ref={this.sidebar}
+          currFocus={this.state.currFocus}
+          getGraph={(name) => this.graph.current.getGraph(name)}
+          graphs={this.state.graphs}
+          graphName={this.state.graphName}
+          highlightFocus={(id) => this.highlightFocus(id)}
```
david-yz-liu, line 89: As above, you can pass `this.highlightFocus` here

js/components/graph/Container.js
```diff
@@ -0,0 +1,100 @@
...
+          reset={() => this.graph.current.reset()}
```
david-yz-liu, line 90: This can be `this.graph.current.reset`

js/components/graph/Focus.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import PropTypes from "prop-types";
+import * as focusInfo from "./sidebar/focus_descriptions.js";
+
+export default class Focus extends React.Component {
+  getDetailsInfo = () => {
+    let detailsStyle, detailsText;
+    if (this.props.openDetails) {
+      detailsStyle = { height:"128px" };
+      detailsText = focusInfo[this.props.pId + "Description"]
```
david-yz-liu, line 10: Insert a semicolon here

js/components/graph/Graph.js
```diff
@@ -668,7 +666,7 @@ export default class Graph extends React.Component {
     }
 
     return (
-      <div>
+      <div id="react-graph" className="react-graph" onClick={() => this.props.closeSidebar()}>
```
david-yz-liu, line 669: As above, you can set `onClick` to be `this.props.closeSidebar`

js/components/graph/Graph.js
```diff
@@ -93,19 +93,7 @@ export default class Graph extends React.Component {
...
-      // link to our graph.
-      // Should be replaced with a more general solution.
-      if (urlSpecifiedGraph === "sta") {
-        graphName = "Statistics";
-      } else if (urlSpecifiedGraph !== null) {
-        graphName = "Computer Science";
-      } else {
-        graphName = localStorage.getItem("active-graph") || "Computer Science";
-      }
+      graphName = this.props.getLocalGraph();
```
david-yz-liu, line 96: Rather than passing `getLocalGraph` down as a prop, pass the `graphName` itself down as a prop. You should be able to make `getGraph` a function that takes no arguments, and always retrieves `this.props.graphName`. Call this function in a `componentDidUpdate` when the prop changes.

(Also, when you're doing this, move the `replace` call on line 99 up to the `Container` class.

js/components/graph/Focus.js
```diff
@@ -0,0 +1,40 @@
...
+        />
+      </div>
+    )
+  }
+}
+
+Focus.propTypes = {
+  focusName: PropTypes.string,
+  highlightFocus: PropTypes.func,
+  openDetails: PropTypes.bool,
```
david-yz-liu, line 38: You can remove the "Details" part. Perhaps just `open` or `expanded`, for example.

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      contentHidden: true,
+      focusDisabled: false,
+      focusActive: false,
+      graphActive: false,
+      graphName: "",
+      toggled: false
+    }
```
david-yz-liu, line 15: Insert a semicolon here

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+    } else {
+      this.setState({
+        focusDisabled: true,
+      });
+    }
+  }
+
+  createGraphButtons = () => {
+    const graphButtons = [];
+    this.props.graphs.forEach((graph, i) => {
```
david-yz-liu, line 47: This function can be simplified by using `map` instead of `forEach`. (The code should look like `return this.props.graphs.map(...)`)

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+      ["NLP", "Natural Language Processing"],
+      ["vision", "Computer Vision"],
+      ["systems", "Computer Systems"],
+      ["game", "Video Games"],
+      ["HCI", "Human Computer Interaction"],
+      ["theory", "Theory of Computation"],
+      ["web", "Web Technologies"],
+    ];
+    const focusComponents = [];
+    computerScienceFocusData.forEach((focus, i) => {
```
david-yz-liu, line 77: See comment about `map` vs `forEach` above.

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+      focusActive: false,
+      graphActive: false,
+      graphName: "",
+      toggled: false
+    }
+  }
+
+  componentWillUpdate(prevProps) {
+    // Check to see if we have a graph from local storage on load
+    if (this.state.graphName.length == 0 && prevProps.graphName !== this.state.graphName) {
```
david-yz-liu, line 20: Always use `===` for value equality, not `==`. (Check this throughout these files.)

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+        graphActive: true,
+      });
+    }
+  }
+
+  showFocuses = focus => {
+    if (focus) {
+      // show focuses
+      this.setState({
+        focusActive: true,
```
david-yz-liu, line 115: It's redundant to have two separate state values that are negations of each other. Instead, have a single state that is an index (either 0 or 1) of which "tab" is open. This will also allow us to generalize to multiple tabs in the future, e.g. for the dynamic graph form.

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,182 @@
...
+    const graphButtons = [];
+    this.props.graphs.forEach((graph, i) => {
+      const graphId = "graph-" + graph.id;
+      const graphName = graph.title;
+      graphButtons.push(
+        <div
+          className="graph-button"
+          id={graphId}
+          key={i}
+          onClick={() => this.handleGraph(graphName)}
```
david-yz-liu, line 55: Right, so similar to my above comment on `getGraph`, this button should call `this.props.getGraph` directly. The `handleGraph` function isn't necessary (this should help simplify the code for this class substantially).

app/Css/Graph.hs
```diff
@@ -271,6 +271,10 @@ graphContainer = do
 
 sidebarCSS :: Css
 sidebarCSS = do
+    ".opened" ? do
+        "width" -: "400px !important"
```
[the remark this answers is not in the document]
ashih2018: This is because by default `#sidebar` has a width of 40px so we want to overwrite it when opened.

js/components/graph/Container.js
stefurac: This cannot be completed. It raises an error on the initial render because the ref we use to reference the graph component is null when this is first called.

js/components/graph/Sidebar.js
```diff
@@ -0,0 +1,178 @@
...
+    this.state = {
+      contentHidden: true,
+      focusDisabled: false,
+      graphActive: 0,
+      graphName: "",
+      toggled: false
+    }
+  }
+
+  componentWillUpdate(prevProps) {
```
ashih2018, line 17: Getting warnings that `componentWillUpdate` is deprecated. Will this be a problem?

js/components/graph/Graph.js
ashih2018: Had to call the function in `componentWillUpdate` because changes did not propogate.

js/components/graph/Sidebar.js
ashih2018: Used helper functions to break this apart.

js/components/graph/__tests__/Node.test.js
```diff
@@ -251,28 +251,5 @@ describe("Course Node", () => {
         expect(aaa303.classList.contains("missing")).toBe(true);
       });
     });
-
-    it("Selecting a course node should always increase the FCE count by 0.5, (currently no support for full-year courses)", async () => {
```
ashih2018, line 255: Added this test to `Sidebar.test.js`

js/components/graph/__tests__/TestGraph.js
```diff
@@ -21,7 +21,11 @@ export default class TestGraph {
       edit: false,
       initialDrawMode: "draw-node",
       initialOnDraw: false,
-      start_blank: false
+      start_blank: false,
+      currFocus: null,
+      graphName: "Computer Science",
+      setFCECount: () => {},
```
ashih2018, line 27: Is it okay if I "mock" these two props functions like so?

david-yz-liu (approved):

PR #1145, opened 2020-10-03
Update parsers for grade and allow parser to find utsc courses

### LINKED ISSUE #1147: Grade after course not found if grade is followed by other characters
For example: 

"CSC111Y1 (70% or higher)" would give RAW "CSC111Y1 (70% or higher)"

"CSC111Y1 (at least 70% or more)" would give J "CSC111Y1" "at least 70% or more"

"CSC263H1 with a minimum grade of 60% or more" would give J "CSC263H1" ""

bettyruntianwang:
When grade is before the course code, there are several cases the coBefParser can now handle, including:
- case insensitive parsing
- A or An at the start of the prerequisite sentence
- optional "of" following key words such as "minimum grade"
- new key words based on the artsci calendar such as "at least", "grade of", "final grade of".
- starting the prerequisite sentence with a grade: eg "60% in MAT137H1"

Note the there is one more case I have not yet handled in the format of "A in MAT137H1". This is specific to the letter grade A with no "-" or "+" following it (this is because I currently skip over the optional "A" at the beginning of the prereq sentence). There are no examples of this in the database, but it should be handled.

When grade is after the course code, we now handle the case with coAftParser where there are more words following the grade, for example:
- "CSC110Y1 (70% or higher)"
- "CSC263H1 with a minimum grade of 60% or more"

UTSC course code has format: 4 letters + 2 digits + H or Y + 1 digit
The parser will now be able to identify these as courses.

Parser tests added for the above.

This fixes #1147

### DIFF
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -208,8 +208,10 @@ categoryParser :: Parser Req
 categoryParser = Parsec.try fcesParser <|> Parsec.try andParser
 
 parseReqs :: String -> Req
-parseReqs reqString =
-    let req = Parsec.parse categoryParser "" reqString
-    in case req of
-        Right x -> x
-        Left e -> J (show e) ""
+parseReqs reqString
+    | all (== ' ') reqString || reqString == "None" = NONE
+    | otherwise =
+        let req = Parsec.parse categoryParser "" reqString
+             in case req of
+                 Right x -> x
+                 Left e -> J (show e) ""
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -5,6 +5,7 @@ import qualified Text.Parsec as Parsec
 import Text.Parsec.String (Parser)
 import Text.Parsec ((<|>))
 import Database.Requirement
+import Data.Char (toLower)
 
 -- define separators
 fromSeparator :: Parser ()
@@ -208,10 +209,12 @@ categoryParser :: Parser Req
 categoryParser = Parsec.try fcesParser <|> Parsec.try andParser
 
 parseReqs :: String -> Req
-parseReqs reqString
-    | all (== ' ') reqString || reqString == "None" = NONE
-    | otherwise =
-        let req = Parsec.parse categoryParser "" reqString
-             in case req of
-                 Right x -> x
-                 Left e -> J (show e) ""
+parseReqs reqString = do
+    let reqStringLower = [toLower c | c <- reqString]
+    if all (== ' ') reqString || reqStringLower == "none" || reqStringLower == "no"
+        then NONE
+        else do
+            let req = Parsec.parse categoryParser "" reqString
+                in case req of
+                    Right x -> x
+                    Left e -> J (show e) ""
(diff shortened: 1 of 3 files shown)


app/WebParsing/ReqParser.hs
```diff
@@ -79,8 +87,8 @@ infoParser= do
 -- a number with or without a percent symbol, or a letter A-F followed by a +/-.
 gradeParser :: Parser String
 gradeParser = do
-    grade <- Parsec.try ((Parsec.between lParen rParen percentParser <|> letterParser) <|> (percentParser <|> letterParser))
-    _ <- Parsec.lookAhead $ Parsec.choice $ map Parsec.try [
+    grade <- (Parsec.between lParen rParen percentParser <|> letterParser) <|> (percentParser <|> letterParser)
```
david-yz-liu, line 90: You don't need parentheses around `(percentParser <|> letterParser)`

app/WebParsing/ReqParser.hs
```diff
@@ -92,31 +100,47 @@ gradeParser = do
 -- parse for cutoff percentage before a course
 coBefParser :: Parser Req
 coBefParser = do
-    _ <- Parsec.choice $ map (Parsec.try . (>> Parsec.space) . Parsec.string) ["minimum grade of", "minimum mark of", "minimum of", "minimum"]
-    Parsec.spaces
-    grade <- gradeParser
+    _ <- Parsec.optional $ caseInsensitiveStr "an " <|> caseInsensitiveStr "a "
```
david-yz-liu, line 103: better to remove the space here and still use `Parsec.spaces` (as before). This will allow for parsing multiple spaces (put in accidentally, for example).

app/WebParsing/ReqParser.hs
```diff
@@ -92,31 +100,47 @@ gradeParser = do
...
+    grade <- cutoffHelper <|> gradeParser
     Parsec.spaces
     _ <- Parsec.manyTill Parsec.anyChar (Parsec.try $ Parsec.lookAhead singleParser)
     req <- singleParser
     return $ GRADE grade req
 
+    where
+    cutoffHelper = do
+        _ <- Parsec.choice $ map (Parsec.try . (>> Parsec.space) . caseInsensitiveStr)
```
david-yz-liu, line 112: Again I don't think you need the `>> Parsec.space` here. It's also okay if you don't require a space after these words (that might be another typo).

app/WebParsing/ReqParser.hs
```diff
@@ -208,8 +247,12 @@ categoryParser :: Parser Req
 categoryParser = Parsec.try fcesParser <|> Parsec.try andParser
 
 parseReqs :: String -> Req
-parseReqs reqString =
-    let req = Parsec.parse categoryParser "" reqString
-    in case req of
-        Right x -> x
-        Left e -> J (show e) ""
+parseReqs reqString = do
+    let reqStringLower = [toLower c | c <- reqString]
```
david-yz-liu, line 251: Use a `map` here instead of the comprehension

david-yz-liu (commented):
@bettyruntianwang looks great! In addition to my comments, please do a merge from master. :)

app/WebParsing/ReqParser.hs
bettyruntianwang: I put the space there because I want to prevent the letter "a" from being consumed if it is part of another word. For example, we might have prereqs in the form of "At least C+ in courseCode". Having that space seems like it is the simplest way to prevent the A from being consumed (Consuming the "A" would cause issues with `cutoffHelper` when trying to parse the grade by looking for the key word "At least").

In order to catch multiple spaces following "An" or "A", I put another call to `Parsec.spaces` after this line.

PR #1152, opened 2020-10-22
Merge PR #1145 first: Update grade to prevent letters and numbers in course codes from being recognized as a grade

### LINKED ISSUE #1146: Prerequisites with grades in the format of "A in coursecode"
When course prereqs are presented in the format of "A in MAT137H1". This is specific to the letter grade A with no "-" or "+" following it (this is because I currently skip over the optional "A" at the beginning of the prereq sentence). There are no examples of this in the database, but it should be handled.

bettyruntianwang:
For example, forestry courses have course code FOR100Y1
The parser may have picked up "F" as a grade requirement and "OR" as "or".

Also add more "and" representations such as "+", "&" and "plus".

This PR also fixes the issue when course prereqs are presented in the format of "A in MAT137H1". Previously, I skipped over the optional "A" at the beginning of the prereq sentence, so the "A" as a grade may be skipped over.

This fixes #1146


david-yz-liu:
@bettyruntianwang please do a pull from upstream/master to update this PR

PR #1154, opened 2020-11-01
Use magick executable for image conversion.

david-yz-liu:
Useful to fix issue on Windows where the pre-build check passes because of a built-in `convert` executable, but Imagemagick isn't actually installed.


david-yz-liu:
Looks like the `magick` executable isn't available on Ubuntu in the standard package libraries, so I'm closing this for now.

PR #1158, opened 2020-11-17
Add GraphOptions record type to keep track of nodes to include when generating prereq graphs

bettyruntianwang:
Add GraphOptions record type.

Previously, a POST request can be sent to the graph-generate endpoint with 2 query parameters, each taking a list of strings (course codes): taken and courses. This has been updated; the POST request will now be sent with a json body.

See [Wiki](https://github.com/Courseography/courseography/wiki/Draft-for-the-graph-generate-endpoint-options) for details on the endpoint options.


app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -22,49 +22,61 @@ import qualified Data.Map.Strict as Map
...
+sampleGraph = fst $ State.runState (reqsToGraph
+    (getDefaultGraphOptions ["MAT237H1", "MAT133H1", "CSC148H1", "CSC265H1"])
+    [("MAT237H1", J "MAT137H1" ""),
     ("MAT133H1", NONE),
     ("CSC148H1", AND [J "CSC108H1" "", J "CSC104H1" ""]),
     ("CSC265H1", AND [J "CSC148H1" "", J "CSC236H1" ""])
     ])
     (GeneratorState 0 Map.empty)
 
+getDefaultGraphOptions :: [Text] -> GraphOptions
```
david-yz-liu, line 59: Two main comments here:

1. Typically we'd put a "default" value with the data type definition itself (i.e.,g in the `GraphOptions` file.
2. Don't make `rootCourses` part of the options---notice that it's fundamentally a separate argument to the `reqsToStmts` function. Think of `GraphOptions` as only configuring the look and structure of the resulting graph.

david-yz-liu (commented):
@bettyruntianwang @jazhawang a good start, I left one inline comment.

However, my major comment is that PR should be *split up* so that the code that introduces the `GraphOptions` type and parses this value from the request is separated from the logic in `reqToStmts` that's actually using the options to change the structure of the graph.

The former is "infrastructure" code, while the latter is "feature" code that builds on top of it, and, importantly, these features can be implemented in separate branches, one for each option.

So what you should do is split up this PR as I described, so that I can get the `GraphOptions` merged in independently of the logic that's using these options.

app/DynamicGraphs/GraphGenerator.hs
bettyruntianwang: Thanks David! I removed courses from GraphOptions and made a separate branch for a [separate PR](https://github.com/Courseography/courseography/pull/1163) to introduce the GraphOptions type.

PR #1162, opened 2020-11-29
Dropdown

stefurac:
Hey David and Alex, 

working on changing the drop down menu as each option is selected. Unfortunately, I am getting undefined as the graphSelection at times, and based on the errors, I need to look into mounting and unmounting the graphSelection before I call it to update the graph.

### DIFF
--- a/js/components/graph/Sidebar.js
+++ b/js/components/graph/Sidebar.js
@@ -12,21 +12,21 @@ export default class Sidebar extends React.Component {
       graphActive: 0,
       graphName: "",
       toggled: false,
-      graphSelection: "Computer Science"
+      graphSelection: ""
     };
     this.handleGraphSelection = this.handleGraphSelection.bind(this);
 
   }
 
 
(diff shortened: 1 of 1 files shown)


js/components/graph/Sidebar.js
```diff
@@ -4,22 +4,32 @@ import Focus from "./Focus";
...
+      toggled: false,
+      graphSelection: ""
     };
+    this.handleGraphSelection = this.handleGraphSelection.bind(this);
+
   }
 
-  componentWillUpdate(prevProps) { 
+
+  componentWillUpdate(prevProps, nextProps) { console.log(prevProps); console.log(nextProps)
```
david-yz-liu, line 22: @stefurac Use `componentDidUpdate`, which takes two arguments `componentDidUpdate(prevProps, prevState)` (at least for our purposes). In the code below you're confusing props vs. state, `graphSelection` is state so you should only be using `prevState` and `this.state`, not `prevProps` and `this.props`.

stefurac: Thank you so much David! That did the trick.

PR #1171, opened 2021-02-05
remove useless `or` and `and` nodes from pruned dynamic graph

jazhawang:
Sometimes a dynamically generated graph might end up with `or` or `and` nodes with a single ingress and egress pointer. These are useless and ugly. 

This PR fixes it by changing the data structure that is used when accumulating `DotStatements` when going traversing the dependency graph. 

Before, `DotStatements` would be stored in a list which would be concatenated with other lists when building the parent node. This wasn't the best representation because we would have to iterate through the entire list when looking for useless bool nodes and their respective edge statements. This PR changes that representation to a tree, which makes it easier to remove useless bool nodes by providing faster lookup (just look at the root of the tree).

### DIFF
--- a/app/DynamicGraphs/GraphGenerator.hs
+++ b/app/DynamicGraphs/GraphGenerator.hs
@@ -30,6 +30,8 @@ import Control.Monad (mapM, liftM)
 import DynamicGraphs.GraphOptions (GraphOptions(..), defaultGraphOptions)
 import Prelude hiding (last)
 import Data.Maybe (mapMaybe)
+import Data.Graph (Tree(Node))
+import Data.Tree (flatten)
 
 -- | Generates a DotGraph dependency graph including all the given courses and their recursive dependecies
 coursesToPrereqGraph :: [String] -- ^ courses to generate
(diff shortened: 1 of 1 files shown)


app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -101,84 +103,72 @@ reqToStmts options (name, req) = do
     if pickCourse options name
         then do 
             node <- makeNode name
-            stmts <- reqToStmts' options (nodeID node) req
-            return $ DN node:stmts
+            stmts <- reqToStmtsTree options (nodeID node) req
+            return $ DN node:(concat $ flatten stmts)
```
david-yz-liu, line 107: `Tree` is an instance of [`Data.Foldable`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Foldable.html#t:Foldable), so you should be able to call `toList` on it.

app/DynamicGraphs/GraphGenerator.hs
```diff
@@ -210,6 +200,11 @@ makeEdge id1 id2 = return $ DotEdge id1 id2 [ID (id1 `mappend` "|" `mappend` id2
 mappendTextWithCounter :: Text -> Integer -> Text
 mappendTextWithCounter text1 counter = text1 `mappend` "_counter_" `mappend` (pack (show counter))
 
+filterEmptyNodes :: [Tree [a]] -> [Tree [a]]
+filterEmptyNodes [] = []
```
david-yz-liu, line 204: You can simplify this by using the built-in `filter` function. Note that `Tree` is an instance of `Eq`, so you can use `(==)` to compare `Tree`s.

app/DynamicGraphs/GraphGenerator.hs
jazhawang: Replaced `flatten` with `toList`

david-yz-liu (approved):

PR #1200, opened 2021-05-20
Lifted up position, Node ID, and InfoBox Visibility from InfoBox to Graph's State

kenneth-miura:


js/components/graph/Graph.js
```diff
@@ -342,24 +342,18 @@ export default class Graph extends React.Component {
...
 
   infoBoxMouseLeave = () => {
-    var infoBox = this.infoBox.current;
-
-    var timeout = setTimeout(function() {
-      infoBox.setState({ showInfobox: false });
+    var timeout = setTimeout(() => {
+      this.setState({showInfoBox: false});
     }, 400);
-
```
david-yz-liu, line 356: Revert this change

js/components/graph/Graph.js
```diff
@@ -342,24 +342,18 @@ export default class Graph extends React.Component {
...
     this.setState({ timeouts: this.state.timeouts.concat(timeout) });
   };
 
   infoBoxMouseClick = () => {
-    var infoBox = this.infoBox.current;
-    var newCourse = infoBox.state.nodeId.substring(0, 6);
+    var newCourse = this.state.infoBoxNodeId.substring(0,6);
```
david-yz-liu, line 356: Include a space after the comma

david-yz-liu (commented):
@kenneth-miura great work. I left a few comments, but also with the change you should remove the `ref` to the `InfoBox` component in `Graph`, because it's no longer necessary.

You can read more about refs here: https://reactjs.org/docs/refs-and-the-dom.html

PR #1201, opened 2021-05-22
Added children to graph state as object

EclipseIsDead:
As said, calculated children in the getGraph in fetch(url).then(data => ...) and verified correct children by checking the corresponding state in the Graph React component using Inspect element. The children are stored as an object literal, where the key is node id and value is a list of the Bool and Node id's.

**I have not removed the edgesJSON map from the BoolGroup render method.**

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -36,7 +36,7 @@ export default class Graph extends React.Component {
       draggingNode: null,
       currFocus: null,
       graphName: null,
-      children: null
+      connections: null
     };
 
     this.svg = React.createRef();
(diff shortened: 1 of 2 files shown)


EclipseIsDead:
Double checked that NodeGroup was taking the props correctly by looking at the rendered Node's childs and outEdges state attributes. Looks correct.

js/components/graph/Graph.js
```diff
@@ -140,6 +141,8 @@ export default class Graph extends React.Component {
         var hybridsList = [];
         var boolsList = [];
         var edgesList = [];
+        var childrenObj = {};
+        var outEdgesObj= {};
```
david-yz-liu, line 145: Insert a space before the `=`

js/components/graph/Graph.js
```diff
@@ -162,6 +165,19 @@ export default class Graph extends React.Component {
             edgesList.push(entry);
           }
         });
+
+        nodesList.forEach(function(node) {
```
david-yz-liu, line 169: This is good, but generally prefer the new arrow function syntax, `node => { ...`

PR #1202, opened 2021-05-23
Documentation for Edge and EdgeGroup

kenneth-miura:


js/components/graph/EdgeGroup.js
```diff
@@ -11,31 +13,52 @@ export default class EdgeGroup extends React.Component {
...
     });
   }
-
+  /**
+   * This function is used as a callback ref. See {@link:https://reactjs.org/docs/refs-and-the-dom.html#callback-refs}
+   * @param {JSON} edgeJSON Represents a single edge
+   * @returns {function} A function that adds/updates a key value pair to EdgeGroup of an Edge's ID
+   *  and the Edge object
+   */
   setRefEntry = edgeJSON =>{
```
kenneth-miura, line 50: I think this function is very confusing, especially for beginners to Javascript. I think it could be easily reimplemented as an if-else statement following the idea below : 
* if `elem` is null, add the key value pair
* otherwise, don't do anything

In my opinion, this would be easier to read. Unless this is a standard pattern for JS that is good for SDS members to be exposed to, I think this should be refactored.

david-yz-liu: This is a common programming idiom (not just in Javascript). It's good that you encountered it, but it doesn't warrant a special comment. Please remove the inline comments you added here.

kenneth-miura: Oh ok, sounds good. Will remove. Thank you!

PR #1203, opened 2021-05-23
Map nodeID to parents

leenlab2:
Created an object mapping nodeIDs to nodeIDs/boolIDs of the nodes preceding the current one (parent nodes). This includes:

- Added a new `parents` state to `Graph`
- Initialized an object `parentsObj` with the `node.id_` as the keys and a blank list for values by iterating over `nodesList`
- Updated the object by iterating over the `edgesList`
- Added a comment explaining why this object was created
- Set `parents` to `parentsObj`

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -162,7 +162,7 @@ export default class Graph extends React.Component {
           }
         });
 
-        // An object connecting a node to its parents/childs improves efficiency in NodeGrp/BoolGrp
+        // An object mapping a node to its parents/childs improves efficiency in NodeGroup/BoolGroup
         nodesList.forEach(node => {
           parentsObj[node.id_] = [];
         });
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -724,6 +724,7 @@ export default class Graph extends React.Component {
             edgesJSON={this.state.edgesJSON}
             highlightedNodes={this.state.highlightedNodes}
             onDraw={this.state.onDraw}
+            parentTree={this.state.parentTree}
           />
           <BoolGroup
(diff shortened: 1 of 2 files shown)


leenlab2:
Update:

- Removed the old `parents` state from Graph
- Added a new `parentTree` state where `parentTree.parents` refers to the old object, and `parentTree.inEdges` refers to a similar object that holds edgeIDs
- Passed `parentTree` down to NodeGroup as a prop
- Passed `parentTree.parents` and `parentTree.inEdges` down to Node as props
- Removed previous code that created these props
- Updated NodeGroup PropTypes

js/components/graph/Graph.js
```diff
@@ -159,6 +162,20 @@ export default class Graph extends React.Component {
             edgesList.push(entry);
           }
         });
+
+        // An object mapping a node to its parents/childs improves efficiency in NodeGroup/BoolGroup
```
david-yz-liu, line 166: Remove this comment. The term "improves" implies a comparison or change that won't make sense once the old code is deleted. This is a situation where the "why" is best described in the pull request description.

david-yz-liu (commented):
@leenlab2 great work! I left one comment. Also, you will need to resolve a merge conflict with master. This one should be pretty simple to solve: do a pull from master, and then look for the conflict markers `>>>>>>>`/`<<<<<<<<` , which mark the sections in the conflict. You can talk to your team members too.

Ref: https://docs.github.com/en/github/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line

leenlab2:
Update:

- I removed the comment like you asked
- I resolved the conflict with master

PR #1207, opened 2021-05-27
moved unchanging style for InfoBox to css

kenneth-miura:
## Motivation and Context

There was a TODO in InfoBox.js for moving the style to CSS.
## Your Changes

**Description**:
I moved style definitions that are not dynamic to App.scss. I kept the opacity in InfoBox.jss because I think it makes more sense to keep, as it needs to be changed depending on whether the infobox should be shown.

## Testing

I manually tested infobox visibility related functionality, meaning I tested that the infobox shows up if you hover over a course or the infobox and that the infobox disappears if you stop hovering over the course or infobox.

Is it better to keep the `opacity` in InfoBox.js, or would it be better to move it to CSS somehow? The only way I see how to do that in CSS is to create 1 class for visible elements and 1 for invisible elements and append the correct class name when InfoBox is rendered.
## Checklist

  passed.

### DIFF
--- a/js/components/graph/InfoBox.js
+++ b/js/components/graph/InfoBox.js
@@ -5,10 +5,7 @@ export default class InfoBox extends React.Component {
 
   render() {
     if (this.props.showInfoBox) {
-      //TODO: move to CSS
       var gStyles = {
-        cursor: "pointer",
-        transition: "opacity .4s",
         opacity: this.props.showInfoBox ? 1 : 0
       };
       var rectAttrs = {
--- a/js/components/graph/InfoBox.js
+++ b/js/components/graph/InfoBox.js
@@ -4,47 +4,42 @@ import PropTypes from "prop-types";
 export default class InfoBox extends React.Component {
 
   render() {
-    if (this.props.showInfoBox) {
-      var gStyles = {
-        opacity: this.props.showInfoBox ? 1 : 0
-      };
-      var rectAttrs = {
-        id: this.props.nodeId + "-tooltip" + "-rect",
-        x: this.props.xPos,
-        y: this.props.yPos,
-        rx: "4",
-        ry: "4",
-        fill: "white",
-        stroke: "black",
-        strokeWidth: "2",
-        width: "60",
-        height: "30"
-      };
+    var className = this.props.showInfoBox ? "tooltip-group-display" : "tooltip-group-hidden";
 
-      var textAttrs = {
-        id: this.props.nodeId + "-tooltip" + "-text",
-        x: this.props.xPos + 60 / 2 - 18,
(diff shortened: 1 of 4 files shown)


style/app.scss
```diff
@@ -2443,3 +2443,9 @@ main-filter input
   left : -388px;
   top  : -33px;
 }
+
+.tooltip-group
+{
+  cursor: pointer;
+  transition: opacity .4s;
+}
```
david-yz-liu, line 2451: Make sure to end the file with a blank line.

david-yz-liu (commented):
@kenneth-miura good question, yes in these situations it's typical to create two classes like `tooltip-group-active` and `tooltip-group-display` with the relevant styles, and then choose the right class in the render method.

Also, here's a different technique than using opacity that I'd like you to try out: https://greywyvern.com/?post=337.

js/components/graph/InfoBox.js
```diff
@@ -5,10 +5,7 @@ export default class InfoBox extends React.Component {
 
   render() {
     if (this.props.showInfoBox) {
-      //TODO: move to CSS
       var gStyles = {
-        cursor: "pointer",
-        transition: "opacity .4s",
         opacity: this.props.showInfoBox ? 1 : 0
```
kenneth-miura, line 9: Taking a second look at this code, the opacity is functionally hardcoded to 1, as the opacity is only set when `this.props.showInfoBox` is true, meaning it's only ever set to 1. The way we are making the infobox disappear is by returning an empty HTML <g> tag if `this.props.showInfoBox` is false. 

Should I refactor it so it's visibility is determined by css, so that I can try out that CSS transitions based approach you mentioned?

david-yz-liu: Sorry I missed this question, but yes the visibility should be determined only by CSS, and not by this `opacity` rule here.

kenneth-miura: Ok. I've made the change, but if you hover over a course till the Infobox fades in, and then you hover over another, the Infobox doesn't fade in, it just appears. 

I think this isn't ideal. Do you want me to take a stab at it in this PR, or wait till a separate one?

kenneth-miura:
In order to test visibility of the `Infobox`, should I be checking that the `Infobox` element has the right `classname` (Since `tooltip-group-display` means it's visible, and `tooltip-group-display` means it's invisible)? Or should I be directly checking the CSS of the `Infobox` element?

It seems like checking CSS is closer to the "source of truth" since there's a chance the CSS classes are not actually affecting visibility. However, there's multiple ways to make a CSS element visible, so I think checking the `classname` makes more sense. Which option is correct?

PR #1209, opened 2021-05-30
Removed the this.marker ref and put attributes in renderArrowHead

leenlab2:
Removed an unnecessary `this.marker` reference, by moving the `markerNode` attributes to the JSX in the `Graph.renderArrowHead` method. 

## Motivation and Context

The attributes of `<marker>`, which describes the arrow heads of the graph, used to be hardcoded into `Graph.componentDidMount` since React did not used to support them. As of React 16, those SVG attributes are fully supported and can be added onto the JSX where the HTML is being rendered directly.

## Your Changes

**Description**: Deleted the `this.marker` reference and the lines setting the attributes of `markerNode`. Instead, I added `refX`, `refY`, `markerWidth`, `markerHeight`, `markerUnits`, `orient` attributes into the `renderArrowHead` method's JSX output.

## Testing

- Checked that no visible change occurred to the graph
- Checked that the attributes of `<marker>` in the page's source code did not change

I wanted to get initial approval on the `markerNode` changes before experimenting with `svgNode`. I'll update this PR if what I did so far seems good.

## Checklist

  passed.

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -44,7 +44,6 @@ export default class Graph extends React.Component {
     };
 
     this.svg = React.createRef();
-    this.marker = React.createRef();
     this.nodes = React.createRef();
     this.bools = React.createRef();
     this.edges = React.createRef();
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@leenlab2 excellent work! Please go ahead with making the same changes with the `this.svg` ref, and you can add those changes directly to this branch.

PR #1210, opened 2021-05-31
Replaced this.modal refs with Graph's state

EclipseIsDead:
## Motivation and Context

Within ```Graph.js``` there were many ```this.modal``` refs which were unnecessary. Instead, I have used Graph's state and ```showCourseModal``` to determine whether or not to show the Info Modal.

## Your Changes

**Description**:

First I added ```showCourseModal``` to store whether or not the Info Modal needs to be displayed. This is updated using the function ```callbackModal``` which along side ```showCourseModal``` is passed as props into the ```CourseModal``` component. In order to determine whether or not the props of ```CourseModal``` have changed I used the ```componentDidUpdate``` function to call ```openModal```.

## Testing
This was manually tested in my web browser with the original Courseography side-by-side. I also verified that the state and props were being correctly updated and passed, respectively. 

Some courses cannot be pulled from the database resulting in MeetingTimes to throw and error in the javascript console, and there was a bug early on that my second commit fixed where you would either need to click  and then move your mouse outside the button or double click to activate the modal. I have no idea why that happened or why the bug stopped.

## Checklist

### DIFF
--- a/js/components/common/react_modal.js.jsx
+++ b/js/components/common/react_modal.js.jsx
@@ -50,7 +50,14 @@ class CourseModal extends React.Component {
   }
 
   closeModal() {
+    this.props.callbackModal();
     this.setState({ modalIsOpen: false });
+  }
+
+  componentDidUpdate (prevProps) {
+    if ((prevProps !== this.props) && (this.props.courseChange !== "")) {
+      this.openModal(this.props.courseChange);
+    };
   }
 
   render() {
--- a/js/components/common/react_modal.js.jsx
+++ b/js/components/common/react_modal.js.jsx
@@ -36,7 +36,6 @@ class CourseModal extends React.Component {
             course: course,
             sessions: course.allMeetingTimes.sort((firstLec, secondLec) =>
               firstLec.meetData.session > secondLec.meetData.session ? 1 : -1),
-            courseId: this.props.courseId,
             courseTitle: `${this.props.courseId.toUpperCase()} ${course.title}`
           });
         })
--- a/js/components/common/react_modal.js.jsx
+++ b/js/components/common/react_modal.js.jsx
@@ -23,40 +23,28 @@ class CourseModal extends React.Component {
       courseId: '',
       course: [],
       sessions: [],
-      modalIsOpen: false,
       courseTitle: '',
     };
-    this.openModal = this.openModal.bind(this);
     this.closeModal = this.closeModal.bind(this);
   }
 
-  openModal(newCourse) {
-    if (newCourse === this.state.courseId) {
-      this.setState({ modalIsOpen: true });
-    } else {
-      getCourse(newCourse)
+  closeModal() {
+    this.props.onClose();
+  }
+
+  componentDidUpdate (prevProps) {
+    if ((prevProps.courseId !== this.props.courseId) && (this.props.courseId !== "")) {
+      getCourse(this.props.courseId)
         .then(course => {
-            //Tutorials don't have a timeStr to print, so I've currently omitted them
+          // Tutorials don't have a timeStr to print, so I've currently omitted them
           this.setState({
             course: course,
             sessions: course.allMeetingTimes.sort((firstLec, secondLec) =>
               firstLec.meetData.session > secondLec.meetData.session ? 1 : -1),
-            modalIsOpen: true,
-            courseId: newCourse,
-            courseTitle: `${newCourse.toUpperCase()} ${course.title}`
+            courseId: this.props.courseId,
+            courseTitle: `${this.props.courseId.toUpperCase()} ${course.title}`
           });
         })
-    }
-  }
-
-  closeModal() {
-    this.props.callbackModal();
-    this.setState({ modalIsOpen: false });
-  }
-
-  componentDidUpdate (prevProps) {
-    if ((prevProps !== this.props) && (this.props.courseChange !== "")) {
-      this.openModal(this.props.courseChange);
     };
   }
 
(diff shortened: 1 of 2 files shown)


js/components/common/react_modal.js.jsx
```diff
@@ -50,9 +50,16 @@ class CourseModal extends React.Component {
   }
 
   closeModal() {
+    this.props.callbackModal();
     this.setState({ modalIsOpen: false });
   }
 
+  componentDidUpdate (prevProps) {
+    if ((prevProps !== this.props) && (this.props.courseChange !== "")) {
```
david-yz-liu, line 58: Never compare the prop object themselves, only specific values. In this case, you should compare the course code prop.

js/components/graph/Graph.js
```diff
@@ -385,10 +385,16 @@ export default class Graph extends React.Component {
...
     var newCourse = this.state.infoBoxNodeId.substring(0, 6);
-    this.setState({ courseId: newCourse });
-    this.modal.current.openModal(newCourse);
+    this.setState({
+      courseId: newCourse,
+      showCourseModal: true
+    });
   };
 
+  callbackModal = () => {
```
david-yz-liu, line 394: This should have a more specific name like `onClose`

js/components/graph/Graph.js
```diff
@@ -641,7 +647,7 @@ export default class Graph extends React.Component {
 
     return (
       <div id="react-graph" className="react-graph" onClick={this.props.closeSidebar}>
-        <CourseModal ref={this.modal} />
+        {this.state.showCourseModal && <CourseModal courseChange={this.state.courseId} callbackModal={this.callbackModal} />}
```
david-yz-liu, line 650: Rather than conditionally render the `CourseModal` component, It's better to pass the `showCourseModal` as a prop that controls whether it appears or not. This prop replaces the `modalIsOpen` state from the `CourseModal` component.

Also, don't think of the course code as a "change", it's just what course that's being shown for the modal. Again, this prop should replace one of the states of the `CourseModal` component.

js/components/common/react_modal.js.jsx
```diff
@@ -50,9 +50,16 @@ class CourseModal extends React.Component {
...
+      this.openModal(this.props.courseChange);
```
david-yz-liu, line 59: Merge this function with the `openModal` function above. The part to keep is the call to `getCourse`.

david-yz-liu (commented):
@EclipseIsDead I left some overall comments, but didn't look into the failing test. That's something you should investigate more by running the tests on your local computer.

js/components/common/react_modal.js.jsx
```diff
@@ -23,41 +23,36 @@ class CourseModal extends React.Component {
...
-    this.openModal = this.openModal.bind(this);
     this.closeModal = this.closeModal.bind(this);
   }
 
-  openModal(newCourse) {
-    if (newCourse === this.state.courseId) {
-      this.setState({ modalIsOpen: true });
-    } else {
-      getCourse(newCourse)
+  closeModal() {
```
david-yz-liu, line 31: This function isn't necessary to define. You can just past `this.props.onClose` directly to the `ReactModal` component.

js/components/common/react_modal.js.jsx
```diff
@@ -23,41 +23,36 @@ class CourseModal extends React.Component {
...
+    this.props.onClose();
+  }
+
+  componentDidUpdate (prevProps) {
+    if ((prevProps.courseId !== this.props.courseId) && (this.props.courseId !== "")) {
```
david-yz-liu, line 36: Remove the parentheses around each subexpression of the `&&`.

js/components/common/react_modal.js.jsx
```diff
@@ -23,41 +23,36 @@ class CourseModal extends React.Component {
...
-  openModal(newCourse) {
-    if (newCourse === this.state.courseId) {
-      this.setState({ modalIsOpen: true });
-    } else {
-      getCourse(newCourse)
+  closeModal() {
+    this.props.onClose();
+  }
+
+  componentDidUpdate (prevProps) {
```
david-yz-liu, line 35: Remove the space before the `(`

js/components/common/react_modal.js.jsx
```diff
@@ -23,41 +23,36 @@ class CourseModal extends React.Component {
...
-            //Tutorials don't have a timeStr to print, so I've currently omitted them
+          // Tutorials don't have a timeStr to print, so I've currently omitted them
           this.setState({
             course: course,
             sessions: course.allMeetingTimes.sort((firstLec, secondLec) =>
               firstLec.meetData.session > secondLec.meetData.session ? 1 : -1),
-            modalIsOpen: true,
-            courseId: newCourse,
-            courseTitle: `${newCourse.toUpperCase()} ${course.title}`
+            courseId: this.props.courseId,
```
david-yz-liu, line 44: You can remove the `courseId` state entirely (in general, do not use state that duplicates a value from a prop).

david-yz-liu (commented):
@EclipseIsDead looking good, I left a review with some smaller details now. I think this should be good to go once you've addressed these!

js/components/common/react_modal.js.jsx
david-yz-liu: @EclipseIsDead I think you missed this comment.

EclipseIsDead: That's odd I swear I saw it in git diff. I implemented it this time though.

PR #1211, opened 2021-05-31
Moving Hybrids Parents & Children to Connections

kenneth-miura:
## Motivation and Context
We wanted to stop recalculating Hybrid's parents & children every time we render.

## Your Changes

**Description**:

I moved the calculation of the hybrid nodes' parents and children to Graph.js. Because finding the relationships between hybrids and nodes is required to find the hybrid's parents, I also moved the calculation of `hybridRelationships` to Graph.js. Finally, I used `hybridRelationships` in Graph.js to add to normal Node's child.

## Testing

No tests added. I manually checked the children and parents of the hybrids, and I also checked the children of the normal nodes that were expected to have hybrid children.
## Checklist

  passed.

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -7,6 +7,7 @@ import Button from "./Button";
 import EdgeGroup from "./EdgeGroup";
 import InfoBox from "./InfoBox";
 import NodeGroup from "./NodeGroup";
+import {parseAnd} from "./NodeGroup";
 import RegionGroup from "./RegionGroup";
 import * as focusInfo from "./sidebar/focus_descriptions";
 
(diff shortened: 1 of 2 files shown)


david-yz-liu (commented):
@kenneth-miura great start! A couple of overall comments:

1. The data representation can be simplified. Hybrids and regular nodes have different ids, so they can be in the same dictionary. So in the `connections` state, add the hybrids to the existing objects, don't create separate ones.
2. The whole `hybridRelationships` object is carried over from the existing code, but I think it's more complicated than it needs to be. Ideally your helper function would take in the hybridJSON and existing connections object(s), and mutate those objects directly. (This might require changing *when* the helper function is called, depending on the implementation.)
3. Move the `parseAnd` and `parseOr` functions to `Graph.js`. These functions, plus the hybrid helper, can be defined as top-level functions outside of the `Graph` class. Put them at the bottom of the file, using a comment to indicate that these are helpers.

kenneth-miura:
I'm unsure if the changes I made to the hybridRelationships setup are what you wanted. I think I could make the helper function that handles populating the parent's array with the hybrid nodes simpler, but I would have to remove all the `parseAnd`, `parseOr`, etc functions. 

I think that should be a separate PR. I would be using string splitting on the string that contains all the pre-reqs. As far as I can tell, the only separator used is the "/". I think with that approach, we wouldn't need to separately handle the case where the pre-reqs string contains separators.

js/components/graph/Graph.js
```diff
@@ -109,7 +109,6 @@ export default class Graph extends React.Component {
       .getElementById("react-graph")
       .removeEventListener("wheel", this.onWheel);
   }
-
```
david-yz-liu, line 112: Revert this change

js/components/graph/Graph.js
```diff
@@ -145,7 +144,6 @@ export default class Graph extends React.Component {
         var inEdgesObj = {};
         var childrenObj = {};
         var outEdgesObj = {};
-
```
david-yz-liu, line 148: Revert this change

js/components/graph/Graph.js
```diff
@@ -168,13 +166,21 @@ export default class Graph extends React.Component {
           }
         });
 
+
```
david-yz-liu, line 169: Revert this change (and the additional blank lines below)

js/components/graph/Graph.js
```diff
@@ -204,7 +211,7 @@ export default class Graph extends React.Component {
             'parents': parentsObj,
             'inEdges': inEdgesObj,
             'children': childrenObj,
-            'outEdges': outEdgesObj
+            'outEdges': outEdgesObj,
```
david-yz-liu, line 214: Revert this change

js/components/graph/Graph.js
```diff
@@ -757,6 +764,173 @@ export default class Graph extends React.Component {
...
+
+  if (s.length === 3) {
+    return [prefix + s, ""];
+  }
+
+  return [s, ""];
+}
+
+
+  /** Helper function that adds parents of hybridNode to the parents object, and adds hybrid nodes as children of the Nodes they represent
```
david-yz-liu, line 869: Unindent this line.

js/components/graph/Graph.js
```diff
@@ -757,6 +764,173 @@ export default class Graph extends React.Component {
...
+ *
+ * @param {Node} hybridNode
+ * @param {Array} nodesJSON
+ * @param {Object} parents
+ * @param {Object} childrenObj
+ */
+function populateHybridRelatives(hybridNode, nodesJSON, parents, childrenObj){
+
+  /**
```
david-yz-liu, line 878: In general don't nest functions definitions unless you need to access some variables from the outer scope (which you could be doing, but aren't, in this case). You can move this function to below `populateHybridRelatives`.

js/components/graph/Graph.js
```diff
@@ -757,6 +764,173 @@ export default class Graph extends React.Component {
...
+            console.error("Could not find prereq for ", hybridText);
+          }
+        });
+        if (orPrereq.length > 0) {
+          nodeParents.push(orPrereq);
+        }
+      }
+    });
+  }
+parents[hybridNode.id_] = nodeParents;
```
david-yz-liu, line 931: Indent this line

js/components/graph/NodeGroup.js
```diff
@@ -48,80 +48,28 @@ export default class NodeGroup extends React.Component {
...
           );
         })}
         {this.props.nodesJSON.map(entry => {
           var highlighted = highlightedNodes.indexOf(entry.id_) >= 0;
-          hybridRelationships.forEach(element => {
-            if (element[0] === entry.id_) {
-              this.props.connections.children[entry.id_].push(element[1]);
-            }
-          });
+          var childs = this.props.connections.children[entry.id_].slice();
```
david-yz-liu, line 72: This is interesting, I'm not sure why creating a copy (using `slice`) is necessary. Is this list being mutated somewhere when rendering the `Node`?

david-yz-liu (commented):
@kenneth-miura great work, this is exactly what I was looking for. I agree the parsing code could be simplified, however it is important to support both "and" and "or" syntax for the hybrid text, using `,` and `/` respectively. We can discuss this in a future PR.

I left a few style comments and one question for you to investigate.

js/components/graph/NodeGroup.js
kenneth-miura: This is a remnant from before the changes this PR. The copying was to stop the issue where nodes that correspond to hybrids would get their hybrid children re-added every time `render()` ran. I'm removing it now that we no longer have `hybridRelationships`.

PR #1213, opened 2021-06-04
Zooming

kenneth-miura:
## Motivation and Context
This is my implementation of zooming. There is also panning implemented, but we plan to overwrite it when we do our final merge. 

## Your Changes

**Description**:
Removed panning buttons, and added buttons for zooming in and out. Implemented zooming in and out, as well as panning, through the keyboard.

## Testing

I tested manually. 

## Checklist

  passed.


david-yz-liu (commented):
@kenneth-miura good stuff, I have one main comment: set the cursor style using a component state + class and then CSS rule (similar to what you did for InfoBox). Also I don't think we need to use a timeout for just changing that style.

PR #1214, opened 2021-06-07
Panning navigation

leenlab2:
Replaced the arrow buttons with drag panning navigation for more intuitive control of the graph. 

## Motivation and Context
The previous use of arrow buttons to move the graph around was a bit difficult, and using drag panning has become more of a standard in map-like applications.

## Your Changes

**Description**: 

1. Removed arrow buttons and associated methods, and updated the css to push the +/- and reset buttons up
2. Removed the old panning functions
3. Removed `verticalPanFactor` and `horizontalPanFactor` states, added new `viewBox` related states instead
4. Added new panning methods and mouseDown listener to svg
5. Altered methods controlling the attributes of the remainig buttons
6. Documented the methods I added

## Testing

Manual testing in web browser to ensure functionality, and compared with other approaches to find possible bugs.

I wasn't sure what datatype to label the `event` parameter with, so I left it as `{*}`. Also, the zooming does not work in this PR yet, as Kenneth is going to make a PR soon that fixes that. Some of the code I removed, such as the containerWidth part, is going to be in that PR as well, since we faced some related bugs during our initial exploration. As for merge conflicts, it would be best to merge in this PR first as it has most of the base code. 

We decided to keep the +/- buttons as a backup control method since they do that on maps (if someone does not have a working mouse wheel, for example). We agreed among ourselves that removing the arrow keys would be the best UI-wise, but we will still implement `onKeyDown` such that one can control the panning with the keyboard arrow keys instead. That change will also be in Kenneth's PR. However, if you disagree we can undo this change.

As for whether the panning causes issues with selecting nodes, it does not seem to. If you just click, it selects the node, but if you click and drag it pans first then selects the node.

Finally, I did not add any new tests for this since I wanted to wait for your feedback first. 

## Checklist

  passed. 
- [ ] I have added tests for my changes.

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -167,6 +167,16 @@ export default class Graph extends React.Component {
           }
         });
 
+        let containerWidth = 0;
+        let containerHeight = 0;
+        if (document.getElementById("react-graph") !== null && document.getElementById("generateRoot")) {
+          containerWidth = document.getElementById("react-graph").clientWidth;
+          containerHeight = document.getElementById("generateRoot").clientHeight;
+        }
+
+        const viewboxWidth = Math.max(this.state.width, containerWidth) * this.state.zoomFactor;
+        const viewboxHeight = Math.max(this.state.height, containerHeight) * this.state.zoomFactor;
+
         this.setState({
           labelsJSON: labelsList,
           regionsJSON: regionsList,
@@ -177,7 +187,7 @@ export default class Graph extends React.Component {
           width: data.width,
           height: data.height,
           zoomFactor: 1,
-          viewBoxPos: {x: 0, y: 0, width: window.innerWidth, height: window.innerHeight},
+          viewBoxPos: {x: 0, y: 0, width: viewboxWidth, height: viewboxHeight},
           graphName: graphName,
           connections: {
             'parents': parentsObj,
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -42,8 +42,7 @@ export default class Graph extends React.Component {
       panning: false,
       panStartX: 0,
       panStartY: 0,
-      viewBoxPos: {x: 0, y: 0},
-      viewBoxDim: {width: window.innerWidth, height: window.innerHeight},
+      viewBoxPos: {x: 0, y: 0, width: window.innerWidth, height: window.innerHeight},
       viewBoxContainerRatio: 1,
       showCourseModal: false
     };
@@ -178,7 +177,7 @@ export default class Graph extends React.Component {
           width: data.width,
           height: data.height,
           zoomFactor: 1,
-          viewBoxPos: {x: 0, y: 0},
+          viewBoxPos: {x: 0, y: 0, width: window.innerWidth, height: window.innerHeight},
           graphName: graphName,
           connections: {
             'parents': parentsObj,
@@ -357,7 +356,7 @@ export default class Graph extends React.Component {
 
   /**
    * Initializes the panning process by recording the position of the mouse pointer and adding event listeners.
-   * @param {*} event
+   * @param {Event} event
    */
   onMouseDownInGraph = event => {
     document.body.style.cursor = "grab";
@@ -376,7 +375,7 @@ export default class Graph extends React.Component {
 
   /**
    * Pans the graph by moving it in the direction that the mouse moved.
-   * @param {*} event
+   * @param {Event} event
    */
   onMouseMoveInGraph = event => {
     if (this.state.panning) {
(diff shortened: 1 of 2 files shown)


js/components/graph/Graph.js
```diff
@@ -41,6 +39,12 @@ export default class Graph extends React.Component {
       infoBoxXPos: 0,
       infoBoxYPos: 0,
       infoBoxNodeId: "",
+      panning: false,
+      panStartX: 0,
+      panStartY: 0,
+      viewBoxPos: {x: 0, y: 0},
```
david-yz-liu, line 45: Let's use one combined dictionary for all 4 viewbox states (x, y, width, height).

js/components/graph/Graph.js
```diff
@@ -174,8 +178,7 @@ export default class Graph extends React.Component {
           width: data.width,
           height: data.height,
           zoomFactor: 1,
-          horizontalPanFactor: 0,
-          verticalPanFactor: 0,
+          viewBoxPos: {x: 0, y: 0},
```
david-yz-liu, line 181: The previous code sets the viewbox width/height as a max of the container's width/height and the `data.width` and `data.height`, so I think you should do the same thing here.

js/components/graph/Graph.js
```diff
@@ -338,6 +341,77 @@ export default class Graph extends React.Component {
...
+      adjustedY: clientY + this.state.viewBoxPos.y
+    };
+  }
+
+  /**
+   * Initializes the panning process by recording the position of the mouse pointer and adding event listeners.
+   * @param {*} event
+   */
+  onMouseDownInGraph = event => {
+    document.body.style.cursor = "grab";
```
david-yz-liu, line 363: You can use css for this. Update the `render` method to add a class to one of the container elements when panning, and then use the `cursor` css rule.

js/components/graph/Graph.js
```diff
@@ -338,6 +341,77 @@ export default class Graph extends React.Component {
...
+  adjustCoordsToViewbox = (clientX, clientY) => {
+    return {
+      adjustedX: clientX + this.state.viewBoxPos.x,
+      adjustedY: clientY + this.state.viewBoxPos.y
+    };
+  }
+
+  /**
+   * Initializes the panning process by recording the position of the mouse pointer and adding event listeners.
+   * @param {*} event
```
david-yz-liu, line 360: You can use [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event)

js/components/graph/Graph.js
```diff
@@ -338,6 +341,77 @@ export default class Graph extends React.Component {
...
+   */
+  onMouseDownInGraph = event => {
+    document.body.style.cursor = "grab";
+    const {adjustedX, adjustedY} = this.adjustCoordsToViewbox(event.clientX, event.clientY);
+
+    document.getElementById("react-graph").addEventListener("mouseup", this.onMouseUpInGraph);
```
david-yz-liu, line 366: It's not a good practice to add/remove event listeners in React; they should always be attached.

I think you mentioned encountering issues about two different event handlers being called; you can experiment with [`Event.preventDefault`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) or perhaps `Event.stopPropagation` to deal with this.

js/components/graph/Graph.js
```diff
@@ -338,6 +341,77 @@ export default class Graph extends React.Component {
...
+      panStartX: adjustedX,
+      panStartY: adjustedY
+    });
+  }
+
+  /**
+   * Pans the graph by moving it in the direction that the mouse moved.
+   * @param {*} event
+   */
+  onMouseMoveInGraph = event => {
```
david-yz-liu, line 381: It is better to use method names that reflect the logical behaviour of the method, not the context you expect it to be called from. So for example, something like `panGraph` is better since it describes the action being taken.

david-yz-liu (commented):
@leenlab2 very exciting. I left some comments for you to work on. When you're done, please also coordinate with the other Courseography team members so that they can update their branches as well (I recommend just doing a git merge).

leenlab2:
Alright, I made the changes you asked for. Let me know if this is better!
Also, should I start making tests, or should I fix generate.js first? Since generate has been broken since we added `connections` state

david-yz-liu:
@leenlab2 you can start writing tests! Why don't you coordinate with the other team members to create a *separate* PR that fixes the generate page (independent of these panning changes).

leenlab2:
Alright, I will! Also, I just implemented the comment about the viewbox width and height, since I missed it in my last commit. We had previously moved this part of the code to one of the zooming methods in Kenneth's PR. I'll get working on the tests now!

leenlab2:
Update: 

- Removed `viewBoxPos` and `viewBoxContainerRatio` and replaced them with `horizontalPanFactor` and `verticalPanFactor`
- Removed the method adjusting coordinates
- Added back in the code that calculates the viewbox attributes in render()

david-yz-liu:
Merged as part of #1213.

Issue #1218, opened 2021-06-25
Graph Page Redesign


kenneth-miura:
This issue details changes to the Graph page’s layout.

## Menu Bar
- [x] Reduce the size of the Courseography logo
- [x] Make the Courseography logo link to the home page (Graph)
---
- [x] Center the navigation bar items (horizontally and vertically)
- [x] Remove the black border/shading around the navigation bar items
- [x] Make the font color change to yellow/add a yellow highlight bar on the current page
---
- [x] Remove the Export button from the navigation bar on all relevant pages
- [x] Add a new Export icon button on the rightmost end of the menu bar
---
- [x] Remove the black border surrounding the menu bar
## Control Buttons
- [x] Alter background to be light grey
- [x] Change white navigation buttons to be borderless
---
- [x] Create Reset View Button with new symbol
- [x] Create Reset Selection Button with new symbol
- [x] Reformat zoom in, zoom out buttons to be one over the other
## Focus Bar
- [x] Remove the focuses section in the old sidebar
 ---
- [x] Add a new button that opens the focus bar at the bottom left
- [x] Implement the page tab styled buttons
- [x] Add a selection animation that enlarges the selected tab
- [x] Add shading to the tabs depending on the selection state
---
- [x] Add information icon on the bottom right
- [x] Make the focus button change into "Close" when the bar opens

## Graph Dropdown
- [x] Remove Sidebar
- [x] Add Dropdown list upon hover of “Graph” on Navbar at the top

## Nodes
- [x] Resize the graph to fit the new layout such that it is centered and fits the height of the screen 
## Disclaimer
- [x] Create new React Component for Disclaimer
- [x] Add disclaimer component and adjust css accordingly
## FCE Counter + Search bar
- [x] Remove FCE count from sidebar
- [x] Implement FCE Count floating dialogue, with hover box label
- [x] Implement removing courses from FCE Count floating dialogue
- [x] search bar for adding courses

PR #1220, opened 2021-06-28
Disclaimer pop up

leenlab2:
Changed the disclaimer into a dismissable popup box right below the menu bar on all pages.

## Motivation and Context

The disclaimer used to take up too much unnecessary space on the page. This PR is part of a set of layout changes detailed in Issue #1218.

## Your Changes

**Description**:

- Removed the old disclaimer
- Resized container to account for removed disclaimer
- Created a new Disclaimer react component
- Inserted that component into `<Container />`, `<Grid />`, and `Generate`.
- Increased the `z-index` of `.overlay` to hide the disclaimer
- Changed link colors on the disclaimer to account for the different background color
- Added a 'Do Not Show Again' checkbox to stop the disclaimer from showing up every time

- [x] Other: User interface changes

## Testing

Manual testing in browser.

None!

## Checklist

  passed.


leenlab2:
Comparison screenshots to showcase changes:

Before
image

After
image

leenlab2:
Alright, I'll get working on it! I used ids instead of classes for some since they were already present in the css file (eg #disclaimerDiv), but I'll work on improving that.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+
+/**
+ * A React component representing the disclaimer popup
+ */
+export default class Disclaimer extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            dontShowDisclaimer: localStorage.getItem('DontShowDisclaimer')
```
david-yz-liu, line 10: The names here could be better. The state name should be something standard like `hidden`, and the local storage name a corresponding `hide-disclaimer`.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    handleClose = () => {
```
david-yz-liu, line 14: Avoid modifying the DOM directly, since we want to use React to do this for us. This method should set the "hide disclaimer" state to `true`. You don't need the `dismiss` class or styling, I don't think the transition is worth it.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+        let disclaimer = document.getElementById("disclaimerDiv");
+        disclaimer.classList.toggle("dismiss");
+    }
+
+    handleCheck = () => {
```
david-yz-liu, line 19: The checkbox can be both selected and deselected. So you need to access the event object to see whether the box is checked or unchecked, as this determines what to set the local storage value to.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+        localStorage.setItem('DontShowDisclaimer', 'true');
+    }
+
+    render() {
+        const timetable = <a href="https://timetable.iit.artsci.utoronto.ca/">Official Timetable</a>;
+        const calendar = <a href="http://calendar.artsci.utoronto.ca/">Calendar</a>;
```
david-yz-liu, line 25: It looks like the new permanent URL is `https://artsci.calendar.utoronto.ca/`, so please update to use this. Also, please change "Calendar" to "Academic Calendar".

js/components/common/Disclaimer.js
david-yz-liu: Oh also, do the conversion to a boolean here rather than in the `render`. Logically, this state should be a bool, not a string.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+                    <label>
+                        <input type="checkbox" id="disclaimerCheck" class="dont-show-checkbox" onClick={this.handleCheck} />
+                        Do not show this again
+                    </label>
+                    <button id="closeDisclaimer" class="close-banner" onClick={this.handleClose}>&times;</button>
+                </div>
+            )
+        }
+    }
+}
```
david-yz-liu, line 44: Make sure to end this file with a blank line.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+                    <p><b>DISCLAIMER:</b> Both the {timetable} and {calendar} take precedence over the information
+                    presented here. It's important that you double-check your course selection, prerequisites,
+                    and your program plans. Some graph edges may represent a corequisite rather than a prerequisite.</p>
+                    <label>
+                        <input type="checkbox" id="disclaimerCheck" class="dont-show-checkbox" onClick={this.handleCheck} />
+                        Do not show this again
+                    </label>
+                    <button id="closeDisclaimer" class="close-banner" onClick={this.handleClose}>&times;</button>
+                </div>
+            )
```
david-yz-liu, line 41: Add a semicolon here.

js/components/common/Disclaimer.js
```diff
@@ -0,0 +1,44 @@
...
+
+    render() {
+        const timetable = <a href="https://timetable.iit.artsci.utoronto.ca/">Official Timetable</a>;
+        const calendar = <a href="http://calendar.artsci.utoronto.ca/">Calendar</a>;
+
+        if (this.state.dontShowDisclaimer === 'true') {
+            return (null);
+        } else {
+            return (
+                <div id="disclaimerDiv" className="popup-banner">
```
david-yz-liu, line 31: Since you've added the classes, I suspect that the ids here and for the inner input/button elements are no longer necessary, and you can remove them.

js/components/grid/grid.js.jsx
```diff
@@ -164,8 +165,10 @@ class Grid extends React.Component {
 
   render() {
     const updatedList = this.state.hoveredLecture ? this.state.selectedLectures.concat(this.state.hoveredLecture) : this.state.selectedLectures;
+
```
david-yz-liu, line 168: Revert this change

style/app.scss
```diff
@@ -233,10 +233,51 @@ a:active
...
 }
 
-#disclaimerDiv
+.popup-banner
 {
-  padding   : 0px 1em 0px 1em;
-  font-size : 11pt;
+  font-size        : 11pt;
+  position         : absolute;
+  background-color : #ffdd61;
```
david-yz-liu, line 240: I'd prefer a yellow that's slightly less saturated (i.e., closer to white).

style/app.scss
```diff
@@ -233,10 +233,51 @@ a:active
...
+.popup-banner.dismiss
+{
+  transition : visibility 0s 0.5s, opacity 0.5s linear;
+  visibility : hidden;
+  opacity    : 0;
+}
+
+.popup-banner a
+{
+  color : #af16af;
```
david-yz-liu, line 254: Don't set a different colour here. Instead, use `text-decoration: underline` (in fact, we should have `text-decoration: underline` active by default for all `a` tags, but that's a different story).

style/app.scss
```diff
@@ -233,10 +233,51 @@ a:active
...
+}
+
+.popup-banner label
+{
+  font-weight : lighter;
+  color       : #696969;
+  position    : absolute;
```
david-yz-liu, line 261: Don't overuse absolute positioning. In this particular case, you shouldn't assume that there will be empty space in the bottom-right corner of the banner, as this depends on screen size. Use a different layout here. For example, put the "X" in the top-right corner, and the checkbox at the bottom.

style/app.scss
```diff
@@ -233,10 +233,51 @@ a:active
...
+  left        : 82%;
+  top         : 30px;
+}
+
+.close-banner
+{
+  cursor      : pointer;
```
david-yz-liu, line 268: This rule is redundant

style/app.scss
```diff
@@ -233,10 +233,51 @@ a:active
...
+}
+
+.popup-banner a
+{
+  color : #af16af;
+}
+
+.popup-banner label
+{
+  font-weight : lighter;
```
david-yz-liu, line 259: In general, do not introduce new styles for text or other elements unless there's a good reason we need something new. In this case, you should remove this rule and the new color.

david-yz-liu (commented):
In addition to the comments I left inline, I'd like you to experiment with different placements of this notification. I'm not sure a full-width banner is the best approach. You can try making the banner less wide and either centered at the top, centered at the bottom, or in a corner (I've seen bottom-left or bottom-right on a few websites). Possibly with a stronger border and/or rounded edges to make it more obvious that it's "on top of" the rest of the page.

leenlab2:
I fixed the issues you asked for, and changed the disclaimer design to look like this:
image

david-yz-liu (commented):
@leenlab2 very nice work! I just left a few more comments. Almost there!

leenlab2:
Alright, should be done now! Let me know if there is anything else I need to fix.

PR #1221, opened 2021-06-28
Graph Navigation Buttons Repositioning

kenneth-miura:
This pull request changes the positioning of graph navigation buttons, and moves the reset selection button to the graph, out of the side bar. This is part of Issue #1218.
## Your Changes

* Switched Reset View Button to use an image instead of text
* Moved Reset Selection button out of Side Bar
* Reformatted zoom in, zoom out buttons to be one over the other

- [x] Other: User Interface Change

## Testing

Manual Testing
## Checklist

  passed.


david-yz-liu (commented):
@kenneth-miura Looking at this a bit more, I think a good approach would be to keep the use of the `Button` component, but instead modify that component to use the special [`props.children`](https://reactjs.org/docs/composition-vs-inheritance.html). This should allow you to pass the image as the contents of the button.

kenneth-miura:
### Before
image
### After
image

style/app.scss
```diff
@@ -547,6 +547,14 @@ a:active
   color            : #9c9c9c;
 }
 
+.graph-control-button img {
+  width: inherit;
+  height: inherit;
+  position: absolute;
```
david-yz-liu, line 553: Similar to my comments to @leenlab2, avoid absolute positioning whenever possible. In this particular case, you should set a *margin* on the parent button component, and then width/height to `100%` here.

js/components/graph/__tests__/Graph.test.js
```diff
@@ -1,5 +1,5 @@
 import TestGraph from './TestGraph';
-import { fireEvent } from "@testing-library/react";
+import { fireEvent} from '@testing-library/react';
```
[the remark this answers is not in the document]
kenneth-miura: Are we supposed to be using double quotes over single quotes in import statements?

style/app.scss
kenneth-miura: What makes absolute positioning bad?

js/components/graph/__tests__/Sidebar.test.js
```diff
@@ -53,14 +53,6 @@ describe("Sidebar", () => {
     expect(container.getByText("FCE Count: 0.5")).toBeDefined();
   });
 
-  it("Clicking the reset button should reset the FCE Count to 0.0", async () => {
```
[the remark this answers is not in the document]
kenneth-miura: I miscommunicated with the team. We are planning to move the reset selection button into a new dialogue rather than add it as a graph button, so I reverted the changes so as to avoid unnecessary changes.

style/app.scss
leenlab2: From what I gather, it's because absolute positioning is akin to customizing which pixels the element is on rather than following the flow of the page. It might not look the same on different computers/screen sizes.

js/components/graph/__tests__/Graph.test.js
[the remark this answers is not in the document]
david-yz-liu: No, the different files are inconsistent on this actually. I'm planning to standardize this (as well as a few other style things) later on this summer.

style/app.scss
david-yz-liu: @kenneth-miura yeah, as @leenlab2 said, elements that use absolute positioning are hard to mix with elements that don't (and the latter is the default). They're essentially two different systems of positioning, and mixing the two can lead to unexpected results, especially as screen sizes change or you zoom in/out of the page.

In the case of this button image it doesn't *really* make a difference because the image is the only child of the button, but the default should be to avoid it.

kenneth-miura: Ok thanks, makes sense. So red flags for absolute positioing causing issues is when absolutely positioned elements are siblings to normal elements?

david-yz-liu: Yeah, I think that's a good way to think about it. I also saw this StackOverflow post with more nuanced answers: https://stackoverflow.com/questions/186044/is-it-considered-bad-practice-to-use-absolute-positioning/187310.

PR #1219, opened 2021-06-26
Menu bar formatting (stage 1)

leenlab2:
This pull request includes basic positioning and format changes to the menu bar.

## Motivation and Context

This pull request is part of a series that alters the layout of the Graph page. The details are outlined in Issue #1218.

## Your Changes

**Description**:

- Reduced size of the Courseography logo and altered position respectively
- Removed the black border around the menu bar and the navigation bar items
- Centered the navigation bar items
- Slightly reduced the size of navigation bar items and altered position respectively

**Type of change** (select all that apply):

- [x] Other: User Interface changes
## Testing

Manual testing in the browser.

## Questions and Comments (if applicable)

I wasn't really sure what type of change to label this with. Also, although I mentioned the issue I'm not sure how to link the PR to it. Finally, do we need to add tests for these layout changes?

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have verified that the CircleCI tests have
  passed.

PR #1223, opened 2021-06-29
moved export out of nav-links, added symbol

kenneth-miura:
## Motivation and Context
Addressing the section related to the export button on issue #1218.

## Your Changes
I altered `MasterTemplate.hs` to replace the Export element with an img. Then, I changed where it was defined so the img would be a direct child of `<nav class="row header">`. Then I added some styling so it would be in the top right corner.

**Description**:

- [x] Other: User Interface Change

## Testing

Clicked the new export button, and it still opens the export modal.

I made Haskell changes that I'm not confident in.
## Checklist

  passed.


david-yz-liu (commented):
@kenneth-miura related to our discussion of accessibility today: it would be good to wrap the image in a button as well, and also set the `alt` text attribute of the button to "Export".

kenneth-miura:
### Before
image

### After
image

app/MasterTemplate.hs
```diff
@@ -43,15 +43,16 @@ header page =
...
-            then H.li $ H.a ! A.id "nav-export" $ "Export"
-            else ""
             H.li $ toLink "/generate" "Generate (beta)"
             -- H.li $ toLink "/timesearch" "Search"
             -- H.li $ toLink "/draw" "Draw"
             -- TODO: re-enable after handling new first-year courses
             -- H.li $ toLink "post" "Check My POSt!"
             H.li $ toLink "/about" "About"
+        if page `elem` ["graph", "grid"]
+        then H.button ! A.id "export" $ do
```
david-yz-liu, line 53: The Haskell changes are indeed correct, but you should keep the old id (`nav-export`). It still fits, and in general it's best to avoid unnecessary changes.

style/app.scss
```diff
@@ -218,9 +218,19 @@ body
   padding-bottom : 3px;
 }
 
-#nav-export
+#export
 {
-  cursor : pointer;
+  cursor: pointer;
```
david-yz-liu, line 223: This rule is not necessary.

style/app.scss
```diff
@@ -218,9 +218,19 @@ body
...
+  position: absolute;
```
david-yz-liu, line 224: Similar to my comment on the other PR, you should try avoiding using absolute positioning.

This is a bit trickier to implement, especially after the nav bar changes in #1219 (which reminds, me you should update from master). A modern CSS approach is to use [**Flexbox**](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) on the nav bar (set `display: flex` and the `flex-direction: row`), and then have the three elements be the logo, the link `ul`, and the export button. After you get the general positioning right, you can use `text-align: right` on the button to make sure it's flush to the right.

PR #1229, opened 2021-07-16
Sidebar Redesign

EclipseIsDead:
## Motivation and Context

In response to #1218 this pull request is tackling the larger re-distributive changes of functionality and design, namely the side bar as well as menu bar changes.

## Your Changes

**Description**:
Removed old sidebar, and improved visibility for sidebar components. Redesigned sidebar to closer resemble mockup, including highlighted courses list.

As of right now, the course list within the side bar are all fully functional and allow you to interact with the graph. This works for any graph.

- [x] Other (please specify): User Interface change

## Testing

Manual testing in the web browser.

## Checklist


js/components/graph/Sidebar.js
```diff
@@ -1,204 +1,100 @@
 import React from "react";
 import PropTypes from "prop-types";
-import Focus from "./Focus";
 
 export default class Sidebar extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       contentHidden: true,
```
david-yz-liu, line 8: Okay, it looks like you have a good opportunity to simplify this component. Currently there are two states (`contentHidden` and `toggled`), which do essentially mean the same thing, and in fact it looks like `contentHidden` isn't really used.

Neither `contentHidden` or `toggled` are great names. Make a new state called `hidden` which is used to determine whether the sidebar is hidden or not, and update the rest of the code accordingly.

js/components/graph/Sidebar.js
```diff
@@ -1,204 +1,100 @@
...
       </div>
     )
   }
 
-  renderSidebarNav = () => {
-    const focusDisabled = this.state.focusDisabled ? "disabled" : "";
-    const focusActiveClass = this.state.graphActive === 0 ? "active" : "";
-    const graphActiveClass = this.state.graphActive === 1 ? "active" : "";
+  renderSearchBar = () => {
+    const searchHidden = this.state.toggled ? "" : "hidden";
```
david-yz-liu, line 45: You don't need this class here; it'll be enough to have the class only on the root sidebar component.

js/components/graph/Sidebar.js
```diff
@@ -1,204 +1,100 @@
...
+        <input type="submit" value="search"/>
+    </form>
     )
   }
 
-  renderSidebarButtons = () => {
-    const focusHiddenClass = this.state.graphActive === 1 ? "hidden" : "";
-    const graphHiddenClass = this.state.graphActive === 0 ? "hidden" : "";
+  renderCourses = () => {
+    const coursesHidden = this.state.toggled ? "" : "hidden";
```
david-yz-liu, line 61: See above comment

js/components/graph/Sidebar.js
```diff
@@ -1,204 +1,100 @@
...
+      <div id="courses" className={coursesHidden}>
+        {this.getCurrentCourses}
       </div>
     )
   }
 
   render() {
     const flippedClass = this.state.toggled ? "flip" : "";
     const sidebarClass = this.state.toggled ? "opened" : "";
+    const currBackGroundColor = this.state.toggled ? "#ffffff" : "";
```
david-yz-liu, line 73: Don't create style rules here. Instead, add these to the CSS.

david-yz-liu (commented):
@EclipseIsDead A good start. I left some general comments, and also you should always include screenshots in your PRs (even drafts) to highlight your changes.

app/MasterTemplate.hs
```diff
@@ -39,7 +39,7 @@ header :: T.Text -> H.Html
 header page =
     H.nav ! A.class_ "row header" $ do
         H.img ! A.id "courseography-header" ! A.src "/static/res/img/logo.png"
-             ! H.customAttribute "context" (textValue page)
+            ! H.customAttribute "context" (textValue page)
```
david-yz-liu, line 42: revert this change (in fact, it's no longer an issue on Master because of your other recent PR!)

js/components/graph/Sidebar.js
```diff
@@ -1,173 +1,137 @@
...
+      results: []
     };
   }
 
-  componentWillUpdate(prevProps) {
-    if (prevProps.graphName !== this.state.graphName) {
-      this.setState({ graphName: prevProps.graphName }, () => {
-        this.handleFocusEnabled();
+  /**
+   * If the active course props change, update the rendered courses that are in the sidebar.
```
david-yz-liu, line 15: Don't use state to store HTML. Instead, this can just be a method that returns the list of course `div`s. (You might worry that this function would be called at each render, which is true, but we can worry about the efficiency of this later, and solve it in a different way.)

js/components/graph/Graph.js
```diff
@@ -50,9 +50,10 @@ export default class Graph extends React.Component {
       infoBoxNodeId: "",
       panning: false,
       panStartX: 0,
-      panStartY:0,
+      panStartY: 0,
       showCourseModal: false,
-      showGraphDropdown:false
+      showGraphDropdown: false,
+      selectedNodes: []
```
david-yz-liu, line 56: Make this a `Set` instead of an `Array`. In terms of display order when rendering inside the sidebar component, you should convert the set to an array, and then sort the array.

js/components/graph/Graph.js
```diff
@@ -686,10 +696,23 @@ export default class Graph extends React.Component {
...
+        <Sidebar
+          ref={this.sidebar}
+          fceCount = {this.props.fceCount}
+          // I will need to update the graph name later once the other PR is done
+          graphName={this.state.graphName}
+          graphs={this.props.graphs}
+          reset={this.reset}
+          updateGraph={this.updateGraph}
+        />
+        <CourseModal
```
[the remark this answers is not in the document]
david-yz-liu: To be clear, what I mean is this component is not actually changed, you just added some new lines. So you should revert those line changes.

david-yz-liu (commented):
@EclipseIsDead I left some more comments. My overall comment is that this PR has gotten too large, and should be split up into two separate PRs: one that creates the new sidebar component that correctly *displays* the FCE count and list of selected courses, and then a separate one that introduces search functionality.

Remember that overall, you should always aim to keep PRs as lean as possible to make them easy to review, and so that they can be merged in quickly.

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  background-color : #cac4d4;
 }
 
 #container
 {
-  height   : calc(100% - 50px);
-  width    : 100%;
-  position : relative;
-  background: #f0f5f8;
+  height     : calc(100% - 50px);
```
david-yz-liu, line 712: revert these changes

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  padding-top      : 8px;
   text-align       : center;
-  padding-left     : 15px;
-  padding-top      : 5px;
-  font-size        : 18px;
+  font-size        : 16px;
 }
 
-#reset
+.label-hidden
```
david-yz-liu, line 677: It doesn't seem like this rule is used, so you should remove it

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  clip-path   : inset(50%);
+  height      : 1px;
+  overflow    : hidden;
+  position    : absolute;
+  white-space : nowrap;
+  width       : 1px;
 }
 
-#reset:hover
+#reset-selections
```
david-yz-liu, line 688: Make this a separate class, as you should avoid ids in CSS selectors whenever possible. Reuse existing styling from the `button` class as much as possible.

Convert the sidebar-related rules below to use classes instead of ids as well.

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
 {
-  background-color : #222222;
-  color            : #ffffff;
+  margin           : 10px;
+  text-align       : center;
+  float            : center;
```
david-yz-liu, line 692: Don't use `float` for any kind of alignment or positioning.

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
 
 #sidebar
 {
-  display          : inline-block;
-  width            : 40px;
-  height           : 100%;
+  border-radius    : 20px;
   float            : left;
-  background-color : #7a6a96;
   position         : absolute;
```
david-yz-liu, line 722: Use this opportunity to move away from absolute positioning. Instead, use flexbox (like your teammates have used for other components) to layout the components of the sidebar.

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
-  border           : solid 1px #000000;
+  margin-top       : 10px;
+  margin-left      : 60px;
+  text-align       : center;
+  z-index          : 2;
+}
+
+#selected-courses
+{
+  padding-top: 10px;
```
david-yz-liu, line 742: Fix the alignment for this rule

style/app.scss
```diff
@@ -644,7 +644,10 @@ a:active
 
 .opened
 {
-  width : 400px !important;
+  height           : 350px;
```
david-yz-liu, line 647: Use `max-height` instead of `height` for this rule, to allow the component to shrink/grow depending on its contents.

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  margin-left      : 60px;
+  text-align       : center;
+  z-index          : 2;
+}
+
+#selected-courses
+{
+  padding-top: 10px;
+  font-size        : small;
+  text-decoration  : underline;
```
david-yz-liu, line 744: Remove this

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  color            : #a47cb8;
+  background-color : none;
+}
+
+.courses
+{
+  overflow         : auto;
```
david-yz-liu, line 751: I think it should be `.sidebar` that has the overflow (though I haven't tried this out, you can reply with screenshots to show me what that looks like).

style/app.scss
```diff
@@ -654,68 +657,110 @@ a:active
...
+  z-index          : 2;
+}
+
+#selected-courses
+{
+  padding-top: 10px;
+  font-size        : small;
+  text-decoration  : underline;
+  color            : #a47cb8;
+  background-color : none;
```
david-yz-liu, line 746: Reminder that you can remove all `background-color` rules for the individual elements once you've implemented my above comment.

david-yz-liu (commented):
@EclipseIsDead great, now that the PR is more focused I left a more detailed review. In addition to my comments, I want to see more thorough test cases, testing both that multiple courses can be added, that courses can be removed as well, that the "Reset selections" button works as intended, and that the sidebar expand/collapse button works properly (you can check whether clicking it sets the right class).

style/app.scss
EclipseIsDead: Here is .sidebar:
Screen Shot 2021-07-22 at 12 54 16 PM
Compared to .courses:
Screen Shot 2021-07-22 at 12 55 26 PM

EclipseIsDead:
**Not toggled**
Before:
Screen Shot 2021-07-22 at 2 15 53 PM
After:
Screen Shot 2021-07-22 at 2 15 30 PM
**Toggled**
Before:
Screen Shot 2021-07-22 at 2 16 49 PM
After:
Screen Shot 2021-07-22 at 2 17 21 PM

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
-  list-style-type    : none;
-  display            : inline-block;
-  width              : 48%;
-  -webkit-transition : all 0.2s;
-  -moz-transition    : all 0.2s;
-  -ms-transition     : all 0.2s;
-  -o-transition      : all 0.2s;
-  transition         : all 0.2s;
+  overflow         : auto;
+  height           : 275px;
```
david-yz-liu, line 749: Use `max-height` here

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
+  width            : 105px;
```
david-yz-liu, line 750: You shouldn't be setting a separate width for any of the children. You can add `padding` to the parent element.

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
 }
 
-#sidebar-nav ul
+#selected-courses
 {
-  width  : 100%;
-  margin : 0 0 0 0;
+  margin-top       : 10px;
+  margin-bottom    : 10px;
+  font-size        : small;
```
david-yz-liu, line 742: Don't use `font-size: small`, set a specific size (this occurs below as well).

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
-  margin-top  : 25px;
-  margin-left : 25px;
-  height      : 86%;
-  overflow-y  : scroll;
+  border-radius    : 10px;
+  line-height      : 40px;
+  height           : 40px;
+  width            : 140px;
+  float            : left;
+  background-color : #ffffff;
```
david-yz-liu, line 779: Once again, you can delete this rule once you've implemented the other changes I've mentioned.

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
+  display          : inline-block;
 }
 
-#sidebar-nav ul li:hover
+#course-selection
 {
-  background-color : #5C497E !important;
+  width: 105px;
+  font-size: small;
+  text-align: center;
```
david-yz-liu, line 758: Delete this rule. I want the contents of the sidebar to be left-aligned.

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
+  height           : 275px;
+  width            : 105px;
+  display          : inline-block;
 }
 
-#sidebar-nav ul li:hover
+#course-selection
 {
-  background-color : #5C497E !important;
+  width: 105px;
```
david-yz-liu, line 756: Again, you can delete this.

style/app.scss
```diff
@@ -642,186 +642,168 @@ a:active
...
+  border-radius    : 10px;
+  line-height      : 40px;
+  height           : 40px;
+  width            : 140px;
+  float            : left;
+  background-color : #ffffff;
+  margin-top       : 20px;
+  text-align       : center;
+  font-size        : 16px;
+  visibility       : visible;
```
david-yz-liu, line 783: This is not necessary. It should only be the specific contents of the sidebar that are hidden where `visibility` (or `display` is set. This part is always visible, which CSS will do by default (no rules required).

david-yz-liu (commented):
@EclipseIsDead good job, I left a few more comments. Overall the Javascript code is almost done, but you need to do more cleaning up of the CSS. In particular:

1. There is duplication in the `.expanded` and `.collapsed` classes that should be moved into a shared class that is *always* on the root sidebar element. That way, `expanded` and `collapsed` only affect how subelements of the sidebar are displayed (or not).
2. Use classes, not ids, for the elements inside the sidebar.
3. Don't use `!important`.
4. I said in my previous review that I wanted the entire sidebar to have a background color, width, border-radius, and box-shadow set. You misunderstood and added this to the `#sidebar` rule, but that's not the entire sidebar, the outermost `div` is what I'm talking about. So (1) you should fix this, and (2) clearly `#sidebar` isn't a good name for a subelement of the sidebar, and you should change this when converting to a class.
5. Do not hide elements by setting their height and width to 0. Instead, use `display: none`.
6. Don't use `float`.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-#sidebar-nav ul li .nav-section
+.sidebar-button:hover
 {
-  color           : #000000;
-  display         : inline-block;
-  line-height     : 30px;
-  text-align      : center;
-  width           : 95%;
-  text-decoration : none;
+  background-color: #cac4d4;
```
david-yz-liu, line 722: This colour is too similar to the button text. Make the button text darker and this lighter.

style/app.scss
david-yz-liu: You missed this comment

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-.focus
+.fcecount
 {
-  display          : block;
-  cursor           : pointer;
-  font-size        : 20px;
-  border           : solid 1px #000000;
-  text-align       : center;
-  width            : 90%;
+  border-radius    : 10px;
```
david-yz-liu, line 733: This should be the same as the one for the root sidebar element. I like 10px better than 20px, so you can change the other one.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
+  box-shadow       : 2px 2px 2px #e8e9eb;
 }
 
-.focus:hover
+.reset-selections
 {
-  background-color : #dedede;
+  margin           : 10px;
+  text-align       : center;
+  width            : 90px;
```
david-yz-liu, line 745: Don't set a fixed width or height. Instead, use padding to add space outside the button text.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
+  height           : 20px;
+  cursor           : pointer;
```
david-yz-liu, line 747: This is not necessary, because this is automatically applied for `button` elements (is part of the built-in browser styles).

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
+  border-radius    : 15px;
+  font-size        : 10px;
```
david-yz-liu, line 749: This is too small. Make it 12

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
+  background-color : #ffffff;
```
david-yz-liu, line 750: Don't make this pure white. Instead, make it a different colour (and then the hover colour darker than that, and the text colour much darker than those).

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
 
 #sidebar
```
david-yz-liu, line 643: Make this a class as well.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-  height           : 100%;
-  float            : left;
-  background-color : #7a6a96;
-  position         : absolute;
-  padding-left     : 23px;
+  display          : flex;
+  flex-direction   : column;
+  z-index          : 1;
+  position         : inherit;
+  width            : 140px;
```
david-yz-liu, line 649: Overall the box is a bit small for my taste. Make this wider, and make the font for the contents be 14px rather than 12px.

js/components/graph/Sidebar.js
```diff
@@ -1,173 +1,72 @@
...
 }
 
 Sidebar.propTypes = {
-  currFocus: PropTypes.string,
   fceCount: PropTypes.number,
-  graphName: PropTypes.string,
-  highlightFocus: PropTypes.func,
   reset: PropTypes.func,
+  activeCourses: PropTypes.instanceOf(Set),
+  nodesJSON: PropTypes.array
```
david-yz-liu, line 71: You can delete this line

js/components/graph/Graph.js
```diff
@@ -708,9 +723,15 @@ export default class Graph extends React.Component {
...
       <div id="react-graph"
         className={reactGraphClass}
-        onClick={this.props.closeSidebar}
         {...reactGraphPointerEvents}
       >
+        <Sidebar
+          fceCount={this.props.fceCount}
+          reset={this.reset}
+          activeCourses={this.state.selectedNodes}
+          nodesJSON={this.state.nodesJSON}
```
david-yz-liu, line 732: You can remove this prop

js/components/graph/Graph.js
```diff
@@ -708,9 +723,15 @@ export default class Graph extends React.Component {
...
+          toggleZoom={this.toggleZoom}
```
david-yz-liu, line 733: This is a leftover prop that you should remove

js/components/graph/Sidebar.js
```diff
@@ -1,173 +1,72 @@
...
+        })}
       </div>
-    )
+    );
   }
 
   render() {
-    const flippedClass = this.state.toggled ? "flip" : "";
-    const sidebarClass = this.state.toggled ? "opened" : "";
+    const collapsedClass = this.state.hidden ? "collapsed" : "expanded";
```
david-yz-liu, line 49: Your change is okay, but missed the actual point of my previous comment: the **state** `hidden` should be renamed.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-  height           : 100%;
-  float            : left;
-  background-color : #5c497e;
-  position         : absolute;
-  border           : solid 1px #000000;
+  visibility : visible;
 }
 
-#sidebar-button:hover
+.expanded > .sidebar-dropdown
```
david-yz-liu, line 661: Change this selector to just `.sidebar-dropdown` (since you will be removing the height rule, see below comment).

david-yz-liu (commented):
@EclipseIsDead great, I went through and added some more comments. I reviewed the actual UI this time, which is why there are more comments specifically to make the component look better.

david-yz-liu:
@EclipseIsDead sorry I forgot to mention, you did remove a `text-align: center` rule, but the courses are still centre-aligned rather than left-aligned. (Probably inheriting this from an ancestor element.) You can set an explicit `text-align: left` for the courses.

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-  position         : absolute;
-  padding-left     : 23px;
+  display          : flex;
+  flex-direction   : column;
+  z-index          : 1;
+  position         : inherit;
+  width            : 140px;
+  background-color : #ffffff;
+  border-radius    : 20px;
+  box-shadow       : 2px 2px 2px #e8e9eb;
```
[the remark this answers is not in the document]
EclipseIsDead: I did the same for FCE count in order to meld the two more and make it look more cohesive as one piece rather than 2.
Screen Shot 2021-07-26 at 9 59 55 AM
Screen Shot 2021-07-26 at 10 00 10 AM

style/app.scss
```diff
@@ -687,131 +642,121 @@ a:active
...
-  position         : absolute;
-  border           : solid 1px #000000;
+  visibility : visible;
 }
 
-#sidebar-button:hover
+.expanded > .sidebar-dropdown
 {
-  background-color : #cac4d4;
+  height           : 350px;
```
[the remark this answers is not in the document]
EclipseIsDead: Updated Sidebar with max-height rule only (dynamic sizing):
Screen Shot 2021-07-26 at 10 03 31 AM

style/app.scss
EclipseIsDead: I'm assuming that this is actually in reference to the reset selections button, and not the sidebar button? However this could be me misunderstanding things again.

style/app.scss
EclipseIsDead: My apologies, I did not elaborate further. With respect to this comment, I tried removing it but this messed up the centring of the text vertically within the div. This is why I set line height to equal the height of the <div> itself as well as the relation between FCE and button because even while collapsed sidebar is a flex box. Should I rework the collapsed side of things to remove the need for line-height?
Screen Shot 2021-07-26 at 10 22 27 AM

EclipseIsDead:
Here is the left-aligned course version:
Screen Shot 2021-07-26 at 10 45 38 AM

style/app.scss
```diff
@@ -685,133 +640,133 @@ a:active
...
-  border           : solid 1px #000000;
-  text-align       : center;
-  width            : 90%;
+  border-radius    : 10px;
+  line-height      : 40px;
   background-color : #ffffff;
-  border-radius    : 6px;
+  text-align       : center;
+  font-size        : 16px;
+  box-shadow       : 4px 4px 4px #e8e9eb;
```
david-yz-liu, line 742: It's a bit strange to have the `box-shadow` set on the `.sidebar` root element when it's expanded, and the `.fcecount` element when collapsed. You can get the same effect by keeping the shadow only on the `.sidebar` element, when both expanded and collapsed.

The same is true for the `border` styling, I think it's better to just have that always be applied to the top `.sidebar` element.

style/app.scss
```diff
@@ -685,133 +640,133 @@ a:active
...
-  position         : absolute;
-  border           : solid 1px #000000;
+  display          : flex;
+  flex-direction   : column;
+  z-index          : 1;
+  position         : inherit;
+  width            : 160px;
+  background-color : #ffffff;
+  border-radius    : 10px;
+  box-shadow       : 4px 4px 4px #e8e9eb;
```
david-yz-liu, line 658: I think the dimensions are good but I'd prefer a dark shadow to a light one. Perhaps just use the same colour as the `graph-button` shadows?

david-yz-liu (commented):
@EclipseIsDead okay good job, I think I've taken what should be the final review on this branch before we get it merged in!

style/app.scss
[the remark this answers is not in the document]
david-yz-liu: Sure, you can reduce the top padding of the header a bit.

style/app.scss
david-yz-liu: Yes it was in reference to the reset selections button, but overall I wanted a greater contrast in both buttons, and you made good changes for both, so no need to make further colour changes.

style/app.scss
EclipseIsDead: The contrast between the white sidebar and reset selections makes the shadow look way darker than it already is so I went with the standard grey like the reset view button.
Screen Shot 2021-07-27 at 10 14 16 AM

PR #1231, opened 2021-07-19
Graph button styling

kenneth-miura:
## Motivation and Context
This addresses the button part of issue #1218. The goal is to modernize the graph control buttons

## Your Changes

**Description**:
I've removed the borders from the graph control buttons, and given them shadows. The zoom-in and zoom-out buttons are now connected. I also changed how the image in the reset view button is positioned, because the other changes made led to it getting uncentered.

- [x] User Interface Change

## Testing
Manual testing. Clicked the buttons and made sure they still worked as expected.

## Checklist

  passed. 

# Before
image

# After
image
## Button Closeup
image


style/app.scss
```diff
@@ -597,32 +597,23 @@ a:active
...
 #reset-view-button
 {
-  top   : 70px;
-  right : 30px;
+  top            : 70px;
+  right          : 30px;
+  display        : flex;
+  align-items    : center;
+  justify-content: center;
+  box-shadow: 1px 0px 5px 0px gray;
```
david-yz-liu, line 616: Align this one with the above rules

style/app.scss
```diff
@@ -2507,3 +2498,23 @@ main-filter input
...
+
+.zoom-button-group
+{
+  top              : 4px;
+  right            : 30px;
+  position         : absolute;
+  box-shadow       : 1px 0px 5px 0px gray ;
+  display          : flex;
+  flex-direction   : column;
+  background-color : rgb(140, 158, 140);
```
david-yz-liu, line 2510: This rule is redundant and can be removed

style/app.scss
```diff
@@ -2507,3 +2498,23 @@ main-filter input
   position         : absolute;
   padding          : 10px;
 }
+
+.zoom-button-group
+{
+  top              : 4px;
+  right            : 30px;
+  position         : absolute;
+  box-shadow       : 1px 0px 5px 0px gray ;
```
david-yz-liu, line 2507: delete the space before the `;`

style/app.scss
```diff
@@ -2507,3 +2498,23 @@ main-filter input
...
+  display          : flex;
+  flex-direction   : column;
+  background-color : rgb(140, 158, 140);
+  border-radius    : 6px;
+  overflow         : hidden;
```
david-yz-liu, line 2512: delete this rule

style/app.scss
```diff
@@ -2507,3 +2498,23 @@ main-filter input
...
+}
+
+#graph-zoom-button
```
david-yz-liu, line 2515: You can delete this rule altogether (see overall comments).

david-yz-liu (commented):
@kenneth-miura great start. I left some specific comments, but my biggest overall comment is that the structure of the HTML for the buttons can still be improved.

- There should be one div that encompasses *all* buttons, and this div should be the only element with an absolute positioning.
- It should have two children which are "button groups", where the first group is the zoom in/out buttons, and the other contains just the single "reset view" button. This class should be the one that has `border-radius: 6px`
- Then the individual buttons should all have the same class (`graph-control-button`) with other styling, but no ids are necessary.

kenneth-miura:
I also tried moving the buttons further down like you suggested. 

image

PR #1228, opened 2021-07-14
Fixed nav bar flexbox

leenlab2:
Added `div`'s to the `nav` such that the flexbox properly aligns the elements.

## Motivation and Context

A previous PR (#1223 ) added a flexbox to the nav bar, but the elements were not positioned properly since they were not in separate `div`'s. These changes are part of Issue #1218.

## Your Changes

**Description**:

- Added `div`'s to the Mastertemplate
- Fixed the logo's and export's divs to 250px width, and added `flex: 1` to the menu bar itself
- Used `text-align` to fix the alignment of the elements

**Type of change** (select all that apply):

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Manual testing in browser

## Questions and Comments (if applicable)

Before:
image

This is what it looks like now:
image

## Checklist

- [x] I have performed a self-review of my own code.
- [ ] I have verified that the CircleCI tests have
  passed. 
- [ ] I have added tests for my changes.

PR #1233, opened 2021-07-20
Linked Courseography icon to Courseography "home page"

EclipseIsDead:
## Motivation and Context

Typically when someone is on a website, they expect to be redirected to whatever the "home page" of the site is when clicking on the icon that represents said site. This PR addresses this concern from #1218.

## Your Changes

**Description**:
Wrapped the img in a `<a href="">` using Haskell.

- [x] User Interface Change

## Testing

Clicked the icon and was redirected correctly.

## Checklist

  passed.

### DIFF
--- a/app/MasterTemplate.hs
+++ b/app/MasterTemplate.hs
@@ -38,7 +38,7 @@ masterTemplate title headers body scripts =
 header :: T.Text -> H.Html
 header page =
     H.nav ! A.class_ "row header" $ do
-        H.a ! A.href "https://courseography.cdf.toronto.edu/graph" $ do
+        H.a ! A.href "/graph" $ do
             H.img ! A.id "courseography-header" ! A.src "/static/res/img/logo.png"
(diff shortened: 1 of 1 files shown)


app/MasterTemplate.hs
```diff
@@ -38,7 +38,8 @@ masterTemplate title headers body scripts =
 header :: T.Text -> H.Html
 header page =
     H.nav ! A.class_ "row header" $ do
-        H.img ! A.id "courseography-header" ! A.src "/static/res/img/logo.png"
+        H.a ! A.href "https://courseography.cdf.toronto.edu/graph" $ do
```
david-yz-liu, line 41: You should use a relative path, just like the other links in the nav bar. (This way the link will work on localhost too, for example.) So in this case it's just `/graph`.

david-yz-liu (commented):
@EclipseIsDead I left one comment, but also there's a merge conflict because of #1228 that you'll need to fix (should be a small one).

PR #1235, opened 2021-07-22
Graph dropdown fixes

kenneth-miura:
## Your Changes

**Description**: 
Fixes the issues from slack, namely:
* Generate page no longer loads
* down triangle removed from graph link
* Graph Dropdown appearing in wrong place

I also made a change in `GraphDropdown.js` so that the Graph Dropdown doesn't appear on hover if the user is in Generate mode.

## Testing
Manual testing by going over graph link, and going into generate mode.

I think it's confusing for the user that the graph link keeps the down arrow even when they are in generate mode, since it implies there should be a dropdown. Do you think it's worth removing the graph link's down arrow when in generate mode? I ask because the way I would implement that wouldn't adhere with React's philosophy. 

I'd check if there are any graphs provided to `GraphDropdown.js` in it's props, and use standard JS DOM selectors to change the text for the Graph link to either include or not include the down arrow. 

## Checklist

  passed.


app/MasterTemplate.hs
```diff
@@ -44,7 +44,7 @@ header page =
                     ! H.customAttribute "context" (textValue page)
         H.div ! A.class_ "nav-middle" $ do
             H.ul ! A.id "nav-links" $ do
-                H.li ! A.id "nav-graph" $ toLink "/graph" "Graph"
+                H.li ! A.id "nav-graph" $ toLink "/graph" "Graph▼"
```
david-yz-liu, line 47: RIght, don't add the arrow here, since it'll look a bit awkward and show up all the time. Instead:

1. You can do what you were suggesting in your question and in the `GraphDropdown` component add a *class* to this element.
2. In the css, use the `::after` pseudoelement that @leenlab2 was mentioning to add the arrow after the link text. This will give you more control over the styling and positioning of the arrow.

js/components/generate/generate.js
```diff
@@ -224,6 +224,7 @@ class GenerateForm extends React.Component {
     <Graph
       ref={this.graph}
       start_blank={true}
+      graphs={[]}
```
david-yz-liu, line 227: I think it's better to actually add this as a default prop to the `Graph` component. See the `defaultProps` class attribute.

js/components/graph/GraphDropdown.js
```diff
@@ -6,8 +6,8 @@ export default class GraphDropdown extends React.Component{
   render() {
     let className = "hidden";
     let graphTabLeft = 0;
-    if (document.querySelector("a[href='/graph']")){
-      graphTabLeft = document.querySelector("a[href='/graph']").getBoundingClientRect().left;
+    if (this.props.graphs.length !== 0 && document.getElementById("nav-graph")){
```
david-yz-liu, line 9: Oh I guess I missed this, but you can add a space before the `{`

js/components/graph/GraphDropdown.js
```diff
@@ -6,10 +6,19 @@ export default class GraphDropdown extends React.Component{
...
-    if (document.querySelector("a[href='/graph']")){
-      graphTabLeft = document.querySelector("a[href='/graph']").getBoundingClientRect().left;
-      if (this.props.showGraphDropdown){
-        className = "graph-dropdown-display";
+    if (this.props.graphs.length !== 0 && document.getElementById("nav-graph")) {
+      let navGraph = document.getElementById("nav-graph");
+      if (this.props.graphs.length === 0) {
+        navGraph.classList.remove("show-dropdown-arrow");
+      }
+      else {
```
david-yz-liu, line 14: Put this on the previous line

js/components/graph/GraphDropdown.js
```diff
@@ -6,10 +6,19 @@ export default class GraphDropdown extends React.Component{
...
+        if (!navGraph.classList.contains("show-dropdown-arrow")) {
+          navGraph.classList.add("show-dropdown-arrow");
+        }
+        if (this.props.showGraphDropdown){
```
david-yz-liu, line 18: Add a space before the `{`

style/app.scss
```diff
@@ -2524,3 +2524,8 @@ main-filter input
   position         : absolute;
   padding          : 10px;
 }
+
+.show-dropdown-arrow::after
+{
+  content : "▼"
```
david-yz-liu, line 2530: Add a semicolon here.

style/app.scss
```diff
@@ -2524,3 +2524,8 @@ main-filter input
   position         : absolute;
   padding          : 10px;
 }
+
+.show-dropdown-arrow::after
```
david-yz-liu, line 2528: Okay, so let's make this look a bit nicer by

1. adding some padding between the arrow and the link text
2. making the arrow a smaller font
3. making the arrow also change colour when the link is hovered over

kenneth-miura:
I implemented your suggested changes, but I found a bug with the generate page. I'll fix it in this PR if it's related to graph dropdown, so please do not merge yet. 

If I find it's not related to my graph dropdown changes, I think this PR should get merged, and then the bug fix should be a seperate PR. Is that a good idea, or should I make the fix in this PR regardless of it's origin?

kenneth-miura:
Ok it turns out I had run into the same bug as Leen had, stemming from an issue with the database. I replicated her fix and it worked. Since the database isn't version controlled, that means that this branch is good to merge if you think it's ok.

PR #1237, opened 2021-07-27
Updated React Testing Library to v12.0.0

leenlab2:
Updating the `@testing-library/react` in `devDependencies` to v12.0.0

## Motivation and Context

We have faced several errors when attempting to use the react testing library, and we are hoping that updating it would fix some of them.

## Your Changes

**Description**: I ran `yarn up @testing-library/react`, and a lot of files were changed but I'm not exactly sure why. I also changed a line in `TestGraph.js` from `await wait(...` to `await waitFor(...` since that is the current version of the wait function.

- [x] Other (please specify): package update

## Testing

Courseography still runs in the browser as expected, and all the tests on master pass.

There were a lot of cache files, so I'm not sure what happened there.

## Checklist

  passed.


david-yz-liu:
Thanks, @leenlab2! The Yarn cache stores versions of our Javascript dependencies, so it's expected that they would change for this PR.

PR #1238, opened 2021-07-30
Search bar for sidebar

EclipseIsDead:
## Motivation and Context
This pull request seeks to add a search bar function to the sidebar component, to quickly search and toggle course nodes without searching for them in the case that a node is difficult to find. This is addressed in #1218 more thoroughly.

## Your Changes

**Description**:
I have added a search bar and styled it to a degree (alongside the rendered results within sidebar). I have also made all items within the sidebar clickable to remotely toggle course nodes.

- [x] Test update (change that modifies or updates tests only)
- [x] User Interface Change

## Testing

I am in the process of writing up tests for this component but I have manually tested it in the browser.

There is an odd bug where typing in one character into the search bar will shift it to the right. Removing the CSS that centers it in the div fixes this but I think the problem is in the OnChange handling.
## Checklist


david-yz-liu (commented):
@EclipseIsDead nice work! A few quick comments:

1. In `Sidebar`, don't pass in `nodesJSON`, instead just pass in the ids (in other words, move the `.map` up to the `Graph` component.
2. `itemClick` isn't a very descriptive name, pick something that reflects what's actually being clicked
3. `htmlFor` connects by id, not class, so you should add the same `id` attribute to the search input.
4. I don't want *any* courses to appear if the search bar is empty
5. In terms of styling, the search dropdown might look better if it's overlaid on top of the course selection list, rather than being above it. (i.e., so that the dropdown looks like it's "coming out of" the input, like similar search dropdowns)

EclipseIsDead:
Screen Shot 2021-08-06 at 1 32 17 PM

EclipseIsDead:
I plan to finish rendering the dropdown over the courses and implementing the last test case either tonight or tomorrow morning.

js/components/graph/Sidebar.js
```diff
@@ -5,14 +5,25 @@ export default class Sidebar extends React.Component {
...
       this.setState({ collapsed: !this.state.collapsed });
   }
 
+  filteredSearch = (posts, query) => {
+    if (!query || !posts) {
+      return;
+    }
+
+    return posts.filter((post) => {
+      return post.includes(query);
```
david-yz-liu, line 23: Convert both post and query to upper case. Also, post is not a good name for this argument. Something like `course` would be a lot better.

style/app.scss
```diff
@@ -755,6 +760,38 @@ div.sidebar.collapsed
...
+{
+  display: none;
+}
+
+.search-bar
+{
+  margin : 0 auto;
+}
+
+.search-dropdown
```
david-yz-liu, line 773: Add a `list-style: none;` rule to this section, so that the "bullet" markers no longer appear

js/components/graph/Sidebar.js
```diff
@@ -5,14 +5,25 @@ export default class Sidebar extends React.Component {
...
+      collapsed: true,
+      results: []
     };
   }
 
   toggleSidebar = () => {
       this.setState({ collapsed: !this.state.collapsed });
   }
 
+  filteredSearch = (Query) => {
```
david-yz-liu, line 17: @EclipseIsDead Parameters should have lowercase names (`query`, not `Query`). Please change this one and `Course` below.

PR #1239, opened 2021-08-02
lined graph buttons up with sidebar

kenneth-miura:
## Motivation and Context
Positioning of graph control buttons looked strange because it was not in-line with the sidebar component.

## Your Changes

**Description**:Changed CSS so top graph control button is lined up with the top of the sidebar component.
Before:
image

After:
image

- [x] User Interface Change

## Testing

Visual check
## Checklist

  passed.

### DIFF
--- a/style/app.scss
+++ b/style/app.scss
@@ -1,3 +1,5 @@
+$graph-overlay-top-margin: 25px;
+$graph-overlay-side-margin: 15px;
 @-webkit-keyframes spin
 {
 0.0%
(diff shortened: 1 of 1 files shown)


david-yz-liu (commented):
@kenneth-miura nice. A couple of things:

1. This is a good time to actually try to use some of the SASS features we have at our disposal. In particular, try defining a [variable](https://sass-lang.com/documentation/variables) and using it in the both rules.
2. 10px is a bit small for my taste, maybe go with 15 or 20.
3. Do the same thing for the left/right margin on the objects, so they're the same horizontal distance from the edge of the screen.

kenneth-miura:
Increased top margin, made side margin for graph buttons same as side margin for the sidebar
image

PR #1241, opened 2021-08-03
Differentiate Hybrid and Course Nodes

kenneth-miura:
## Motivation and Context
Course Nodes & Hybrid course nodes are only differentiated by color. I wanted to make it clear graphically that the Hybrid nodes are not clickable, and that the Course nodes are clickable.

I searched online for advice on making elements look more clickable. I adapted advice around making buttons clearly recognizeable as clickable to the course nodes.

## Your Changes

**Description**:
I added a box-shadow for all the Course Nodes to add a sense of depth. I also made the hybrid nodes have sharp corners, and made the course nodes have more rounded edges than they already have. This was to differentiate them without making changes to the SVG.

Before:
image

After:
image

- [x] User Interface Change

## Testing

Tested generate tab by adding CSC207H1 manually. Did general testing for graph tab.
## Checklist

  passed.

### DIFF
--- a/js/components/graph/Node.js
+++ b/js/components/graph/Node.js
@@ -262,7 +262,18 @@ export default class Node extends React.Component {
     return (
       <g {...gAttrs} id={this.props.JSON.id_} className={newClassName} data-testid={this.getDataTestId()}>
         {ellipse}
-        <rect {...rectAttrs} style={rectStyle} />
+        <filter id="dropshadow" height="130%">
+          <feGaussianBlur in="SourceAlpha" stdDeviation="3"/>
+          <feOffset dx="2" dy="2" result="offsetblur"/>
+          <feComponentTransfer>
+            <feFuncA type="linear" slope="0.8"/>
+          </feComponentTransfer>
+          <feMerge>
+            <feMergeNode/>
+            <feMergeNode in="SourceGraphic"/>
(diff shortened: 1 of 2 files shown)


kenneth-miura:
New appearance after commit 7dca806. 
image

Made the changes based on my conversations with you and Leen during last week's meeting. Removing the borders further differentiates hybrids from course nodes.

js/components/graph/Node.js
```diff
@@ -262,7 +265,18 @@ export default class Node extends React.Component {
     return (
       <g {...gAttrs} id={this.props.JSON.id_} className={newClassName} data-testid={this.getDataTestId()}>
         {ellipse}
-        <rect {...rectAttrs} style={rectStyle} />
+        <filter id="dropshadow" height="130%">
```
david-yz-liu, line 268: This is cool! You should define the `filter` once at the root of the svg (in the `Graph` render); you should still be able to refer to it by id for each node, but this way it won't be duplicated once per node.

david-yz-liu: @kenneth-miura great thanks, I think you also need to update the test snapshots.

kenneth-miura: Ah sorry, didn't notice. tests are updated.

PR #1245, opened 2021-08-10
Sidebar course labels change

leenlab2:
Changed the retrieved course label on the "Selected Courses" list in the new sidebar from the `id_` of the nodes to their textual representation on the screen.

## Motivation and Context

Currently, the sidebar takes in the value of the node's `id` when listing the selections. This results in some nodes like the Calc 1 courses node to be represented as "mat135136137157", which is not very legible. Additionally, those "long" nodes sometimes have a simpler text label (e.g. "Calc 1") that is already listed on the SVG.

Also, on the Generate page, the nodes have odd ids, such as `CSC111_counter_6`. Following the SVG's text also makes it clearer for the user the correspondence between the sidebar and the selected nodes.

## Your Changes

**Description**:

- Changed the `selectedNodes` state in graph to store `node.text[].text` rather than `node.id_` (and fixed all relevant functions)
- Used the last item in the `node.text` array, so that nodes which have simpler representations (Calc 1, etc) will use those (this may be temporary until the SVG is updated later on)
- Added this sidebar functionality to the Generate tab, since earlier the Generate tab was not showing any selected courses at all

on Graph:
image

on Generate:
image

- [x] Test Update 

## Testing

The sidebar in Generate may not work properly until the FCE count change is merged in, as the sidebar functions rely on that.

## Checklist

  passed. 
- [x] I have added tests for my changes


leenlab2:
I added some changes to accommodate the search bar. This includes a temporary fix in `Graph.js` where Sidebar is rendered, the line `node.text.length > 0 ? .... : ""` should've just been `node.text[node.text.length - 1].text`, but due to the bug I mentioned where nodesJSON has an empty element. This empty "node" appears as the first element in nodesJSON on generate, even when a graph is there. It is returned by the fetch call, so the issue is in Haskell.

PR #1247, opened 2021-08-10
Hiding sidebar and buttons

leenlab2:
Hide the sidebar and buttons in Generate when no graph has been generated yet.

## Motivation and Context

Having the sidebar and graph panning/zooming controls available before the graph itself is there feels unintuitive.

## Your Changes

**Description**:
- Conditionally render sidebar and buttons when nodesJSON has more than 1 entry (this is because on Generate, invalid entries still result in a JSON that has 1 entry)
- fixed a resulting bug where the queryselector calling .sidebar continues even if the sidebar isnt there

## Testing

I am not entirely sure why there is still a nodeJSON entry when the graph is empty?

## Checklist

  passed.

### DIFF
--- a/js/components/graph/Graph.js
+++ b/js/components/graph/Graph.js
@@ -86,7 +86,9 @@ export default class Graph extends React.Component {
         .addEventListener("mouseleave", this.hideGraphDropdown);
     }
 
-    document.querySelector(".sidebar").addEventListener("wheel", (event) => event.stopPropagation())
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@leenlab2 thanks! I'm not sure why `nodesJSON` might have length 1 for invalid graphs. You could look into this a bit more to see what the response is from the server, and again the issue might be on the Haskell side of things.

PR #1248, opened 2021-08-10
Temporary fix for duplicate nodes

leenlab2:
Added a filter to remove duplicate nodes from occurring.

## Motivation and Context

On certain graphs, such as `(unofficial) Statistics` and `(unofficial) Finnish`, there are "ghost nodes" that persist even after changing graphs. This is because there are duplicate nodes on top of hybrid nodes. This issue does not exist in the original SVG files, but does in the resulting database. Since tinkering with Haskell might be a bit complicated for now, this PR contains a front-end by-pass.

image

## Your Changes

**Description**:
Since the duplicates carry on in the `nodesJSON` state of Graph, I filtered it by creating a duplicate-free copy of `nodesList`. The code we use to populate `this.state.connections` already iterates over `nodesJSON`, so I just filtered by whether that node was already checked/ added to `connections`. 

Before:
image

After:
image

## Testing

At first, I was going to add a more detailed filter that checks if there exists a hybrid node in the same position, but this basic filter seemed to do the job. A more detailed filter may require more runtime, and as this is a temporary fix anyway, I decided to go with the simpler fix.

Should I add the haskell fix as an issue?

## Checklist

  passed.


david-yz-liu (commented):
@leenlab2 the fix is good, I just have two overall comments:

1. It would be good to leave a comment in the code explaining why we're doing this check.
2. Use a set rather than a list, since the former supports constant time lookup. 😄

leenlab2:
@david-yz-liu I made it into a list since `nodesJSON` and the other `JSON` states were initialized as lists. Since `noDuplicatesNodesList` is later assigned to the `nodesJSON` state, doesn't it make more sense to be a list?

david-yz-liu:
@leenlab2 ah yes, you're totally right, I think I read this too quickly last night. Thanks for clarifying!

PR #1260, opened 2022-05-15
Add program requirement parser

ChingChang9:
## Motivation and Context
The program requirements are currently parsed as RAW text. We want to parse them to do more awesome things with them in the future :)

## Your Changes

**Addendum:**
- A `programParser` that parses anything until a sentence terminator (eg. eof, `.`), an or separator or a degree keyword (eg. "major", "program")
- A `programGroupParser` that parses a group of programs using `programParser`. Programs need to be parsed in groups so that sentences like "CS or Math major or specialist" can be parsed and reconstructed as ("CS major" or "Math major" or "CS specialist" or "Math specialist")
- A `programOrParser` that parses groups of programs related through an OR clause
- Test cases!
- **New: Generate a node for each program requirement in the auto-generated graph. The display might need more work (screenshots in discussion below)**

**Amendment:**
- Append more orSeparators and andSeparators to the existing lists so program requirements and course requirements can be joined together with and/or clauses
- Allow `courseOrProgParser` (previously `courseParser`) to try matching program requirements before falling back to the RAW parser

## Testing

Below are the test cases I tried. Only the ones that passed (checked) are added. Reasoning of the each test is in parantheses after the test string. Let `X`, `Y` and `Z` be some program names.

| What we're testing | Test String | Added? Why not
|-|-|-|
basic | "Admission to X" | ✅
one program with two degree types to concatenate (major and specialist) | "Admission to X Major or Specialist program" | ✅
X on its own, Y concatenated with 2 degree types, connected with a raw text through an or clause | "Enrolment in the X program or in a Y major or specialist program, or permission of instructor" | ✅
2 programs to be concatenated (Y and Z, but not X) | "Enrolment in the X program or in a Y or `Z` major or specialist program" | ✅
make sure "enrolment in a ..." doesn't slice the "A" in "ASMAJ1618". Related to Raw text through an and clause | "Enrolment in ASMAJ1618. A student must be in third or fourth year." | ❌ Parses correctly but trims the second sentence. The version on master misparses it as `GRADE "A" (J "CSC148H1" "")`
program before course | "Enrolment in the X program, and PSY309H1, and one of PSY319H1/ PSY329H1/ PSY339H1" | ✅
program after course | "MGT100H1 or equivalent, or enrolment in the X Specialist or Major"  | ~~❌ "MGT100H1 or equivalent, or CSC148H1" doesn't work either~~ <br /> **Update: I added this test back but without the " or equivalent" so we can still test the case where the program requirement comes after the course. However, there are currently no instances where this occurs in the timetable without " or equivalent"**
multiple programs with nothing to concatenate; program reqs after FCEs | "At least 8.0 credits, and enrolment in a science, mathematics, or engineering program" | ~~❌ Works on its own without the frontal FCEs reqs, but "At least 8.0 credits, and CSC148H1" doesn't work either~~ <br /> **Update: we can consider adding this back after merging #1261**
program before FCEs | "Enrolment in an X Program of Study; completion of at least 1.0 credit of 200-level ESS courses." | ~~❌ Parses program correctly but takes FCEs as RAW. Fails because the current parser doesn't handle lowercase "completion"~~ <br /> **Update: The parser on master uses either FCEs or andParser, but not both. So this happens for all cases where there are both FCEs and courses, but should probably be fixed in a different PR?**
program after FCEs | "ENV222H1/ ( TRN140Y1, TRN141Y1), 12.0 credits and enrolment in the X." | ~~❌ x.x credits and ... doesn't work with courseID either~~ <br /> **Update: Same as above**
grade before program | "70% in SOC212H1 and enrolment in X program" | ✅
parentheses | "(70% in SOC212H1 and enrolment in X program)" | ✅
should keep it as RAW despite the "admission to" | "Instructor's permission required for admission to course" | ✅

- All the test cases that failed failed because of some pre-existing shortcoming in the existing parsers, not because of this change (error persisted after replacing the program substring with a course code). As a result, I'm not sure if I should fix them in the same PR. I think I know how to fix some of them though.
- To concatenate the programs with their degree type, I had to use a nested case because I couldn't find a way to pattern match on two variables at once. I'm not sure if it's the most elegant way of doing it.
- `progGroupSeparator` separates programs within a group, and `progOrSeparator` separates groups of program. They are both very similar to `orSeparator` but the things they parse are crucially different, so I think we should create them even though they may seem like duplicate code. I've checked and confirmed that no program at UofT has " or " in its name.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,8 @@ Betty Wang,
 Fullchee Zhang,
 Minfan Zhang,
 Alex Shih,
-Cassandra Stefura
+Cassandra Stefura,
+Zi Kai Xu
 
 ## Privacy Policy and Licensing
 
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ This project would not exist without the contributions of many students in the D
 Alex Baluta,
 Alexander Biggs,
 Kelly Bell,
+Ching Chang,
 Christina Chen,
 Eugene Cheung,
 Kael Deverell,
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -51,13 +51,11 @@ degreeType = do
     return degree
 
 programSuffix :: Parser String
-programSuffix = do
-    space <- Parsec.space
-    suffix <- Parsec.choice (map caseInsensitiveStr [
+programSuffix = Parsec.spaces
+    >> Parsec.choice (map caseInsensitiveStr [
         "program of study",
         "program"
         ])
-    return $ space:suffix
 
 fceSeparator :: Parser ()
 fceSeparator = Parsec.choice (map (Parsec.try . Parsec.string) [
@@ -116,10 +114,7 @@ progGroupSeparator = do
 
 progOrSeparator :: Parser String
 progOrSeparator = do
-    separator <- Parsec.choice [
-        Parsec.space >> caseInsensitiveStr "or in a",
-        caseInsensitiveStr ", or"
-        ]
+    separator <- Parsec.space >> caseInsensitiveStr "or in a"
     space <- Parsec.space
     return $ separator ++ [space]
 
@@ -375,7 +370,7 @@ andParser = do
 programGroupParser :: Parser Req
 programGroupParser = do
     progs <- Parsec.sepBy (Parsec.try programParser) progGroupSeparator
-    degrees <- Parsec.sepBy (Parsec.try degreeType) (Parsec.string "or")
+    degrees <- sepByNoConsume (Parsec.try degreeType) orSeparator
     _ <- Parsec.optional $ Parsec.choice $ map Parsec.try [
         degreeType,
         programSuffix
@@ -396,7 +391,6 @@ programOrParser = do
     Parsec.spaces
     _ <- programPrefix
     progs <- Parsec.sepBy (Parsec.try programGroupParser) progOrSeparator
-    _ <- Parsec.many $ Parsec.noneOf ".;,)"
     case progs of
         [] -> fail "Empty Req."
         [x] -> return x
@@ -417,6 +411,16 @@ fcesParser = do
 -- | Parser for requirements separated by a semicolon.
 categoryParser :: Parser Req
 categoryParser = Parsec.try fcesParser <|> Parsec.try andParser
+
+-- Similar to Parsec.sepBy but stops when sep passes but p fails,
+-- and doesn't consume failed characters
+-- Modified from https://hackage.haskell.org/package/parsec-3.1.15.1/docs/src/Text.Parsec.Combinator.html#sepBy
+sepByNoConsume :: Parser String -> Parser String -> Parser [String]
+sepByNoConsume p sep = (do
+    x <- p
+    xs <- Parsec.many $ Parsec.try (sep >> p)
+    return (x:xs))
+    <|> return []
 
 -- Flattens nested ORs into a single OR
 -- eg. OR [OR ["CS major, "Math major"], RAW "permission from instructor"]
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -385,7 +385,7 @@ programOrParser = do
     Parsec.spaces
     _ <- programPrefix
     progs <- Parsec.sepBy (Parsec.try programGroupParser) progOrSeparator
-    _ <- Parsec.many $ Parsec.noneOf ".;,"
+    _ <- Parsec.many $ Parsec.noneOf ".;,)"
     case progs of
         [] -> fail "Empty Req."
         [x] -> return x
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -3,8 +3,8 @@ module WebParsing.ReqParser where
 
 import Data.Char (isSpace, toLower, toUpper)
 import Database.Requirement
-import Text.Parsec ((<|>))
 import qualified Text.Parsec as Parsec
+import Text.Parsec ((<|>))
 import Text.Parsec.String (Parser)
 
 -- define separators
@@ -25,6 +25,27 @@ completionPrefix = Parsec.choice (map (Parsec.try . Parsec.string) [
     "Completion of"
     ])
     >> Parsec.spaces
+
+-- TODO: should only use lowercase admission to not match
+-- "Instructor's permission required for admission to course"
+-- but the rest should be case insensitive
+programPrefix :: Parser ()
+programPrefix = Parsec.choice (map caseInsensitiveStr [
+    "admission to",
+    "enrolment in the",
+    "enrolment in an",
+    "enrolment in a",
+    "enrolment in"
+    ])
+    >> Parsec.spaces
+
+programSuffix :: Parser String
+programSuffix = Parsec.spaces
+    >> Parsec.choice (map caseInsensitiveStr [
+    "major",
+    "specialist"
+    ])
+
 
 fceSeparator :: Parser ()
 fceSeparator = Parsec.choice (map (Parsec.try . Parsec.string) [
@@ -63,6 +84,14 @@ andSeparator = Parsec.choice $ map caseInsensitiveStr [
     "&",
     "+",
     "plus"
+    ]
+
+programOrSeparator :: Parser String
+programOrSeparator = Parsec.choice $ map caseInsensitiveStr [
+    ",",
+    " or in a",
+    " or a", -- frontal space to prevent matching "history"
+    " or "   -- trailing space to prevent matching "organization"
     ]
 
 oneOfSeparator :: Parser String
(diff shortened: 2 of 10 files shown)


ChingChang9:
Unrelated to this particular PR: I did not have `hlint` and `stylish-haskell` required for the pre-commit even though I followed through the install guide. Maybe we need to add
```
cabal install hlint
cabal install stylish-haskell
```
in the installation guide in some future commit?

ChingChang9:
Since the program name can be long, the text often overflows the box (image 1), which gets even worse when there are multiple programs (image 2). We can maybe allow word-wrap in the text nodes so the the texts can span over multiple lines. But maybe that belongs to a different PR?

#### Image 1
image 

#### Image 2
image

david-yz-liu (commented):
@ChingChang9 great work! I will take a closer look in a bit, but for now one thing you can do is split off your changes to the `andSeparator` and `orSeparator` specifically into a separate pull request, and add test cases for the existing and/or parsers. And I'll probably merge that one in first, so you can add yourself to the changelog in that branch. :)

Also for the `andSeparator`, add `; and` as another option.

ChingChang9:
I wasn't able to truncate the texts with `text-overflow: ellipsis;`, so I increased the size of the nodes from the backend. It turned out the width of each node was automatically calculated by the `Data.GraphViz.Attributes.Complete` library using the default `NodeSize GrowAsNeeded` option, which calculated the node sizes inaccurately when there were spaces in the node label. The course nodes didn't have spaces so they were fine, but the program nodes had spaces and so their sizes were inaccurate.

The magnitude of the error was not proportional to the number of spaces in the node label, which was very strange since a long string with no space had less error than a short string with one space, which didn't always have more error than another string with more than one space, so it wasn't something about the way the width of the spaces was calculated. I was unable to figure out the exact cause of this but figured that padding the node label with `n` spaces where `n` is the length of the program name (not including the "Enrolled in ") was the best approximation, and the text would still get displayed correctly because we trimmed the trailing spaces in the actual display. Very quirky workaround but I'm not sure what else to do :(

## Other things I tried
- I thought it could be the font family messing up the approximation but specifying the font family to the ones defined in `Css.Constants` and in the ones in the frontend didn't seem to fix anything
- I tried ditching the default `GrowAsNeeded` option and passed in the width/height manually but all nodes would be as wide as the widest node. Manually calculating the minimum width for each node is difficult since we aren't using a monospace font

## What I ended up changing
- Padding the program label with spaces
- Specifying the `GrowAsNeeded` option for `NodeSize`. This is unnecessary because it's the default option but making it explicit might help future readers (took me a while to figure out what the code was doing)
- Specifying the font size to use for the node width calculation. Default is 14. We are using 12, so we should specify

## Some previews
image
image
image

## Other issue for the future
- Hovering over program nodes and clicking on "info" should perhaps show the program info instead of the info of the last course we clicked on

app/WebParsing/ReqParser.hs
```diff
@@ -26,6 +26,32 @@ completionPrefix = Parsec.choice (map (Parsec.try . Parsec.string) [
     ])
     >> Parsec.spaces
 
+programPrefix :: Parser ()
+programPrefix = Parsec.choice (map caseInsensitiveStr [
+    "admission to",
+    "enrolment in the ",
+    "enrolment in an ",
+    "enrolment in a ", -- trailing space to prevent matching "enrolment in A..."
```
david-yz-liu, line 34: This is an excellent point. Really, though, *each* of the phrases in this category should end with a space (even `admission to`). So rather than add a space manually to these phrases, you should parse a single space character *after* this part.

This could look like an extra `>>` before the `Parsec.spaces`, or you can use [`skipMany1`](https://hackage.haskell.org/package/parsec-3.1.11/docs/Text-Parsec.html#v:skipMany1) combined with `isSpace` to check for 1 or more spaces instead of 0 or more.

app/WebParsing/ReqParser.hs
```diff
@@ -26,6 +26,32 @@ completionPrefix = Parsec.choice (map (Parsec.try . Parsec.string) [
...
+    "enrolment in"
+    ])
+    >> Parsec.spaces
+
+degreeType :: Parser String
+degreeType = do
+    Parsec.spaces
+    degree <- Parsec.choice $ map caseInsensitiveStr [
+        "major",
+        "specialist"
```
david-yz-liu, line 44: Make sure to add `minor` here as well, and tests for it.

app/WebParsing/ReqParser.hs
```diff
@@ -26,6 +26,32 @@ completionPrefix = Parsec.choice (map (Parsec.try . Parsec.string) [
...
+        ]
+    Parsec.spaces
+    return degree
+
+programSuffix :: Parser String
+programSuffix = Parsec.choice $ map caseInsensitiveStr [
+    " program of study",
```
david-yz-liu, line 51: See comment above about spaces. Try to find a different way of checking for spaces other than inserting it into the string literal.

README.md
```diff
@@ -83,6 +83,7 @@ This project would not exist without the contributions of many students in the D
 Alex Baluta,
 Alexander Biggs,
 Kelly Bell,
+Ching Chang,
```
david-yz-liu, line 86: This is showing changes that have now been merged in from your other PR. You should update this branch by doing a merge from `upstream master`, and these changes will no longer appear. 👍

app/WebParsing/ReqParser.hs
```diff
@@ -65,6 +93,19 @@ andSeparator = Parsec.choice $ map caseInsensitiveStr [
     "plus"
     ]
 
+progGroupSeparator :: Parser String
+progGroupSeparator = Parsec.choice $ map (Parsec.try . Parsec.string) [
+    ",",
+    " or a ", -- frontal space to prevent matching "history"
```
david-yz-liu, line 99: see above comment about spaces

app/WebParsing/ReqParser.hs
```diff
@@ -247,46 +288,107 @@ justParser = do
...
 
+-- | Parser for programs grouped together
+-- | Parses program names and degree types, then concatenate every combination
+-- | eg. (CS or MAT major) implies (CS major) or (MAT major)
+-- | (CS major or specialist) implies (CS major) or (CS specialist)
+-- | (CS or MAT major or specialist) implies (CS maj) or (MAT maj) or (CS spec) or (MAT spec)
+programGroupParser :: Parser Req
+programGroupParser = do
+    progs <- Parsec.sepBy (Parsec.try programParser) progGroupSeparator
+    degrees <- Parsec.sepBy (Parsec.try degreeType) (Parsec.string "or")
```
david-yz-liu, line 365: The `or` literal here is a bit strange; seems like this needs to be combined with some of the other parsers you defined above for "or" separators.

app/WebParsing/ReqParser.hs
```diff
@@ -247,46 +288,107 @@ justParser = do
     markInfoParser = do
         Parsec.try (fmap Left percentParser <|> fmap Left letterParser <|> fmap Right infoParser)
 
--- parse for single course with or without cutoff OR a req within parentheses
-courseParser :: Parser Req
-courseParser = Parsec.between Parsec.spaces Parsec.spaces $ Parsec.choice $ map Parsec.try [
+-- Parses for single course or a group of programs with or without cutoff OR a req within parentheses
+-- Programs need to be parsed in groups because of the concatenation issue
+-- explained in the docstring of `programGroupParser`
+courseOrProgParser :: Parser Req
```
david-yz-liu, line 294: Don't change `courseParser`, it also has a good logical function. Instead you can create a new parser that tries either `programOrParser` or `courseParser`. This will keep the different parsers easier to follow.

app/WebParsing/ReqParser.hs
```diff
@@ -247,46 +288,107 @@ justParser = do
...
+        degreeType,
+        programSuffix,
+        progGroupSeparator,
+        progOrSeparator,
+        Parsec.oneOf ".;" >> return "",
+        Parsec.eof >> return ""
+        ]
+    return $ PROGRAM program
+
+-- Flattens nested ORs into a single OR
```
david-yz-liu, line 316: This is a helper function. Move it to the bottom of the file (after all of the parsers).

app/WebParsing/ReqParser.hs
ChingChang9: The problem I had here was that the degree types were only ever separated by `or`s, and other separators in `orSeparator` and `programOrSeparator` would overmatch. For example, the `, or` in `orSeparator` would split "Enrolment in a History major, or permission of instructor" (because we're using `Parsec.sepBy`) and would try to read the "permission of instructor" as a degree type, which would then fail because it is not one of "major", "specialist", "program"... etc.

Should I define a `degreeOrParser` (similar to `programOrParser`) that matches only `or`s and reference it here to make it more clear?

app/WebParsing/ReqParser.hs
ChingChang9: Minor doesn't appear anywhere in the timetable. Should we still add it?

I think that it is not by luck that we don't have "minor"s in the timetable, because every requirement that requires a minor would also accept a major/specialist, so the timetable would just list them as "program". For example, instead of saying "Enrolment in the CS specialist, major, or minor", the timetable would just say "Enrolment in the CS program". So it seems like "minor" will never appear?

I can still add it of course. Just double checking if we should add it :)

david-yz-liu: Yes you should add it, it's good for future-proofing. In principle there could be courses that are accessible only to students in a minor program; we must be careful not to overfit to the existing calendar.

app/WebParsing/ReqParser.hs
david-yz-liu: In general we shouldn't rely on different separators like this. The parser should also work on `Enrolment in a History major or permission of instructor`. (See above comment about overfitting.)

Seems like the problem is failing parsers consuming characters we don't want to be consumed. (i.e., the `or` being consumed even if the `degreeType` parser fails.) If this is the case, then it seems like we'll need to do something more sophisticated. You can take a look at [how `sepBy` is implemented](https://hackage.haskell.org/package/parsec-3.1.15.1/docs/src/Text.Parsec.Combinator.html#sepBy) to get some ideas about implementing something similar, perhaps.

david-yz-liu (approved):

PR #1267, opened 2022-05-29
Make fcesParser compatible with andParser

ChingChang9:
## Motivation and Context
The FCES parser has lower precedence than and/or parsing, so it currently can't be combined with other requirements. For example, "Completion of 9.0 credits or permission of the instructor" is parsed as `FCES "9.0" (RAW "or permission of the instructor")` rather than `OR [FCES "9.0" (RAW ""), RAW "permission of the instructor"]`

## Your Changes
- Moved `fcesParser` into `courseOrProgParser`, which is passed into `andParserOf` (originally `andParser`; see next bullet point) so the FCEs, courses, and programs can all be treated the same and parsed as ANDs of ORs
- Refactored `andParser :: Parser Req` that parsed ANDs of ORs of `courseOrProgParser` into a more general function `andParserOf :: Parser Req -> Parser Req` that takes in a parser and returns a parser that parses ANDs of ORs of the given parser
  - This is because `fcesParser` uses an andParser for courses (not including programs and rawText) but `categoryParser` uses an andParser for all categories
- Added `fcesRawParser`, a special rawParser for fcesParser that parses like `rawTextParser` but with more specific constraints (some unnique separators from or/andSeparators and strictly not followed by certain texts)

## Testing
| What we're testing | Test String
|-|-|
`OR [FCE "..." (RAW ""), RAW "..."]` | "Completion of 9.0 credits or permission of the instructor"
`AND [FCE "..." (RAW ""), RAW "..."]` | "Completion of 9.0 credits. Permission of the instructor"
`OR [FCE "..." (RAW "..."), RAW "..."]` | "At least one 0.5 credit at the 400-level or permission of the instructor"
`FCE "..." (RAW "...")` where there's an "or" in the RAW that should not be split (because it's "or higher") | "1.0 credits at the 300-level or higher"
Same as above but it's "or [program name]" | "1.0 credit in HPS or HIS"
Multiple FCEs | "1.0 credits or 1.0 credit in Canadian Studies"
Multiple FCEs with andParser and rawText | "1.0 credits, CSE240H1/ NEW240Y1, CSE342H1, an additional 0.5 credit at the 300+ level from the Critical Studies"

- I realized that `courseOrProgParser` now also includes `fcesParser` so I'd like to rename the variable. I am thinking of renaming it to `categoryParser` and renaming the original `categoryParser` to `reqParser`, but I haven't done that in this PR because I thought it would be confusing to review. Let me know if I should change it or if there are better names
- I remember what you said about reusing the separators and not overfitting to the current timetable, but I feel that the `fcesTerminator` deserves a separate definition instead of reusing and/orSeparator because it accepts many substrings in the and/orSeparators. For example, `"from a 400-level CSC/BCB course"` should not be parsed as `OR [RAW "a 400-level CSC", RAW "BCB course"`. The same problem exist with `.`, `;`, `+` as well so I thought it'd be cleaner to create a new separator

## Checklist


david-yz-liu (commented):
@ChingChang9 I took a quick look at this and the other PR. Overall good work, and here are a couple of quick comments:

1. In the other PR you should be adding test cases for all changes. Remember that the test cases can be made up, and do not have to match exact entries in the academic calendar.
2. You can perform the renaming of `reqParser` and `categoryParser` you proposed in a separate PR (good idea, out of scope here).
3. In general for this type of parsing, we want to try to specify *what to parse* rather than *when to stop parsing*---this translates to avoiding things like `Parsec.notFollowedBy` and instead building up parsers that will naturally parse what they can. Of course, this is made trickier in our case by the fact that the `RAW` parser will consume everything.

    So, we have two options. Drop the `RAW` parsing for `FCES` and just stick with something like `andParserOf courseParser` and add to it over time, or keep the `RAW` parsing as a last resort and either let it go until the end of the string or impose a hard stop (e.g. the `fcesOrTerminator` you created).

    To keep things simple for the purposes of this PR, let's go with the second option, but stop whenever a `andSeparator` or `orSeparator` is encountered. (This will lead to some parsing errors like the ones you described in the condition, but those will be fixed by creating better parsers for those cases, not by modifying the separators.)

    Then, to make the structure of the `fcesParser` clearer, you can change it to something like

    ```haskell
    fcesParser :: Parser Req
    fcesParser = do
        _ <- Parsec.optional completionPrefix
        fces <- creditsParser
        _ <- Parsec.spaces
        _ <- fceSeparator
        modifiers <- fcesModifiersParser
        return ...
     ```

     where `fcesModifiersParser` tries both the `and/or courseParser` and the `fcesRawParser`. That way, the future work will be adding just to `fcesModifiersParser`.

4. Finally, you should be trying to minimize all changes you make. So, move the `and` and `or` parsers back to where they were (in general do not reorder existing functions), leave them all top-level, and don't rename them. But it's okay to add additional parameters to them to generalize them, as you did here.

ChingChang9:
Hi David, thank you for the detailed feedback. For clarification on your third point, am I understanding it correctly that you'd like me to replace the `fcesOrTerminator` I wrote with the existing `orSeparator` and `andSeparator` and accept the parsing errors in this PR, then create another PR to create parsers for strings like "a 300/400-level course", "CSC/BCB courses", "the 300+ level from History", and then add those parsers to `fcesModifiersParser` so it tries them before falling back to `fcesRawParser`?

app/WebParsing/ReqParser.hs
```diff
@@ -405,12 +411,30 @@ fcesParser = do
     _ <- Parsec.spaces
     _ <- fceSeparator
     _ <- Parsec.optional $ Parsec.try includingSeparator <|> Parsec.try fromSeparator
-    req <- Parsec.try andParser <|> Parsec.try orParser
-    return $ FCES fces req
+
+    FCES fces <$> fcesModifiersParser
```
ChingChang9, line 415:
```hs
modifiers <- fcesModifiersParser
return $ FCES fces modifiers
```
might be easier to read but the haskell linter was telling me to write it this way

david-yz-liu: yes that's fine

app/ParserTests/ParserTests.hs
```diff
@@ -71,18 +71,25 @@ fcesInputs = [
...
+    , ("Completion of 9.0 credits. Permission of the instructor", AND [FCES 9.0 (RAW ""), RAW "Permission of the instructor"])
     , ("Completion of at least 9.0 FCE", FCES 9.0 $ RAW "")
     , ("Completion of a minimum of 4.0 FCEs", FCES 4.0 $ RAW "")
     , ("Completion of a minimum of 9 FCEs", FCES 9.0 $ RAW "")
     , ("Completion of 4.0 credits", FCES 4.0 $ RAW "")
     , ("at least 4.0 credits", FCES 4.0 $ RAW "")
-    , ("at least one additional 0.5 credit from a 300-/400-level course.", FCES 0.5 $ RAW "a 300-/400-level course.")
-    , ("at least 1.0 additional credit from 300-/400-level CSC/BCB courses", FCES 1.0 $ RAW "300-/400-level CSC/BCB courses")
-    , ("at least one 0.5 credit from a 400-level CSC/BCB course", FCES 0.5 $ RAW "a 400-level CSC/BCB course")
+    --, ("at least one additional 0.5 credit from a 300-/400-level course.", FCES 0.5 $ RAW "a 300-/400-level course.")
```
david-yz-liu, line 81: Don't comment these out. If the actual value changed, you can update the test to show the change.

david-yz-liu (commented):
@ChingChang9 good work, and please also do a merge from master.

app/ParserTests/ParserTests.hs
ChingChang9: The actual value of some tests do not make sense since the modifiers are not implemented yet. For example, `at least one additional 0.5 credit from a 300-/400-level course.` is parsed as `OR [FCES 0.5 (RAW "a 300-"),RAW "400-level course."]`.

I resolved this by removing these tests completely and adding simpler tests. The tests removed are documented in my notes and shared with Ben, which we will add back in future PRs.

For the tests with values that make sense, I uncommented them 👍🏻

PR #1271, opened 2022-06-04
Add modifier data type for FCEs reqs

ChingChang9:
## Motivation and Context
The current Req data structure does not represent FCEs accurately. We want to introduce a modifier data type for the FCEs modifiers

## Your Changes
- Changed the second parameter type of the FCEs constructor from `Req` to `Modifier`
- Added the `Modifier` data type with a `REQUIREMENT` constructor that takes a `Req`
  - NOTE: The Modifier _data type_ is declared as a `newtype` instead of a `data` in this PR because HLint was failing here since we only had one constructor (`REQUIREMENT`). Once we add the other constructors in future PRs (eg. level, department), we will be able to change it to `data`
- Updated the test cases from `FCES Float Req` to `FCES Float (REQUIREMENT Req)`

**Description**:

## Testing
- Ran the tests in `app/ParserTests/ParserTests.hs` and ensured they all passed
- Generated a graph with FCEs requirements and ensured it looked fine
  - image
  - image (wow 😮; see comment below)

Neither of AND and OR makes sense in cases like this. If we use AND, all 4 nodes need to be selected. If we use OR, we can get away with only 0.5 FCEs. We migjt need a new checking mechanism but I feel this is out of the scope of this PR?

## Checklist

### DIFF
--- a/app/Database/Requirement.hs
+++ b/app/Database/Requirement.hs
@@ -7,14 +7,18 @@ Module containing data type that represents a "Requirement".
 We will use parsed data to create instances of this type.
 -}
 
-module Database.Requirement
-( Req(..) ) where
+module Database.Requirement (
+    Req (..),
+    Modifier (..)
+    ) where
 
 data Req = NONE
          | J String String
          | AND [Req]
          | OR [Req]
-         | FCES Float Req
+         | FCES Float Modifier
          | GRADE String Req
          | PROGRAM String
          | RAW String deriving (Eq, Show)
+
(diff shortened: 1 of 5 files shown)


david-yz-liu (approved):

david-yz-liu:
@ChingChang9 you're absolutely correct that we need a better way to display FCES requirements from a list of courses. But that has to do with the rendering, not the parsing, so please make a note of this and we'll handle it later. :)

PR #1273, opened 2022-06-05
Department modifier

ChingChang9:
## Motivation and Context
The current parser on master parses department codes in FCEs requirements as RAW text (eg. the "CSC" "1.0 credits in CSC"). It's better to have a department constructor for it.

## Your Changes
- Added `departmentParser` for when the andParser for courses fails. It matches any character until "course(s)", "credit" and other similar words, or/and separator, or the end of line
- Added a `many1Till` parser that is similar to `Parsec.manyTill` but matches at least one character. Used in `departmentParser` so we don't get an empty string
- ~~Added `deptBefFcesParser` to parse fces requirements where the department name comes before the fces (eg. "1.0 CSC credits")~~
- ~~Renamed `fcesParser` to `plainFcesParser` and added a new `fcesParser` that tries `plainFcesParser` before trying `deptBefFcesParser`~~
  - ~~I considered the option of calling the new function something other than `fcesParser` to avoid the renaming in this PR but realized I would need to replace places where `fcesParser` is called with the new function anyway, so I renamed it in the end~~

## Testing
Why | Test string
-|-
Simple department modifier | "0.5 credit in HPS"
Department modifier followed by "courses" | "1.0 credit in MST courses and 0.5 credit in HIS"
Department modifier before fcesSeparator | "2.0 ENG credits"
UI | image

## Checklist


app/WebParsing/ReqParser.hs
```diff
@@ -402,10 +402,22 @@ programOrParser = do
         [x] -> return x
         (x:xs) -> return $ OR $ flattenOr (x:xs)
 
+-- | Parser for FCE requirements following the dept name
+-- | eg. "... 9.0 CSC FCEs ..."
+deptBefFcesParser :: Parser Req
```
david-yz-liu, line 407: You don't need to create a separate parsing function here. Use an [`optionMaybe`](https://hackage.haskell.org/package/parsec-3.1.15.1/docs/Text-Parsec.html#v:optionMaybe) in the `fcesParser` to attempt to parse a department after the credit number, and then you can use an `if` expression to branch on whether that parsing succeeded or failed.

ChingChang9:
Removed 4 tests temporarily since the department parser now parses until the end of line, and/or separator, or fce/course separator. This means the levels are being parsed as departments which doesn't make sense. This will be fixed in #1275 when I add `levelParser` above `departmentParser`, making `fcesModifierParser` parse for level first, and only parse for department when level fails.

- "At least one 0.5 credit at the 400-level"
- "At least 1.5 credits at the 400-level"
- "NEW240Y1, an additional 0.5 credits at the 300 level from the Critical Studies"
- "At least one 0.5 credit at the 400-level or permission of the instructor"

app/WebParsing/ReqParser.hs
```diff
@@ -408,15 +408,20 @@ fcesParser :: Parser Req
 fcesParser = do
     _ <- Parsec.optional completionPrefix
     fces <- creditsParser
+    dept <- Parsec.optionMaybe $ Parsec.try departmentParser
     _ <- Parsec.spaces
     _ <- fceSeparator
     _ <- Parsec.optional $ Parsec.try includingSeparator <|> Parsec.try fromSeparator
-    FCES fces <$> fcesModifiersParser
+    modifiers <- fcesModifiersParser
```
ChingChang9, line 415: I can move this to the `Nothing` clause but it's fine here because in #1275, I will do some pattern match with `modifiers` to append department to level

david-yz-liu: You should move it. Code changes shouldn't be predicated on future PRs.

app/ParserTests/ParserTests.hs
```diff
@@ -82,11 +82,10 @@ fcesInputs = [
     , ("Completion of a minimum of 9 FCEs", FCES 9.0 $ REQUIREMENT $ RAW "")
     , ("Completion of 4.0 credits", FCES 4.0 $ REQUIREMENT $ RAW "")
     , ("at least 4.0 credits", FCES 4.0 $ REQUIREMENT $ RAW "")
-    , ("At least one 0.5 credit at the 400-level", FCES 0.5 $ REQUIREMENT $ RAW "the 400-level")
```
david-yz-liu, line 85: Don't remove the tests. If they currently produce `DEPARTMENT` values, that's fine.

app/WebParsing/ReqParser.hs
```diff
@@ -425,6 +430,30 @@ courseAsModParser = do
     req <- andParser courseParser
     return $ REQUIREMENT req
 
+-- | Parses a literal "course" or "courses"
+courseLiteralParser :: Parser String
+courseLiteralParser = Parsec.spaces
```
david-yz-liu, line 435: Don't include the spaces here. It's fine if the spaces are parsed as part of the `dept` below, you can strip them out when creating the `DEPARTMENT` value

app/WebParsing/ReqParser.hs
```diff
@@ -425,6 +430,30 @@ courseAsModParser = do
...
+
+-- | Parses a department code in the fces requirement
+-- | eg. the "CSC" in "1.0 credit in CSC" or "1.0 credit in CSC courses"
+departmentParser :: Parser Modifier
+departmentParser = do
+    Parsec.spaces
+    Parsec.notFollowedBy fceSeparator
+    dept <- many1Till Parsec.anyChar $ Parsec.choice (map Parsec.try [
+        courseLiteralParser,
+        Parsec.spaces >> Parsec.lookAhead fceSeparator >> return "",
```
david-yz-liu, line 449: Same comment about spaces here as above. Also I'm not sure why `lookAhead` is needed here but not elsewhere.

app/WebParsing/ReqParser.hs
```diff
@@ -425,6 +430,30 @@ courseAsModParser = do
...
+        "courses",
+        "course"
+        ])
+
+-- | Parses a department code in the fces requirement
+-- | eg. the "CSC" in "1.0 credit in CSC" or "1.0 credit in CSC courses"
+departmentParser :: Parser Modifier
+departmentParser = do
+    Parsec.spaces
+    Parsec.notFollowedBy fceSeparator
```
david-yz-liu, line 446: Overall this and the next part are a bit too complicated. You can use `Parsec.manyTill`, and then explicitly fail if an empty string (or a string containing only whitespace) is consumed. You can use the `fail` function here this is technically part of the `MonadFail` type class, which `Parser` is a member of).

PR #1274, opened 2022-06-05
Parse "any" in fces requirements

ChingChang9:
## Motivation and Context
- "Any 3.0 credits" was not being parsed
- The "any subject" in "3.0 credits in any subject" is redundant and can be ignored when parsing the fces modifier
- The lists of `fromSeparator`s and `completionPrefix`es are not exhaustive

## Your Changes
- Added "any" in the list of fces `completionPrefix`es
- Added a parser for "any subject" and "any field"
- Added "a" in `completionPrefix`
- Added "of", "from the", and "at the" in `fromSeparator`

## Testing
- "Any 9.0 credits"
- "9.0 credits in any field"
- "9.0 credits in any subject"
- Replaced "the x00-level" in RAW requirements to "x00-level" (which will be replaced with LEVEL in #1275)
- Test strings including "of" and "a "

## Checklist

### DIFF
--- a/app/WebParsing/ReqParser.hs
+++ b/app/WebParsing/ReqParser.hs
@@ -33,8 +33,7 @@ completionPrefix = Parsec.choice (map (Parsec.try . caseInsensitiveStr) [
     "Any",
     "a"
     ])
-    >> Parsec.space
-    >> Parsec.spaces
+    >> Parsec.skipMany1 Parsec.space
 
 programPrefix :: Parser ()
 programPrefix = Parsec.choice (map caseInsensitiveStr [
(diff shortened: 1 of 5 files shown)


app/WebParsing/ReqParser.hs
```diff
@@ -437,6 +445,17 @@ rawModifierParser = do
         ]
     return $ REQUIREMENT $ RAW text
 
+-- | Parses "any field" or "any subject" in an fces modifier since they are redundant
+anyModifierParser :: Parser String
```
david-yz-liu, line 449: Change the type into `Parser ()`. We don't need to return the redundant string.

PR #1276, opened 2022-06-07
add cGPA parser and cGPA parser tests

benxu3:
## Motivation and Context

There are certain programs with requirements for a minimum cGPA. The cGPA parser will parse these strings in to a GPA constructor (which is a Req type). 

## Your Changes

- implemented cGPA parser 
- wrote tests for cGPA parser 

**Description**:

## Testing

Using a script, we compiled different cases of Strings from the FAS academic calendar where there was a requirement for minimum cGPA. Then from these cases I broke the strings in to appropriate substrings for the cGPA parser to parse. I wrote the expected output from the cGPA parser for these substrings and then parsed them using the cGPA parser to get the desired output. Then, I added these tests to the ParserTests.hs file as cgpaInputs and included cgpaTests to the reqTestSuite. Running these tests produces no errors. 

If I understand the feedback correctly, I will be changing the Req argument of the GPA constructor to be a String. I think this would require me to first fit the cGPA parser in to the structure of other parsers such that it is composable with other functions as a Req for AND and OR lists. 

Then for cases such as "A minimum CGPA of 1.0, completion of 10.0 credits and permission of the College Program Director" instead of having GPA 1.0 (AND [FCES 10.0 (RAW ""), RAW "permission of the College Program Director"]) be the output we can expect AND [GPA 1.0 "", FCES 10.0 (RAW ""), RAW "permission of the College Program Director"] to be the output instead. 

## Checklist


david-yz-liu:
> If I understand the feedback correctly, I will be changing the Req argument of the GPA constructor to be a String. I think this would require me to first fit the cGPA parser in to the structure of other parsers such that it is composable with other functions as a Req for AND and OR lists.
>
> Then for cases such as "A minimum CGPA of 1.0, completion of 10.0 credits and permission of the College Program Director" instead of having GPA 1.0 (AND [FCES 10.0 (RAW ""), RAW "permission of the College Program Director"]) be the output we can expect AND [GPA 1.0 "", FCES 10.0 (RAW ""), RAW "permission of the College Program Director"] to be the output instead.

@benxu3 yes that's exactly right

PR #1286, opened 2022-07-20
refactor: rename constructors to PascalCase

ChingChang9:
## Your Changes

Replaced all mentions to constructors (including the comments) with the PasalCase version of their name.
Only exceptions are the "cGPA" in the `cgpaPrefix`, to reduce unnecessary changes. It should not affect the program because all string literals are case insensitive by the implementation.

## Testing
- Ran `stack test` and everything still passed
- Ran `stack run` and ensured that the site is still functional

[PascalCasing for acronyms seems to be opinion-based](https://stackoverflow.com/q/15526107/10254049). I am following [this suggestion](https://stackoverflow.com/a/15526526/10254049).
So `FCES` and `GPA` were changed to `Fces` and `Gpa`.

## Checklist


app/DynamicGraphs/GraphNodeUtils.hs
```diff
@@ -11,56 +11,56 @@ import Database.Requirement (Modifier (..), Req (..))
...
         Just xs -> " from " ++ concatModor xs
 
     in show creds ++ deptFormatted ++ deptsFormatted ++ " FCEs"
         ++ levelFormatted ++ levelsFormatted ++ rawFormatted ++ rawsFormatted
 
--- | Formats a MODOR into FCEs string
--- | Assumes all modifiers in the MODOR have the same constructor
+-- | Formats a ModOr into FCEs string
+-- | Assumes all modifiers in the ModOr have the same constructor
 formatModor :: Float -> [Modifier] -> String
```
david-yz-liu, line 49: Right, so this function should be renamed `formatModOr` (and similar with the `concatModor` function below)

david-yz-liu (approved):

PR #1287, opened 2022-07-20
Zero width space tests

ChingChang9:
## Important Update
I misidentified the source the error. The real source of the `None` error is explained in #1288. The html processing library actually handles the zero-width spaces already, so this PR might be entirely useless now 😢. I'll keep it open in case we want to handle the zero-width spaces from the `reqParser` nonetheless, but feel free to close this.

## Motivation and Context
All requirements containing a zero-width space are being parsed as `None` since `reqParser` cannot handle zero-width spaces.

## Your Changes
- Replaced all instances of `Parsec.space` and `Parsec.spaces` with the newly defined `space` and `spaces` that handle spaces and zero-width spaces

- [x] Test update (change that modifies or updates tests only)

## Testing
- Put `reqParser` into a Main module (`app/RequirementTests/ReqParserMain.hs`) so we can compile it for testing
- Added `app/RequirementTests/zeroWidthSpaces.sh` that compiles `app/RequirementTests/ReqParserMain.hs`, tests it against `app/RequirementTests/zeroWidthSpacesTests`, and exit with 0 if all tests passed or 1 otherwise
- Ran `app/RequirementTests/zeroWidthSpaces.sh` in `app/RequirementTests/tests.hs` and made it exit with the exit code of the shell script

Screenshots:
Cases | Screenshot
-|-
Passed | image
Failed 1 | image
Failed 1+ | image

You said to try Python for testing in the meeting, but I decided to use Bash in the end because we need to compile the Haskell program using `ghc` and run the resulting executable, which to my understanding, require a shell process. I thought it would be cleaner to do everything in a shell script instead of spawning another process in Python.

## Checklist


david-yz-liu (commented):
@ChingChang9 thanks for the update (and good work originally finding this issue!). Zero-width space characters might pop up and it would be good for the parser to handle it, though it doesn't seem like this will be a blocker now.

However, the thing with the zero-width space characters is they can appear anywhere, not just in between words. So even something like `CSC<zw-space>108H1` could look fine when reading the calendar but not actually be parsed correctly. So to make the parser "handle" zero-width spaces, we should probably just do some pre-processing on the input text itself to remove all such characters *before* parsing the text.

Also in terms of adding test cases, you found the Unicode escape `\8203` for the space character---you can use this in string literals as well, and ghc should handle this just fine. (So in other words, you can write a test case with the input string `"CSC108H1/\8203 CSC148H1"`.) No need for the separate bash test script.

ChingChang9:
Closing since this is already implemented

https://github.com/Courseography/courseography/blob/5c1ed07348777739f7f45ef89c9540fb8ddbbdcc/app/WebParsing/ArtSciParser.hs#L175-L179

PR #1289, opened 2022-07-23
Post parsing

ChingChang9:
## Motivation and Context
The code and description of post aren't being stored in the database. We want to update their parser and write them in the database.

## Your Changes
- Updated `postInfoParser` to parse the post code, name, and department
- Stored the code and raw requirement description into the database

## Testing
- Created `app/RequirementTests/PostParserTests.hs` and added it to `app/RequirementTests/tests.hs`
- Renamed `app/RequirementTests/ParserTests.hs` to `app/RequirementTests/ReqParserTests.hs`
- Manually inspected every input-output pair after running `stack run database`
- Ensured `stack run database` puts as many rows into the database as the number of posts on art&sci's list of posts

- `app/WebParsing/PostParser.hs` imports the `trim` helper function from `app/WebParsing/ReqParser.hs`. Would it be better if I moved the `trim` function (and perhaps other helper functions) to a `app/WebParsing/ParserUtils.hs` and imported it in both files?
- Should I revert d56dbaa for readability now that we have the `OverloadedString` extension, or keep it for more accurate types?
- For the following cases:

Index | Input | Ideal output
:-:|:-|:-
1 | "CS Major: A" | "CS: A"
2 | "CS (Major: A)" | "CS (A)"
3 | "CS (Major) A" | "CS A"

(1) suggests `postTypeParser` shouldn't parse the `:`. (2) suggests the parser needs to parse `:` when it is inside of `()`. To know whether it is inside of `()`, the parsing for `()` needs to be in `postTypeParser`. But putting it in `postTypeParser` means `()` is not parsed by `postNameParser`, so the output won't contain the `()`. If we try to parse it from `postNameParser` and pass a boolean of whether the last character is `(` into `postTypeParser`, then we will need to drop the `(` from `postNameParser` when concatenating the outputs together depending on whether it is case (2) or (3), and there is no way of knowing that from `postNameParser`. My final approach was to always parse the `:` following the post type, so the output of (1) would be "CS A".

## Checklist

### DIFF
--- a/app/WebParsing/PostParser.hs
+++ b/app/WebParsing/PostParser.hs
@@ -55,12 +55,12 @@ postInfoParser = do
         void postCodeParser,
         P.eof
         ]
-    code <- postCodeParser P.<|> return ""
+    code <- postCodeParser P.<|> return (T.pack "")
 
     case deptNameBef of
-        "" -> return $ Post postType (T.pack $ trim deptNameAft) code ""
-        xs | last xs == '(' -> return $ Post postType (T.pack $ trim $ trim deptNameBef ++ trim deptNameAft) code ""
-           | otherwise -> return $ Post postType (T.pack $ trim $ trim deptNameBef ++ " " ++ trim deptNameAft) code ""
+        "" -> return $ Post postType (T.pack $ trim deptNameAft) code $ T.pack ""
+        xs | last xs == '(' -> return $ Post postType (T.pack $ trim $ trim deptNameBef ++ trim deptNameAft) code (T.pack "")
+           | otherwise -> return $ Post postType (T.pack $ trim $ trim deptNameBef ++ " " ++ trim deptNameAft) code (T.pack "")
 
 postTypeParser :: Parser PostType
 postTypeParser = P.try postNameParser P.<|> P.between (P.char '(') (P.char ')') postNameParser
--- a/app/WebParsing/PostParser.hs
+++ b/app/WebParsing/PostParser.hs
@@ -6,9 +6,9 @@ module WebParsing.PostParser
 import Control.Monad.Trans (liftIO)
 import Data.Either (fromRight)
 import Data.Functor (void)
-import Data.List (find, intercalate)
+import Data.List (find)
 import Data.List.Split (keepDelimsL, split, splitWhen, whenElt)
-import Data.Text (strip)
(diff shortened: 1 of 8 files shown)


app/WebParsing/PostParser.hs
```diff
@@ -30,36 +32,64 @@ addPostToDatabase programElements = do
...
                 Nothing -> return ()
     where
         isRequirementSection tag = tagOpenAttrNameLit "div" "class" (T.isInfixOf "views-field-field-enrolment-requirements") tag || tagOpenAttrNameLit "div" "class" (T.isInfixOf "views-field-field-completion-requirements") tag
 
-
 -- | Parse a Post value from its title.
 -- Titles are usually of the form "Actuarial Science Major (Science Program)".
 postInfoParser :: Parser Post
 postInfoParser = do
-    deptName <- parseDepartmentName
```
david-yz-liu, line 47: Okay overall this parsing is getting too complicated. This was appropriate for the prerequisite parsing because that's all over the place in the calendar, but we really don't need this here.

Let's switch strategies and have the post name be the full name up to the final `" - "`, and the post code everything after that (which is the current behaviour). Then, parse the post type from the post code: the text from chars 2-4 map to the post type, `SPE` -> `Specialist`, etc.

ChingChang9: I implemented this, but there are five cases where the post string doesn't have a post code[^1]. In those cases, I made the parser fall back to look for the post type in the department name. Hope it is not too complex!

[^1]: "Focus in Green Chemistry"<br>"Focus in Practical Chemical Analysis"<br>"Focus in Data Science in Business"<br>"Focus in Managing in Diverse Economies (UofT Global Scholar)"<br>"Biological Physics Specialist"

ChingChang9:
In addition to the changes you mentioned, I also added a `requirements` column in the `Post` table in the database. The raw requirements were stored under the `description` column before. Now, they are stored under the `requirements` column, and the `description` column stores the description.
image

The implementation uses `partitions` and matches both the description and the requirements, so that they are split into different arrays, and then takes the first array, to filter out the requirement text.

david-yz-liu (approved):
@ChingChang9 excellent work, the code is very clean and the description of your changes was clear. :+1:

PR #1290, opened 2022-07-31
Resolve type errors between Text and String

ChingChang9:
## Motivation and Context
There are some places in the code where the function is expecting `Text` but receiving `String`, and places where type definitions are not allowed. These errors are somehow not thrown by ghc when compiling the main executable and so the current code on master compiles fine. However, #1289 adds test cases to the post parser, which requires `app/WebParsing/PostParser.hs` and all its dependencies with errors to be added to the test suite in `courseography.cabal`. These errors show up when running `stack test`.

## Your Changes
- (Un)packed `String`/`Text` to `Text`/`String` where needed
- Added `{-# LANGUAGE ScopedTypeVariables #-}` in `app/WebParsing/PostParser.hs` to allow type signatures in do blocks

## Testing
- Ran `stack test` and ensured everything still worked.
- Applied the changes to #1289 locally and ran `stack test` to ensure every error was fixed, and every test passed

## Checklist

### DIFF
--- a/app/Database/Tables.hs
+++ b/app/Database/Tables.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE DataKinds, DeriveGeneric, DerivingStrategies, EmptyDataDecls, FlexibleContexts,
              FlexibleInstances, GADTs, GeneralizedNewtypeDeriving, MultiParamTypeClasses,
-             QuasiQuotes, StandaloneDeriving, TemplateHaskell, TypeFamilies,
+             QuasiQuotes, ScopedTypeVariables, StandaloneDeriving, TemplateHaskell, TypeFamilies,
              UndecidableInstances #-}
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@ChingChang9 oh that's a good catch, though the solution is simpler than what you've done. We in `courseography.cabal` the main executable has a `default-extensions` field that automatically loads GHC extensions when compiling the files; that field is missing from the "test" executable. You can fix the issue you're discussing by adding these default extensions to that section of the cabal configuration.

(BTW you might want to read up on the `OverladedStrings` extension, it's very related to the changes you were making.)

ChingChang9:
Oh my... The three lines of code resolved hours of work. Thank you for redirecting me to the proper fix!

Closing this as the fix is now implemented in #1289.

PR #1291, opened 2022-08-01
Use focus data

ChingChang9:
## Your Changes

**Type of change** (select all that apply):
- [x] Refactoring (internal change to codebase, without changing functionality)

## Checklist

- [ ] I have performed a self-review of my own code.
- [ ] I have verified that the CircleCI tests have passed. 
- [ ] I have added tests for my changes.

PR #1296, opened 2022-08-12
feat: store post html instead of text

ChingChang9:
## Your Changes
- Added a `pruneHtml` function that removes all the attributes from the html except the "style" attribute. It also trims the `<a>` tags
- Ran `pruneHtml` and `renderTags` (converts tags to html) before putting the data into the database

## Testing
- Tested `pruneHtml` when there is
  - an `<a>` tag
  - a style attribute
  - a class attribute
  - a tag with both a class attribute and a style attribute
- Manually inspected some rows in the `post` table of the database after parsing and ensure things were correct
- Applied the changes to #1291 and ensured the modals looked pretty

I did not create a column to store the raw HTML and I could not find one from the current schema. The current schema has a column that stores the post requirements as plain text. I am replacing the data in that column with the pruned HTML. The raw text is still extracted and parsed, but they are no longer in the database (but the parsed data is).

I decided to trim the `<a>` tags because they are all links to the course page on the art&sci calendar, which show the exact same information we have on the graph when the user clicks on the nodes.

The test file looks a little empty to have an entire for it now but I'm thinking it can be used for #1287 too

## Checklist

### DIFF
--- a/app/RequirementTests/PreProcessingTests.hs
+++ b/app/RequirementTests/PreProcessingTests.hs
@@ -1,5 +1,5 @@
 {-|
-Description: Test some pre-processing functions for requirement parsing and database data.
+Description: Test some pre-processing functions that clean up the text before running the parsers
 
 -}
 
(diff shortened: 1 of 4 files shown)


david-yz-liu (commented):
@ChingChang9 good work. Two comments:

1. I'd like you to store the post description as HTML as well (the `postDescription` record field)
2. I don't think the pruning is necessary, it's okay to link back to the FAS Calendar since that's our "official" source of information.

ChingChang9:
Hi David thanks for the feedback.
For the second point, most (but not all) hrefs on the calendar are relative links (some are absolute links to UTM/grad courses), so we need to pattern match the first char of href and re-format them to make the links work when put directly into the front-end. Is that still more preferable than pruning?

david-yz-liu (approved):

david-yz-liu:
@ChingChang9 ah good point, yeah the pruning is good in this case 👍

PR #1298, opened 2022-08-13
Use focus data (no 304)

ChingChang9:
## Motivation and Context
#1291 but without the 304 response part

## Your Changes
- Moved the server endpoint for `/post` to `/post-progress` and used `/post?code=` for retrieving post data
- Removed `js/components/sidebar/focus_description.js` and used the data from `/post?code=` instead
- Changed the focus labels (pId) to their post code on the timetable. They all start with "ASFOC1689" but I decided to keep the redundancy instead of hardcoding "ASFOC1689" in the `getPost` call, in case we want to expand it to other departments in the future
- Modified the highlightFocus matcher to match substrings because mat235 is now parsed as just mat235 and not hardcoded as mat235237257calc2
- Moved the http request logic from `componentDidMount` to `componentDidUpdate` to only fetch data on open
- Moved the http request logic from `UNSAFE_componentWillUpdate` to `componentDidUpdate` to only make the request once
- Removed the `FocusDescription` component since the logic is a log simpler now

## Testing
- Set up a mock server for `/post?code=` that always returns the data of the AI focus
- Updated the test for course highlighting to be asynchronous
- Used the Network tab in the developer's console to monitor the network calls and ensure we only send a request when we click on the focus, and receive 304 every time after the first request
- We might want to test that mat235237257calc2 is highlighted but it seems like the graphs aren't rendered in the test suites so maybe manual inspection is fine for only one test

- `createJSONResponse` in `app/Util/Happstack.hs` escapes the `\` character, which turns all `\n` into `n` in the response. I wasn't able to figure out why the `\` was escaped in the first place. I removed it and everything still seemed to work. I don't know if this would break something that I overlooked.
- The async setState shows a warning when running `yarn test`. However, [it is safe and the warning is just a false positive that is fixed in a React v18.x](https://github.com/facebook/react/pull/22114)
- [`dangerouslySetInnerHTML` is dangerous for the potential XSS attack](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml). But since we are only calling it on the server data, it should be fine because we can trust our server and art&sci's calendar to not have malicious data

## Checklist


courseography.cabal
```diff
@@ -50,6 +50,7 @@ test-suite RequirementTests
     test-framework-hunit,
     test-framework-quickcheck2,
     text,
+    time,
```
david-yz-liu, line 53: @ChingChang9 everything looks great, I think you just need to remove this dependency (since you've extracted out the timestamps into a separate PR)

david-yz-liu (approved):

PR #1306, opened 2022-10-08
issue-bool-parent-child

nazzaninghazzi:
Modified getGraph function in Graph.js and BoolGroup.js such that the "parent-child" relationship for all types of nodes are computed in getGraph.

## Motivation and Context

"parent-child" relationships for "Node" and "hybrid" nodes used to be computed in  getGraph function in Graph.js and the parent-child relationships for bool nodes were separately computed in BoolGroup. We modified both files so that computing this relationship would become simpler and done at the same place.

## Your Changes

**Description**:
Added a new state called boolNodeToRelationship to Graph component which is an object with bool ids as keys and {parent, children, inEdges, outEdges} as values.
Moved generateBool from BoolGroup.js to getGraph function to generate parents, children, in edges, and out edges for each bool node. This function was ran for all bool nodes in boolsList and the state boolNodeToRelationship was passed in to BoolGroup.js as props. 

## Testing

After modifying the files, we ran `yarn test` to make sure that all the tests passed like before.
## Checklist

### DIFF
--- a/js/components/graph/BoolGroup.js
+++ b/js/components/graph/BoolGroup.js
@@ -36,30 +36,17 @@ export default class BoolGroup extends React.Component {
 
   // Generate data for a Bool node
   generateBool = boolJSON => {
-    var parents = []
-    var childs = []
-    var outEdges = []
-    var inEdges = []
-    this.props.edgesJSON.map(edge => {
-      if (boolJSON.id_ === edge.target) {
-        parents.push(edge.source)
-        inEdges.push(edge.id_)
-      } else if (boolJSON.id_ === edge.source) {
-        childs.push(edge.target)
-        outEdges.push(edge.id_)
-      }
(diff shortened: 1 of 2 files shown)


david-yz-liu (commented):
@nazzaninghazzi @leowrites this is looking good! A couple of overall comments:

1. In general don't define nested functions (`generateBoolRelationships`), put the body of the function in the `forEach` instead.
2. You don't need to create a separate `boolNodeToRelationship` state. Instead, you should try to add onto the existing `connections` state, since the bool node ids should be distinct from the node ids. (You can check this by inspecting the state. This is good time to remind you about using the React Developer Tools browser extension (available for both Firefox and Chrome).

david-yz-liu (approved):
@leowrites @nazzaninghazzi great work!

PR #1308, opened 2022-10-14
js: lift status state from edge and edgegroup to Graph

leowrites:
Moved all state from EdgeGroup that tracks if an edge missing to Graph and passed in the states as a prop to EdgeGroup, and a callback function to update the state. 
Moved individual edge status from Edge to Graph. Passed down the states to Edge as props and a callback for updating that edge.

## Motivation and Context
This pull request lifted states for Edge and EdgeGroup, so in the future we can implement a state management architecture.

## Your Changes
Created a new edgeMissingStatus state in Graph.
Created a new function updateEdgeMissingStatus in Graph to update if an Edge is "missing" and passed it to EdgeGroup.
EdgeGroup now calls update EdgeMissingStatus to update if an edge is missing.
Created a new edgesStatus state in Graph.
Created a new function updateEdgeStatus in Graph to update the state of edgeStatus. Passed updateEdgeStatus to Edge to update its status in Graph.
Modified code in Node and Bool for updateEdgeStatus to work.

**Description**:

## Testing

`yarn test` was ran to make sure all components are still working as expected.

## Checklist


js/components/graph/Graph.js
```diff
@@ -33,6 +34,8 @@ export class Graph extends React.Component {
       hybridsJSON: [],
       boolsJSON: [],
       edgesJSON: [],
+      edgeMissingStatus: {},
```
david-yz-liu, line 37: This status isn't actually used anywhere, and can be deleted, along with all of the `updateEdgeMissingStatus` code.

js/components/graph/EdgeGroup.js
```diff
@@ -19,28 +15,20 @@ export default class EdgeGroup extends React.Component {
...
-    this.props.edgesJSON.forEach(edgeJSON => {
-      this[edgeJSON.id_].updateStatus()
-    })
+    this.props.updateEdgeMissingStatus(edgeID, isMissing)
   }
 
   /**
    * Set all Edges to the inactive status
    */
   reset = () => {
```
david-yz-liu, line 26: Instead of modifying this method, you should move the corresponding logic into the `Graph.reset` method so that it can reset the edge status state itself (rather than this indirect approach of `Graph.reset` calling `EdgeGroup.reset`.

This will also allow you to delete the `ref` that the `Graph` has to the `EdgeGroup` child component.

js/components/graph/EdgeGroup.js
```diff
@@ -7,10 +7,6 @@ import Edge from "./Edge"
 export default class EdgeGroup extends React.Component {
   constructor(props) {
     super(props)
-    // EdgeGroup's state is used to keep track of the edgeIDs of
```
david-yz-liu, line 10: This is good, but now that you've removed all of the state there's no reason to override the constructor at all, so you can just delete this entire method.

js/components/graph/Edge.js
```diff
@@ -72,5 +47,6 @@ Edge.propTypes = {
   /** Node that the edge is pointing to */
   target: PropTypes.string,
   /** function called when the edge's state has changed */
+  edgeStatus: PropTypes.string,
```
david-yz-liu, line 50: Name this prop `status` instead of `edgeStatus`. It's not necessary to specify "edge" here since that's already the name of this component class.

js/components/graph/Node.js
```diff
@@ -200,7 +214,7 @@ export default class Node extends React.Component {
     })
     this.props.inEdges.forEach(function (edge) {
       var currentEdge = svg.edges.current[edge]
-      if (currentEdge.state.status === "missing") {
+      if (currentEdge.props.status === "missing") {
```
david-yz-liu, line 217: As a comment, this line couldn't have been working properly before because you had named the prop `edgeStatus` instead of status. This is a good example of a case where tests should catch an error.

js/components/graph/Graph.js
```diff
@@ -285,6 +293,45 @@ export class Graph extends React.Component {
...
+    })
+  }
+
+  /**
+   * Update the status of the Edge, this will take in a source, a target so it can be looked up
+   * We will loop through the edgesJSON and find use the source, target directly from edgeJSON
+   * the state will be stored and updated in edgeGroup object.
+   */
+  updateEdgeStatus = (source, target, status, edgeID) => {
+    const sourceNode = refLookUp(source, this)
```
david-yz-liu, line 310: Move this logic (up to line 325) into the `Edge.updateStatus` function. The `Graph.updateEdgeStatus` function should only take in the `status` and `edgeID` to update.

david-yz-liu (commented):
@leowrites good start. In addition to addressing the comments I left, please work on adding tests to this PR.

js/components/graph/Edge.js
```diff
@@ -8,46 +8,33 @@ import { refLookUp } from "../common/utils"
 export default class Edge extends React.Component {
   constructor(props) {
     super(props)
-    this.state = { status: "inactive" }
```
david-yz-liu, line 11: Now that you've deleted the only state, you don't need to override the constructor at all.

js/components/graph/Edge.js
```diff
@@ -8,46 +8,33 @@ import { refLookUp } from "../common/utils"
...
+        status = "inactive"
+      } else if (!targetNode.isSelected()) {
+        status = "takeable"
+      } else {
+        status = "active"
+      }
     }
+    this.props.updateEdgeStatus(status, this.props.edgeID)
   }
-
```
david-yz-liu, line 50: Revert this change

js/components/graph/Graph.js
```diff
@@ -791,7 +817,7 @@ export class Graph extends React.Component {
             var currentEdge = svg.edges.current[edge]
             var sourceNode = refLookUp(currentEdge.props.source, svg)
             if (!sourceNode.isSelected()) {
-              currentEdge.setState({ status: "missing" })
+              currentEdge.updateStatus("missing")
```
david-yz-liu, line 820: Because this function is now in `Graph`, we shouldn't need to call `currentEdge.updateStatus` with a specific argument. Instead, you can call the `updateEdgeStatus` function directly here.

js/components/graph/Graph.js
```diff
@@ -1038,7 +1060,13 @@ export class Graph extends React.Component {
             updateNode={this.updateNode}
             focusPrereqs={this.focusPrereqs}
           />
-          <EdgeGroup svg={this} ref={this.edges} edgesJSON={this.state.edgesJSON} />
+          <EdgeGroup
+            svg={this}
+            ref={this.edges}
```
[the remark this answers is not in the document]
david-yz-liu: Based on our conversation today, you can keep this for now but aim to remove it in the future.

js/components/graph/Edge.js
leowrites: Should I move the logic from line 18 to 37 into Graph like we discussed?

david-yz-liu: @leowrites you don't need to do that for this PR. It's better to get this one merged in first, and then do subsequent updates in the Node PR.

david-yz-liu (approved):

PR #1310, opened 2022-10-19
Issue region label

nazzaninghazzi:
## Motivation and Context
Remove the RegionGroup component, and instead make renderRegions and renderLabels methods in the Graph component so that the regions and labels can be directly rendered from there.

Simplifies the coding and removes excessive components.

## Your Changes

**Description**:
Deleted RegionGroup component. Created new methods inside Graph components, renderRegion, renderLabels and renderRegionsLabels. The latter renders groups regions and labels inside a <g> tag. The body of these methods is the same as the previous version. Also, moved the tests for RegionGroup one level up to Graph. 

## Testing

Moved the tests for RegionGroup one level up to Graph. Added a new snapshot for Graph.tests.js to test the rendered value of the new method. This step required initializing a Graph component, passing the test examples as inputs of Graph.renderRegionsLabels method, and testing if the returned value matched the snapshot, by running yarn test.

The new methods for rendering labels and regions uses the same code as before. When mapping over regionsJSON and labelsJSON, tags are returned. The key attribute in these tags has to be set to the index of the elements of the object. However, when calling these methods and comparing them to the snapshots, key attribute is not rendered. I have temporarily modified the snapshots not to include key attribute in order for the tests to pass. But, not having a key attribute could be an issue. I would appreciate some hints.

## Checklist


david-yz-liu (approved):

david-yz-liu:
@nazzaninghazzi great job. Regarding your question about the snapshots, I suspect the issue is that the old tests rendered React components with `.shallow()`, but now you're testing a specific method call's return value. (`key` is a special React attribute, just used by React but not in the HTML/DOM.) I don't think it's a big deal to omit it from the snapshots, but will keep an eye on it for later.

PR #1311, opened 2022-10-21
js: refactor var to let/const using eslint

modderme123:
## Motivation and Context

`var` has [subtle quirks](https://javascript.info/var) which lead to accidental bugs. A modern alternative is to try using let and const statements which have the added benefit of making immutable variables explicit. 

## Your Changes

**Description**:

Replace `var` with `let`, and where possible `let` with `const`, automatically generated by applying the eslint rules [no-var](https://eslint.org/docs/latest/rules/no-var) and [prefer-const](https://eslint.org/docs/latest/rules/prefer-const) to *.js and *.jsx (excluding the ignored files)

## Testing

- `yarn run test` (completes with no errors)
- Manually testing that the graph still displays and info shows a popup

This left a few var statements inside ifs that will need to be manually refactored but I thought it would be easier to review the straightforward changes in one PR (the diff here was generated by using eslint with the `--fix` flag, the others will have to be fixed manually). 

Also, there are still many uses of vars in js/components/draw/* but I searched for imports and couldn't find any uses yet so I'm unsure if a PR to change those is wanted.

## Checklist

### DIFF
--- a/js/components/common/export.js.jsx
+++ b/js/components/common/export.js.jsx
@@ -30,7 +30,7 @@ export class ExportModal extends React.Component {
   }
 
   getGraphImage() {
-    var necessaryLS = new Object()
+    const necessaryLS = new Object()
     for (const elem in localStorage) {
       if (!localStorage.hasOwnProperty(elem)) {
         continue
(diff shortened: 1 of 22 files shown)


david-yz-liu (approved):
Thanks @modderme123!

PR #1312, opened 2022-10-24
Issue bool group

nazzaninghazzi:
## Motivation and Context
For further simplicity, moved the status state of each Bool and thus the methods setting this state (updateNode and focusPrereqs) up to the Graph component.

## Your Changes

**Description**:
Moved the status state of Bool Component up to the Graph component. This required adding a status key with the initial value of inactive while setting the value of boolsJSON. So boolsJSON is a list of boolJSON where each element is a list of key-value pairs with the new key-value pair("status": "inactive") added. The methods updateNode and focusPrereqs which used setState to change the state of status where moved up to the Graph component since status didn't exist in Bool anymore. Thus, the status of each boolNode, updateNode and focusPreps are passed as props to boolGroup and then Bool. Every component calling these functions, including boolGroup, Node, and Edge) had to be changed to call the updated version.

## Testing

Run `yarn test` for all the test suites. Run `stack run` and open the link in the Chrome browser. Using the React developer tools, track the changes of status in Graph component by clicking and un clicking on nodes.
## Checklist


js/components/graph/Bool.js
```diff
@@ -6,19 +6,17 @@ import PropTypes from "prop-types"
 export default class Bool extends React.Component {
   /**
    * Create a boolean node.
-   * @param {string} state - The status of the node which can be active, inactive, or missing
    */
   constructor(props) {
```
david-yz-liu, line 10: Now that you've removed the only state, you can delete this method entirely (no need to override the constructor).

js/components/graph/Graph.js
```diff
@@ -153,6 +154,7 @@ export class Graph extends React.Component {
           } else if (entry.type_ === "Hybrid") {
             hybridsList.push(entry)
           } else if (entry.type_ === "BoolNode") {
+            entry["status"] = "inactive"
```
david-yz-liu, line 157: Overall, I don't want you to modify the existing JSON data that's fetched from the server. Instead, you should create a new object that maps bool node id to its status (which can indeed start as inactive).

js/components/graph/Edge.js
```diff
@@ -20,7 +20,18 @@ export default class Edge extends React.Component {
...
       return
     }
-    if (!source.isSelected() && target.state.status === "missing") {
+
+    if (
+      !source.isSelected() &&
+      target.props.className === "bool" &&
+      target.props.JSON.status === "missing"
+    ) {
+      this.setState({ status: "missing" })
```
david-yz-liu, line 29: It's not clear why you changed this if statement, since the code (`this.setState(...)`) in this branch is identical to the code in the following else if branch.

js/components/graph/BoolGroup.js
```diff
@@ -18,8 +18,7 @@ export default class BoolGroup extends React.Component {
    */
   reset = () => {
     this.props.boolsJSON.forEach(boolJSON => {
-      var bool = this[boolJSON.id_]
-      bool.setState({ status: "inactive" })
+      this[boolJSON.status] = "inactive"
```
david-yz-liu, line 21: Now that the bool status is stored in the `Graph` component, it is no longer necessary to define a `reset` method in this component. Instead, you can modify the `Graph.reset` method to change the status of each bool node.

js/components/graph/Graph.js
```diff
@@ -728,6 +730,67 @@ export class Graph extends React.Component {
...
+    var svg = boolNode.props.svg
+    var newState = boolNode.arePrereqsSatisfied() ? "active" : "inactive"
+    var boolId = boolNode.props.JSON.id_
+
+    this.setState(
+      prevState => {
+        let new_boolsJSON = prevState.boolsJSON
+        let node = new_boolsJSON.find(element => element.id_ === boolId)
+        new_boolsJSON[new_boolsJSON.indexOf(node)].status = newState
+        return { new_boolsJSON }
```
david-yz-liu, line 746: This is not the correct way to update existing state.

js/components/graph/Node.js
```diff
@@ -120,8 +120,13 @@ export default class Node extends React.Component {
         localStorage.setItem(nodeId, newState)
         this.props.childs.forEach(function (node) {
           var currentNode = refLookUp(node, svg)
-          if (currentNode !== undefined) {
```
david-yz-liu, line 123: Don't change this code. Instead, define an `updateNode` function in `Bool` that simply calls `this.props.updateNode()`. Similarly for `focusPrereqs` below.

js/components/graph/Graph.js
```diff
@@ -728,6 +730,67 @@ export class Graph extends React.Component {
...
+    var svg = boolNode.props.svg
+    var boolId = boolNode.props.JSON.id_
+    // Check if there are any missing prerequisites.
+    if (boolNode.props.JSON.status !== "active") {
+      this.setState(
+        prevState => {
+          let new_boolsJSON = prevState.boolsJSON
+          let node = new_boolsJSON.find(element => element.id_ === boolId)
+          new_boolsJSON[new_boolsJSON.indexOf(node)].status = "missing"
+          return { new_boolsJSON }
```
david-yz-liu, line 776: See above comment.

david-yz-liu (commented):
@nazzaninghazzi good start. In addition to the comments I made, you should be adding several additional tests to check your changes carefully. There were errors in this code that should have been caught by test cases.

david-yz-liu (approved):

PR #1314, opened 2022-11-07
lift status state from Node to Graph

nazzaninghazzi:
## Motivation and Context

For further simplicity, moved the `status` and `selected` state of each Node up to the Graph component. Thus, two new methods updateNodeStatus and updateNodeSelected were added to Graph component two update these states.
## Your Changes

**Description**:
Moved the `status` and `selected` state of each Node component to Graph Component. This required adding a new state called nodesState in Graph which is an object with node ids as keys. The value of each node id is an object of status key: status value and selected key: selected value. This state is passed as props to Node Group and Node components. 
Since states don't exist in Node anymore, two new methods updateNodeStatus and updateNodeSelected were created in Graph to change nodesState and were called in Node's methods that changed `status` and `selected`. These two methods were also passed as props to Node Group and Node. 
The Edge component had to be changed to call the updated version of node's state.

## Testing

A few tests were added to Node to test resetting of nodes since the reset method was modified.
`yarn test` was ran to make sure all components are still working as expected.
## Checklist


js/components/graph/Graph.js
```diff
@@ -116,7 +117,6 @@ export class Graph extends React.Component {
     let url = new URL("/get-json-data", document.location)
     const params = { graphName: graphName }
     Object.keys(params).forEach(key => url.searchParams.append(key, params[key]))
-
```
david-yz-liu, line 119: Revert this change

js/components/graph/Graph.js
```diff
@@ -64,8 +65,8 @@ export class Graph extends React.Component {
     }
 
     this.nodes = React.createRef()
-    this.bools = React.createRef()
     this.edges = React.createRef()
+    this.bools = React.createRef()
```
david-yz-liu, line 69: Revert this change

js/components/graph/Graph.js
```diff
@@ -31,6 +31,7 @@ export class Graph extends React.Component {
       labelsJSON: [],
       regionsJSON: [],
       nodesJSON: [],
+      nodesState: {},
```
david-yz-liu, line 34: This name should be consistent with the edge and bool states below.

david-yz-liu (commented):
@nazzaninghazzi @leowrites Good job, I left some comments for you to work on.

In addition to these, in general you should not rely on passing a method from `Graph` down to `Node` as a prop without a good reason. Now that all of the state is in `Graph`, it should be able to call all of its own methods without calling any `Node` methods.

For any child-child communication (e.g. an `Edge` method calling `Node.updateNode`), you should be able to now modify that code so that the child method calls a `Graph` method instead, since the `Graph` object is passed as the `svg` prop down the tree.

js/components/graph/Graph.js
```diff
@@ -783,27 +828,125 @@ export class Graph extends React.Component {
...
+            const currentEdge = this.edges.current[edge]
+            currentEdge.updateStatus()
+          })
+        }
+      )
+      localStorage.setItem(nodeId, newState)
+    }
+  }
+
+  toggleSelection = targetNode => {
```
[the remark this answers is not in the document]
leowrites: if we just pass in the node ID, can we use refLookUp to find that node? We still need to call updateNode in the callback.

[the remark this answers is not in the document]
david-yz-liu: @leowrites yes, I think `refLookUp` is acceptable here.

david-yz-liu (approved):

PR #1316, opened 2022-11-26
Restore edge and bool status on page reload

leowrites:
Fixed a bug where the status of the edges are not restored when the page is reloaded.

## Motivation and Context

This PR fixes the bug mentioned along with tests to verify the fix is working.

## Your Changes

In `Graph.js` , instead of initializing all edge statuses to "inactive", we now compute the correct status based on `nodesStatus` and `boolsStatusObj`. The status computation is similar to `updateEdgesStatus`.

During the process, I found that bool statues are also initialized to "inactive" which is incorrect. Instead, like nodes, I retrieved the bool statuses from localStorage when initializing bools.

**Description**:

## Testing

I have added two tests to the states are restored when `window.location.reload()` is called, which is supposed to simulate page refresh.

## Checklist


js/components/graph/Graph.js
```diff
@@ -160,7 +160,9 @@ export class Graph extends React.Component {
           } else if (entry.type_ === "Hybrid") {
             hybridsList.push(entry)
           } else if (entry.type_ === "BoolNode") {
-            boolsStatusObj[entry.id_] = "inactive"
+            boolsStatusObj[entry.id_] = localStorage.getItem(entry.id_)
```
david-yz-liu, line 163: It's a common pattern to use `x || y` instead of `x ? x : y`, taking advantage of the `||` short-circuiting behaviour (and the fact that it actually returns one of `x`/`y`, not a boolean).

david-yz-liu (commented):
@leowrites great job! I just left one small comment

david-yz-liu (approved):

PR #1317, opened 2022-11-27
Turning the JSON state in graph from arrays into objects

nazzaninghazzi:
Turning the `JSON` state in `graph` from arrays into objects (mapping ID to graph object).
## Motivation and Context

Make lookup codes simpler and faster in the future. Accessing the properties of a node, hybrid, bool, or edge is done through reference lookup. That is `refLookUp` method is called on an id and returns the component so we could access the props. This process is necessary because the nodes, hybrids, bool, and edges of the `Graph` were stored as arrays and iterating over an array to find a node or edge was inefficient. In the future, JSON states can be used to access properties of a component.

## Your Changes

**Description**:
When the data is retrieved from the server in `getGraph`, store the data in `nodesJSON`, `hybridsJSON`, `boolsJSON`, `edgesJSON`, `regionsJSON`, and `labelsJSON` as an object (mapping ID to graph object), rather than an array.
Whenever looping over a state was necessary, the code was changed to loop over values of state object instead of the state object itself.

## Testing

Run `yarn test` and ensure all the tests pass after refactoring the types of the states in `Graph`.
## Checklist


david-yz-liu (commented):
@nazzaninghazzi this is good work, but this PR is doing too much at the same time. I only asked you to modify the state values for the "JSON" collections, not to change the other code to make use of it (that can come later in a separate PR).

Please revert the other changes in this PR so that it *only* makes changes for how this state is stored and updated. This will make this PR smaller and easier to review (and easier to find bugs).

js/components/graph/EdgeGroup.js
```diff
@@ -38,8 +38,8 @@ export default class EdgeGroup extends React.Component {
   render() {
     // Missing edges must be rendered last. The sort
     // method custom sorts a copy of edgesJSON so that all missing edges
-    // are last in the list. Then render based on that list.
-    const edges = this.props.edgesJSON
+    // are last in the list. Then render based on tha
```
david-yz-liu, line 41: This is a typo, I might have introduced it when fixing the merge conflicts (keep the "that list.").

js/components/graph/BoolGroup.js
```diff
@@ -8,7 +8,7 @@ export default class BoolGroup extends React.Component {
    * For each entry in boolsJSON update the id with the svg
    */
   componentDidMount() {
-    this.props.boolsJSON.forEach(boolJSON => {
+    Object.values(this.props.boolsJSON).forEach(boolJSON => {
```
david-yz-liu, line 11: Use `Object.keys` here, as that will let us access the `id`s directly

js/components/graph/Graph.js
```diff
@@ -187,13 +190,14 @@ export class Graph extends React.Component {
...
-            noDuplicatesNodesList.push(node)
+            noDuplicatesNodesJSON[node.id_] = node
           }
         })
 
-        hybridsList.forEach(hybrid => {
+        Object.values(hybridsJSON).forEach(hybrid => {
           childrenObj[hybrid.id_] = []
           outEdgesObj[hybrid.id_] = []
+          var nodesList = Object.values(nodesJSON)
```
david-yz-liu, line 200: Use `const` instead of `var` here

js/components/graph/Graph.js
```diff
@@ -217,12 +221,12 @@ export class Graph extends React.Component {
           }
         })
 
-        boolsList.forEach(boolJSON => {
+        Object.values(boolsJSON).forEach(boolJSON => {
```
david-yz-liu, line 224: Similar to my above comment, use `Object.keys` here to access the ids directly.

js/components/graph/Graph.js
```diff
@@ -237,7 +241,7 @@ export class Graph extends React.Component {
           inEdgesObj[boolJSON.id_] = inEdges
         })
 
-        noDuplicatesNodesList.forEach(node => {
+        Object.values(noDuplicatesNodesJSON).forEach(node => {
```
david-yz-liu, line 244: Here as well (use `Object.keys`, not `Object.values`).

js/components/graph/Graph.js
```diff
@@ -305,9 +309,12 @@ export class Graph extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    if (prevState.nodesJSON !== this.state.nodesJSON) {
+    if (
```
david-yz-liu, line 312: This change isn't good; it's bad practice to do a deep comparison of object values in `componentDidUpdate`, as that takes time linear in the size of the object.

If you're finding this if condition isn't triggering correctly, that means that some code isn't creating a new `nodesJSON` object, but instead mutating the existing object. You can find and modify the problematic code to create a copy of the `nodesJSON` state.

js/components/graph/Graph.js
```diff
@@ -494,15 +501,15 @@ export class Graph extends React.Component {
...
+        for (const node of Object.values(this.state.nodesJSON)) {
           if (node.id_ === this.state.draggingNode) {
             currentNode = node
           }
         }
         currentNode.pos = [newPos.x - 20, newPos.y - 15]
         currentNode.text[0].pos = [newPos.x, newPos.y + 5]
-        const newNodesJSON = [...this.state.nodesJSON]
-        newNodesJSON.push(currentNode)
+        const newNodesJSON = [...Object.values(this.state.nodesJSON)]
```
david-yz-liu, line 511: You should be creating an object, not an array here. You can try using just `{...this.state.nodesJSON}` instead. This occurs a few more times below.

david-yz-liu (approved):

PR #1319, opened 2022-11-29
build: enable webpack production mode by default

modderme123:
## Motivation and Context

Building in production enables some optimizations in React and ships a significantly smaller js bundle to users. However, this comes at the tradeoff of worse debugability and slower compile times (40s vs 1s), so should only be used for production. Thus we split the server into two modes: development (`stack run dev-server`) and production (`stack run server`)

## Your Changes

**Description**:
In this PR we allow for running a custom client build command when we run the server. This allows us to swap in production and development commands.

We then add a `dev-server` command for running webpack in dev and watch mode locally, and make the default `server` run `yarn run build` instead of `yarn run watch` (organized this way so that the production server doesn't have to change the run command)

- [x] Documentation update (change that modifies or updates documentation)

## Testing

`stack test` and `yarn test` and quick visual testing in the browser.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ $ stack build   # Compile Courseography and all Haskell dependencies (this will 
 1. Run `$ stack run` to start the server
 2. Navigate to `http://localhost:8000/graph` in your browser
 
+When running in development you should run `$ stack run dev-server` instead of `$ stack run` to start the server. This will automatically recompile the client when you make changes to the source code.
+
 #### Developers
 
 If you are contributing to Courseography, you should install our pre-commit hooks:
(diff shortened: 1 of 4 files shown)


david-yz-liu (commented):
@modderme123 thank you! My main comment is that I want the default to still be the development server, so please swap to `stack run server` and `stack run prod-server`

modderme123:
Updated! Looks like CircleCI is failing because the `sudo` command is not found...

david-yz-liu (approved):

david-yz-liu:
Thank you @modderme123, indeed it looks like the docker image we were using for CircleCI was updated 2 days ago (https://hub.docker.com/r/fpco/stack-build/)!

modderme123:
Neat, thanks!

PR #1322, opened 2022-12-07
Remove unnecessary refs by using Graph states

nazzaninghazzi:
## Motivation and Context

Now that `nodesJSON`, `hybridsJSON`, `boolsJSON`, and `edgesJSON` states in `Graph` are stored as objects instead of lists, we can take advantage of them to access the props of components rather than creating references.

## Your Changes

**Description**:
Accessing the properties of a node, hybrid, bool, or edge was previously done through reference lookup. 
`this.nodes.current`, `this.bools.current`, and `this.edges.current` are mapping of component ids to component objects. In order to get a component and access a property, we had to look into these mappings and fetch corresponding components. This is unnecessary since most of these properties are stored in Graph states and now that Graph states are objects, we can simply access information through them. 
Removed any redundant reference lookup in `Graph` methods and instead got the property value from states. 
Some redundant props that were passed from `Graph` to child components including `NodeGroup`, `Node`, `BoolGroup`, `Bool`, `EdgeGroup`, `Edge` were deleted as well.

## Testing

Run `yarn test` and ensure all the tests pass after removing some references in `Graph`.
## Checklist


js/components/graph/Node.js
```diff
@@ -124,19 +112,15 @@ export default class Node extends React.Component {
...
-  inEdges: PropTypes.array,
   JSON: PropTypes.object,
   onClick: PropTypes.func,
   onMouseEnter: PropTypes.func,
   onMouseLeave: PropTypes.func,
-  outEdges: PropTypes.array,
   status: PropTypes.string,
-  selected: PropTypes.bool,
   parents: PropTypes.array,
   svg: PropTypes.object,
```
david-yz-liu, line 125: You can remove this prop as well, and instead pass in the specific callback functions that are needed (`onWheel`, `onKeyDown`)

js/components/graph/Graph.js
```diff
@@ -11,7 +11,6 @@ import NodeGroup from "./NodeGroup"
 import GraphDropdown from "./GraphDropdown"
 import Sidebar from "./Sidebar"
 import { parseAnd } from "../../util/util.js"
-import { refLookUp } from "../common/utils"
```
david-yz-liu, line 14: Good! It looks like this function is no longer being used, so you can delete it altogether.

david-yz-liu (commented):
@nazzaninghazzi @leowrites great job! I just left a few minor comments. Please also do a merge with master, as I made a couple of small bug fixes to the Graph component yesterday. (#1321)

david-yz-liu:
@nazzaninghazzi @leowrites you should double-check the prop passing for the callbacks. Make sure to test this carefully.

david-yz-liu (approved):

PR #1328, opened 2023-03-30
Add course links

nazzaninghazzi:
## Motivation and Context
Clicking on the info box of a course, displays a modal including the information on that course. This information contains some course names in the description and prerequisite sections. The goal was to convert those names to links so by clicking on them, the modal would be switched to show the information for the clicked course.

## Your Changes

**Description**:
Identified the pattern of the course names and searched for them using a regular expression. The words matching the pattern were added as <a></a> tags to the `result` list and other non-special words as strings. Some attributes including the cursor style and onClick action were passed to the link tags.
The method `clickRelated` made some changes to the component state and these changes were considered for updating the component and switching to the information of the clicked course.

## Testing

Run `stack run`, click on courses to ensure the course names are correctly converted to links and clicking links would jump to the modal of the clicked course.

## Checklist


js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
       course: [],
       sessions: [],
       courseTitle: "",
+      clicked: 0,
     }
   }
+  /**
+   * Change the clicked and courseId state, whenever a course link is clicked.
+   */
+  clickRelated = s => {
```
david-yz-liu, line 43: Pick a better parameter name. In general do not use single-letter variable names.

Also, the function name should describe what the function does, not where it's going to be used. "click" is something the user does, but this function name should describe *what happens when a course name is clicked*.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+    let copyState = this.state
```
david-yz-liu, line 44: In general, you shouldn't update state like this. Instead, pass into `setState` a new object that has just the state values you want to update, e.g. `this.setState({courseId: s, clicked: 1})`.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+  }
+
+  /**
+   * Convert the course names to <a> tags and return a list of strings and <a> tags.
+   */
+  getRelatedCourses = content => {
+    const result = []
+    if (content !== null) {
+      content.split(" ").forEach((word, i) => {
+        if (word.match(/[A-Z]{3,4}[0-9]{3}[H|Y]1/gi)) {
```
david-yz-liu, line 57: Pull out this regular expression into a constant.

david-yz-liu: Also, about the regex itself:

1. I'm not sure there are any course codes in FAS that have 4 letters, did you find such examples?
2. Make sure this regex matches the entire string, not just a substring. Otherwise it'll result in false positives.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+          let symbol = ""
+          // check if the last character is not 1 (is a symbol). If so, remove it from word and
```
david-yz-liu, line 59: Don't split up the string. As I've said before, your code should work correctly passing in the complete course code, including the H1/Y1.

js/components/common/react_modal.js.jsx
david-yz-liu: Additionally: it's *very* important to note that assigning `let copyState = this.state` does **not** create a copy of the object, but rather creates an alias. The name copy, here and further below, is very misleading and should be avoided.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+        result.push(" ")
+      })
+    }
+    return result
+  }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.courseId !== this.props.courseId && this.props.courseId !== "") {
-      getCourse(this.props.courseId).then(course => {
+    let changed = 0
```
david-yz-liu, line 92: Don't use this variable. Instead you can check whether the props changed or not when initializing `newCourse` below directly.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+    if (this.props.showCourseModal !== prevProps.showCourseModal) {
+      changed = 1
+    } else if (this.state.clicked === 1) {
+      changed = 2
+    }
+    if (changed !== 0) {
+      let newCourse = ""
+      changed === 1
```
david-yz-liu, line 100: Do not use multi-line ternary ifs

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+    let copyState = this.state
+    copyState.courseId = s
+    copyState.clicked = 1
+    this.setState(copyState)
+  }
+
+  /**
+   * Convert the course names to <a> tags and return a list of strings and <a> tags.
+   */
+  getRelatedCourses = content => {
```
david-yz-liu, line 53: This is not a good name for what this function does.

js/components/common/react_modal.js.jsx
```diff
@@ -34,19 +34,86 @@ class CourseModal extends React.Component {
...
+      })
+    }
+    return result
+  }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.courseId !== this.props.courseId && this.props.courseId !== "") {
-      getCourse(this.props.courseId).then(course => {
+    let changed = 0
+    if (this.props.showCourseModal !== prevProps.showCourseModal) {
```
david-yz-liu, line 93: Overall the management of the `courseId` has gotten more complex, since we need to take into account both the props (when the user opens a modal from the graph) and changes to the state (when a user clicks on the link in the existing modal).

To keep the logic of `componentDidUpdate` relatively simple, and also preserve the caching behaviour for when a user opens the modal from the same course twice, please make the following changes:

1. If the prop `courseId` changes (which is what the original code checked), don't call `getCourse`, instead `setState` to modify the `state.courseId`. (This will trigger another call to `componentDidUpdate`.)
2. If the *state* `courseId` changes, call `getCourse`. This should then handle both types of "course id state changes", and remove the need for the `clicked` state that you added.

david-yz-liu (commented):
@nazzaninghazzi great work. I left some comments to help simplify your code, but the other thing I'd like you to do is *please write test cases* to test this new functionality.

js/components/common/react_modal.js.jsx
nazzaninghazzi: @david-yz-liu The splitting by space is done for the entire string, not the course code. So the course code would be for example 'CSC207H1' . Should I still keep the split?

david-yz-liu: Here I meant the splitting of the "H1"/"Y1" is unnecessary (but splitting on spaces is good!).

js/components/common/react_modal.js.jsx
david-yz-liu: Right, what I mean by "caching" here is that if the user opens the modal for the same course twice, the data is fetched only once (`getCourse`), precisely because `prevProps.courseId === this.props.courseId`.

david-yz-liu (approved):

PR #1334, opened 2023-05-11
convert Edge component to function-based component

jaeyonglee3:
## Motivation and Context
Function-based components are recommended over class components in React as they are simpler and easier to read. They can be written using less code and are better for maintaining best practices. 

## Your Changes
The Edge graph component, which was previously defined as a class component, is now defined as a function-based component. 

**Description**:

## Testing
The js/components/graph/__tests__/Edge.test.js file was used to test that the function-based edge component did not change in behaviour from when it was defined as a class component. 

## Checklist


david-yz-liu (commented):
@jaeyonglee3 your actual code change looks good, but it seems like there are other changes in this branch in `.yarn`, `Icon`, `package-lock.json`, and `yarn.lock`. I'm not sure what caused those changes (possibly using `npm` instead of `yarn` to install dependencies), but you should revert them. You can look up how to revert changes to specific files with git.

In the future, please make sure to review the files you're committing carefully to make sure you only commit the files related to the task at hand, and avoid "shortcuts" to add/commit all files at once.

js/components/graph/Edge.js
```diff
@@ -5,14 +5,14 @@ import React from "react"
  * Function-based component representing an edge from a Node/Bool to a Node/Bool
  */
 export default function Edge(props) {
-  const pathAttrs = { d: "M" }
+  let pathDescription = "M"
   props.points.forEach(p => {
-    pathAttrs.d += p[0] + "," + p[1] + " "
+    pathDescription += p[0] + "," + p[1] + " " + " "
```
david-yz-liu, line 10: There seems to be an additional `" "` which is unnecessary, and causing some of the snapshot tests to fail. (Remember you should run the Javascript tests locally with `yarn run test`)

david-yz-liu (approved):

PR #1335, opened 2023-05-14
Convert Bool from a class component to a functional component

MaryamTaj:
## Motivation and Context

By converting Bool from a class component to a functional component, Courseography can have more concise, cleaner, and less complex code. 

## Your Changes

**Description**: Bool was a class-based component for the React graph. I converted Bool into a function-based component. 
- I redefined Bool as a function.
- I passed props as an argument to Bool (I did not use destructuring syntax, and pass individuals props because I felt that
passing the whole object props itself would better support changes to the code. i.e., If you add another property later on). 
- I removed Bool's render method.
- I removed references to 'this.'
- I removed event handler bindings.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Refactoring (internal change to codebase, without changing functionality)
- [ ] Test update (change that modifies or updates tests only)
- [ ] Documentation update (change that modifies or updates documentation only)
- [ ] User Interface Change
- [ ] Other (please specify):

## Testing

The Bool.test.js file contained tests to ensure that the class-based component, Bool, ran correctly. After converting Bool into a functional component, I ran all the tests in Bool.test.js again, and they passed. Thus, I could see that even as a functional component, Bool did not change in its behavior. 

I ran Bool.test.js to ensure that despite changing Bool to a functional component, the boolean nodes rendered and functioned the same. **However, after closer inspection, I realized that Bool.test.js had tests like 'expect(andBool.classList[0]).toBe("bool")', and they passed. I do not understand why they passed, given that Bool is no longer a class.** I have looked through the code, and I wonder if creating a test graph allows Bool.test.js to pass all of its tests? In that case, we are not actually testing on the Courseography graph. 

## Checklist


david-yz-liu (approved):
@MaryamTaj great work. In response to your question, `.classList` is checking for the HTML tag's `class` attribute, which is set in the React code by `className={...}`. So it's not checking whether an element was created by a React class-based component, which is why the tests are still passing after your change.

PR #1338, opened 2023-05-19
Convert Group components to methods within Graph component

jaeyonglee3:
## Motivation and Context
It is not necessary to define BoolGroup, EdgeGroup, and NodeGroup as standalone components. These components can be moved to be within the Graph component where they each have their own method that will render the components.  

## Your Changes

**Description**: Moved the EdgeGroup, BoolGroup, and NodeGroup components to be defined within the Graph component and rendered using their own Graph instance methods. The test files for each of these components have been modified to use the Graph instance methods to render them. 

## Testing

The provided JS testing suite has been used to test that the refactoring changes have not affected the behaviour of the application. 

## Checklist


js/components/graph/Graph.js
```diff
@@ -1219,6 +1219,127 @@ export class Graph extends React.Component {
...
+   * @param {JSON} boolsJSON
+   * @param {object} boolsStatus
+   * @param {object} connections
+   * @return {JSX.Element}
+   */
+  renderBoolGroup = (boolsJSON, boolsStatus, connections) => {
+    return (
+      <g id="bools">
+        {Object.values(boolsJSON).map(boolJSON =>
+          generateBool(boolJSON, boolsStatus, connections)
```
david-yz-liu, line 1233: `generateBool` doesn't need to be a separate function, write the logic directly in here.

david-yz-liu: And similarly for `generateEdge` below

david-yz-liu (commented):
@jaeyonglee3 great work. To address your comments/questions:

- I made an inline comment about `generateBool`/`generateEdge`. 
- Yes, `editMode` and `onDraw` are two old props that are no longer used, so it's great to remove them.

david-yz-liu (commented):
@jaeyonglee3 Looks good. The test files are generally in one-to-one correspondence with component files, and so after deleting the three component files, you should delete the corresponding test files and move the tests to the `Graph` test file.

david-yz-liu (approved):

PR #1340, opened 2023-05-24
Convert Node from a class based component to a function based component

MaryamTaj:
## Motivation and Context

By converting Node from a class component to a functional component, Courseography can have more concise, cleaner, and less complex code.

## Your Changes

**Description**:  Node was a class-based component for the React graph. I converted Node into a function-based component.

- I redefined Node as a function.
- I passed props as an argument to Node (I did not use destructuring syntax and pass individual props because I felt that
passing the whole object props itself would better support changes to the code. i.e., If you add another property later on).
- I removed Node's render method.
- I removed references to 'this.'

- [X] Refactoring (internal change to codebase, without changing functionality)

## Testing

The Node.test.js file contained tests to ensure that the class-based component, Node, ran correctly. After converting Node into a functional component, I ran all the tests in Node.test.js again, and they passed. Thus, I could see that even as a functional component, Node did not change in its behavior.

- Line 79 has the following statement **"TODO: Look at this.props to see what we need to give the g."** 
- Currently, we pass the following properties to g; gAttrs specifies textRendering, shapeRendering, onKeyDown, onWheel, onMouseEnter, onMouseLeave, and onClick. Additionally, we specify id, className, and data-testid. 
-  However, Node also includes properties like editMode, **highlighted**, hybrid, JSON, status, parents, and nodeDropshadowFilter. **Given that passing in a property to g implies that <ellipse>, <rect>, and <text> will also inherit those properties, should we pass this.highlighted to g?** (i.e., This property would ensure that if <ellipse> is highlighted, <rect> and <text> will also be highlighted. In other words, all the elements are highlighted or none at all). 

## Checklist

### DIFF
--- a/js/components/graph/Node.js
+++ b/js/components/graph/Node.js
@@ -76,7 +76,6 @@ export default function Node(props) {
 
   const textXOffset = props.JSON.pos[0] + props.JSON.width / 2
 
-  // TODO: Look at this.props to see what we need to give the g
   return (
     <g
       {...gAttrs}
--- a/js/components/graph/Node.js
+++ b/js/components/graph/Node.js
@@ -23,92 +23,88 @@ import React from "react"
  *  - Hybrid nodes are the smaller, grey nodes on the graph that represent another course node
  *    farther away. They can only be either 'active' or 'inactive'
  */
-export default class Node extends React.Component {
-  getDataTestId = () => {
-    if (this.props.hybrid) {
-      return `h(${this.props.parents.join(",")})`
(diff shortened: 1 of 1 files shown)


david-yz-liu (commented):
@MaryamTaj great work. You can **delete the TODO on line 79**, that's from a while ago and I think the props have since been reviewed and simplified. The `highlighted` prop specifically doesn't need to passed to the `<g>` element; that prop is used to create the `ellipse` that "highlights" the node when it is part of a selected focus.

PR #1344, opened 2023-06-01
finished initial implementation

harshkpatel:
I am adjusting the seeding for graphs to make it less clumsy by either just performing the insert if the graph doesn't exist and then replacing if it existed before.

## Motivation and Context

The code is clumsy, this refactor allows the code to be cleaner and easier to scale and modify. 

## Your Changes

**Description**: I removed deleting all the graphs, added just deleting the graph that was being parsed, and doing just inserting if the graph already existed.

## Testing

I ran the test suite and ran into no issues. I added the print statements to indicate if the right intended action was executing, which it is, and I tested it both on an existing graph and a graph that doesn't exist. (Those print statements are now gone, but that can be added in as I used pattern matching to split up the cases of the graph existing or not existing.)

## Checklist


app/Svg/Database.hs
```diff
@@ -31,11 +31,15 @@ insertElements (paths, shapes, texts) = do
...
-deleteGraphs = do
-    runMigration migrateAll
-    deleteWhere ([] :: [Filter Text])
-    deleteWhere ([] :: [Filter Shape])
-    deleteWhere ([] :: [Filter Path])
-    deleteWhere ([] :: [Filter Graph])
+-- | Delete a graph with the given name from the database
+deleteGraph :: T.Text -> SqlPersistM ()
+deleteGraph graphName = do
+  runMigration migrateAll
```
david-yz-liu, line 37: Oh we must have missed this in your last PR, but running the migration here is redundant and should be removed.

app/Svg/Parser.hs
```diff
@@ -71,10 +71,19 @@ performParse :: T.Text -- ^ The title of the graph.
              -> String -- ^ The filename of the file that will be parsed.
              -> SqlPersistM ()
 performParse graphName inputFilename = do
+    deleteExistingGraph graphName
```
david-yz-liu, line 74: This is good, but make sure to add a note to the documentation of this function to describe the new behaviour.

app/Svg/Parser.hs
```diff
@@ -71,10 +71,19 @@ performParse :: T.Text -- ^ The title of the graph.
...
     liftIO . print $ "Parsing graph " ++ T.unpack graphName ++ " from file " ++ inputFilename
     graphFile <- liftIO $ T.readFile (graphPath ++ inputFilename)
     performParseFromMemory graphName graphFile False
 
+-- | Check if a graph exists and delete it if it does
+deleteExistingGraph :: T.Text -> SqlPersistM ()
+deleteExistingGraph graphName = do
+  maybeResponse <- liftIO $ getGraph graphName
```
david-yz-liu, line 82: It's good that you found the `getGraph` function, but you shouldn't actually call it here for two reasons. First, it executes a much larger database query than is necessary for this function; and, its type is `IO` rather than `SqlPersistM`, which basically means that `getGraph` executes in its own database transaction, and we should avoid executing multiple database transactions in the same "action" when possible.

Instead, you should take just the part that's relevant for this function (getting the Graph!) and use it here.

app/Svg/Parser.hs
```diff
@@ -71,10 +71,19 @@ performParse :: T.Text -- ^ The title of the graph.
...
+  case maybeResponse of
+    Just _ -> deleteGraph graphName
```
david-yz-liu, line 84: After doing what I've said above, you don't need to pass in the `graphName` here; instead, you should be able to pass in the graph Id itself, so that the type of `deleteGraph` becomes something like `Key Graph -> SqlPersistM ()`.

david-yz-liu (approved):

PR #1349, opened 2023-06-19
Separate Timetable and Artsci Calendar Parsing Actions

harshkpatel:
## Motivation and Context

Courseography gets information from Artsci Calendar and Artsci Timetable, and both update at different times, so this refactor was done for usefulness.

## Your Changes

**Description**: Modified parseAll into two separate functions: parseCalendar and parseTimetable

## Testing

Ran the test suite with no issues, and complied the code, it is working properly.

## Checklist


app/Database/Database.hs
```diff
@@ -34,19 +34,28 @@ setupDatabase = do
...
+runPreliminaries = do
     setupDistributionTable
     print distTableSetUpStr
     setupBreadthTable
     print breathTableSetUpStr
-    parseAll
     seedVideos
 
+-- | Sets up the course information from Artsci Calendar
+populateCalendar :: IO ()
```
david-yz-liu, line 50: Combine with above (see above comment)

app/WebParsing/ParseAll.hs
```diff
@@ -1,11 +1,14 @@
 module WebParsing.ParseAll
-     (parseAll) where
+     (parseCalendar, parseTimetable) where
```
david-yz-liu, line 2: Now that we're splitting things up, this file is no longer necessary. Move the definition of `parseCalendar` into `ArtSciParser`, and rename `getAllCourses` to `parseTimetable`. Then change the codebase to import from those modules rather than `ParseAll`.

david-yz-liu (approved):

PR #1350, opened 2023-06-19
Stack Installation Modification

harshkpatel:
## Motivation and Context

When running stack run database-courses, there would be a bus error, and after further investigation, that bus error was to do internally with the version of stack and nothing to do with the codebase itself.

## Your Changes

**Description**: Added a distinction between Windows and Mac/Linux installation links.

## Testing

I ran everything again, and now there is no longer a bus error and database-courses executes as its supposed to.

You mentioned earlier that I should comment about not using the homebrew installation, but I never had the homebrew installation and was still running into the error. After further deduction I think it was also a possibility it was because there was an issue with my path. The reason I ran into the issue with path was because when installing stack, I had to modify the path myself on my system, so instead I provided a link to a different installation that also prompts for appending to the path. I left the windows version on the same website as that website has an actual windows installer which modifies the path. I think this method takes care of both potential issues.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ See [CONTRIBUTING.md](https://github.com/Courseography/courseography/blob/master
 
 First install the following:
 
-1. Stack ([Windows](https://docs.haskellstack.org/en/stable/README/)) ([Mac/Linux](https://www.haskell.org/ghcup/))
+1. [Haskell](https://www.haskell.org/ghcup/)
 2. [Node.js](https://nodejs.org/en/download/)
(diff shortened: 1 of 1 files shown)


david-yz-liu (commented):
@harshkpatel okay great, I looked into it and I think it's good to switch to GHCup for all OSes, including Windows. Also switch the name from "Stack" to "Haskell" since GHCup really installs an entire suite of Haskell tools, not just Stack.

david-yz-liu (approved):

PR #1362, opened 2023-06-28
Optimize timetable database operations

harshkpatel:
## Motivation and Context

Right now for Meeting sections and times, the functionality is just inserting, which is very clunky. This can be done a lot better, which is what this pull request is taking care of. 

## Your Changes

**Description**: Modified Meeting data to either insert or update in the database after adding a uniqueness constraint. Modified Times data to first delete all the entries and then re-insert the new entries as there is no uniqueness constraint.

## Testing

I ran the test suite, as well as the function that is responsible for artsci timetable parsing and ran into no issues.

## Checklist


app/WebParsing/UtsgJsonParser.hs
```diff
@@ -11,10 +11,8 @@ import Data.Aeson.KeyMap as KM hiding (insert, map)
 import qualified Data.HashMap.Strict as HM
 import Data.Maybe (catMaybes)
 import qualified Data.Text as T
-import Database.Persist.Sqlite (Filter, SqlPersistM, deleteWhere, insert, insertMany_, runSqlite,
-                                selectFirst, (==.))
-import Database.Tables (Courses (..), EntityField (CoursesCode), MeetTime (..), Meeting (..),
-                        Times (..), buildTimes)
+import Database.Persist.Sqlite
```
david-yz-liu, line 14: Don't remove named imports. It's always better to be explicit with imports (moreover, in general you should not be changing code style like this).

harshkpatel:
I am not sure why the CircleCI tests are failing now as they were passing before, and I really don't think that me adding the specific imports would cause the tests to fail, so I am assuming that is an external issue.

david-yz-liu (approved):

david-yz-liu:
Thanks @harshkpatel, yes the failing test is unrelated to your work.

PR #1372, opened 2023-07-14
Update Artsci Calendar Parsing

harshkpatel:
## Motivation and Context

Arts and Science Calendar updates their formats yearly and as a result in order to continue to parse information for courseography to use, our parsers have to be updated accordingly.

## Your Changes

**Description**: Modified the specific HTML elements that were being detected in the parse.

## Testing

I ran the test suite and ran into no issues. I was also very careful about verifying each HTML element and triple checking about what was being added/removed from the list of HTML elements. I also looked at what was in the database and It looks like it is working properly now.

## Checklist


david-yz-liu (approved):
@harshkpatel great work!

PR #1378, opened 2023-07-27
Migrating front end tests from enzyme to RTL

MaryamTaj:
## Motivation and Context

It is important to migrate front-end tests from Enzyme to RTL because RTL tests increase our confidence in the application by testing components from a user's perspective (rather than focusing on implementation details). 

## Your Changes

**Description**:

- Using Jest's mocking feature, we modified Graph.test.js since it used Enzyme's shallow() for shallow rendering. 
- We added react-test-renderer as a dependency.
- Using react-test-renderer, we modified Edge.test.js, Button.test.js, Bool.test.js, InfoBox.test.js, GraphDropdown.test.js and Node.test.js, given that they also used Enzyme's shallow() for shallow rendering.

- [X] Test update (change that modifies or updates tests only)

## Testing

After modifying the targeted test suites, we ran `$yarn run test -- <path>`. This ensured that we maintained the tests' logic, despite modifying the implementation. 

## Checklist

- [X] I have verified that the CircleCI tests have passed.


david-yz-liu (commented):
@MaryamTaj @jaeyonglee3 great job! The tests are all good.

Regarding react-shallow-renderer: it's okay, but I do suspect that shallow rendering is more of an enzyme technique, and generally shouldn't be used in our tests. Why don't you try modifying the tests to avoid using shallow rendering?

david-yz-liu (approved):

david-yz-liu:
@MaryamTaj @jaeyonglee3 great job! I fixed the merge conflicts, which installed new dependencies (which is why the diff looks bigger than your original).

PR #1385, opened 2023-08-08
Update Timetable Parsing

harshkpatel:
## Motivation and Context

Artsci Timetable has migrated to a more general site : ttb.utoronto.ca, and as a result, all the functions associated with parsing information from timetable needs to be updated.

## Your Changes

**Description**: Updated the HTTP request, updated the parsers and removed some redundant and unnecessary code (see comment about this).

## Testing

I ran the test suite and there were no issues and I also checked that inserting as well as updating in the database worked correctly.

Just two things that I think can be removed/cleaned up.

There was a python script that handled something similar to what parse timetable is doing, but i feel like in an older iteration of courseography that is how it was done, but it seems outdated so I removed that script (I can add if back if you think it is necessary). I also removed the function associated with parsing from "courses.json". It seems that there were a couple of functions in the parseUTSGjson file such as parseallcourses, which seemed to be used in the past before this direct use of an http request so I removed those (one again I can add them back if you believe that have a purpose). Another note is that if the script were to be included, it would have to be a lot more complex because it would have to go through all the orgs in a different manner than what was included earlier because the orgs are now obtained from the courses schema in the database.

The second point being that I saw quite a bit of stuff in the Tables file that was noted TODO: redundant, so I can create a separate PR for those things because they seem similar in the manner that there were older version where that format for parsing made more sense, but now it is outdated. 

## Checklist


app/WebParsing/UtsgJsonParser.hs
```diff
@@ -1,83 +1,95 @@
...
-    resp <- liftIO $ decodeFileStrict coursesJson
-    let coursesLst :: Maybe (HM.HashMap T.Text (Maybe DB)) = resp
-        courseData = maybe [] (map dbData . catMaybes . HM.elems) coursesLst
-        (_, sections) = unzip courseData
-        meetings = concat sections
-    mapM_ insertMeeting meetings
+import Database.Persist.Sqlite (SqlPersistM, Update, Entity, entityKey, entityVal, deleteWhere, upsert,
+                                insert, insertMany_, runSqlite, selectFirst, selectList, (==.), (=.))
+import Database.Tables (Courses (..), EntityField (..), MeetTime (..), Meeting (..), buildTimes)
+import Network.HTTP.Conduit
```
david-yz-liu, line 13: it looks like you removed the explicit import (`simpleHttp`), but always use explicit imports

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -1,83 +1,95 @@
...
 
 -- | Parse all timetable data.
 parseTimetable :: IO ()
 parseTimetable = do
     orgs <- getOrgs
     runSqlite databasePath $ mapM_ insertAllMeetings orgs
 
--- | Return a list of all the "orgs" in FAS. These are the values which can be
---   passed to the timetable API with the "org" key.
+-- | Get all the orgs from the Courses schema in the database
```
david-yz-liu, line 21: "Courses" should not be capitalized. Also "schema" -> "table"

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -1,83 +1,95 @@
...
-    let coursesLst :: Maybe (HM.HashMap T.Text (Maybe DB)) = decode resp
-        courseData = maybe [] (map dbData . catMaybes . HM.elems) coursesLst
-        -- courseData contains courses and sections;
-        -- only sections are currently stored here.
-        (_, sections) = unzip courseData
-        meetings = concat sections
-    mapM_ insertMeeting meetings
+
+    -- Define the request
+    let reqBody = object [ "campuses" .= ([] :: [T.Text])
```
david-yz-liu, line 36: This value should be defined as a separate value in the config file, since it's really tied to the API, similar to the url being used. (You can make the value a function which takes in an `org`.) Similarly for the request headers.

Also, you need to make changes to both `Config.hs` and `DevelopmentConfig.hs`; only the latter is committed to the repo.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -1,83 +1,95 @@
...
+    -- Make the request
+    manager <- liftIO $ newManager tlsManagerSettings
+    response <- liftIO $ httpLbs request' manager
+    let respBody = responseBody response
+
+    -- decode the response
+    let meetings :: Maybe DBList = decode respBody
+    case meetings of
+        Nothing -> return ()
+        Just (DBList dbs) -> forM_ dbs $ \(DB meetTimes) -> mapM_ insertMeeting meetTimes
```
david-yz-liu, line 73: This is a bit inelegant. You should create a helper function `DBList -> [MeetTime]` so that you can do a `mapM_ insertMeeting $ ... dbList`

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -1,83 +1,95 @@
...
 
 -- | Insert or update a meeting and then delete
 --   and re-insert the corresponding Times into the database.
 insertMeeting :: MeetTime -> SqlPersistM ()
 insertMeeting (MeetTime meetingData meetingTime) = do
     -- Check that the meeting belongs to a course that exists
```
david-yz-liu, line 79: This function looks like it's confusing a *course* (like CSC110) with a *meeting* (like CSC110 LEC0101). Clarify what the purpose of this function is and then reword all of the comments, variable names, etc. to make them consistent.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -91,14 +103,26 @@ meetingUpdates m = [ MeetingCode =. meetingCode m
...
+  parseJSON = withObject "Expected an Object for DBList" $ \o -> do
+    maybeOrg <- o .:? "payload"
+    case maybeOrg of
+      Just payload -> do
+        pageableCourse <- payload .: "pageableCourse"
+        courses <- pageableCourse .: "courses"
+        dbList <- mapM parseCourse courses
+        return $ DBList dbList
+        where
+          parseCourse courseObj = do
```
david-yz-liu, line 122: This should really still be in the typeclass definition for `FromJSON DB`, so that you can use `mapM parseJSON courses` or something similar.

david-yz-liu (commented):
@harshkpatel good work. I just reviewed the changes in `UtsgJsonParser.hs` but not `Tables.hs` (I'll do that in another round after you've made the first set of changes.)

It's okay to delete the old Python script. And yes I'd be happy for you to work on simplifying some of the data types indicated in TODOs in a separate PR after getting this one finished. 😄

app/WebParsing/UtsgJsonParser.hs
harshkpatel: For making this change I had to add a dependency as there was a type mismatch which could only be resolved with adding the HTTP TYPES.

app/Database/Tables.hs
```diff
@@ -224,23 +219,38 @@ instance FromJSON SvgJSON
...
-                     newTitle
-                     newDescription
-                     newPrereqs
-                     newExclusions
-                     Nothing -- breadth
-                     Nothing -- distribution
-                     Nothing -- (Just prereqString)
-                     newCoreqs
-                     []
+    codeExtraChars <- o .:? "code" .!= "CSC?????"
```
david-yz-liu, line 222: This isn't great, we don't want `CSC???` to appear in the database anywhere. I'd just do `o .: "code"` and let the parsing fail if the key is missing.

app/Database/Tables.hs
```diff
@@ -224,23 +219,38 @@ instance FromJSON SvgJSON
 -- JSON encoding/decoding
 instance FromJSON Courses where
```
david-yz-liu, line 220: I started reviewing this function and then realized that I don't think we need this anymore? Since we're now parsing course information just from the calendar, but not from the timetable API. If so we can delete this code.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -91,14 +85,27 @@ meetingUpdates m = [ MeetingCode =. meetingCode m
...
-    parseJSON (Object o) = do
-      course <- parseJSON (Object o)
-      session :: T.Text <- o .:? "section" .!= "F"
-      meetingTimesMap :: HM.HashMap T.Text MeetTime <- o .:? "meetings" .!= HM.empty
-      let allMeetingsTimes = map (\m -> m {meetInfo = (meetInfo m) { meetingCode = coursesCode course, meetingSession = session}}) (HM.elems meetingTimesMap)
-      return $ DB (course, allMeetingsTimes)
-    parseJSON _ = fail "Invalid section"
+  parseJSON = withObject "Expected an Object for DB" $ \o -> do
+    course <- parseJSON (Object o)
+    session :: T.Text <- o .:? "sectionCode" .!= "F"
```
david-yz-liu, line 94: Yikes, even though this was the previous default I don't think this is good. We shouldn't "guess" a section code. If there isn't one, we should let the parser fail (i.e. by using `o .: "sectionCode"` instead of `.:?`)

david-yz-liu (commented):
@harshkpatel great work! I just left a few comments.

david-yz-liu (approved):

PR #1413, opened 2024-05-31
Button to deselect course in Sidebar

mimischly7:
## Proposed Changes

In the panel list of selected courses, currently the only way to deselect is by clicking
again on the course node or at the course code link within the panel (this has the FCE bug described in #1). The latter method is counter-intuitive and undesired. Instead, we want to add an **X** button next to each course code in the panel so that a given course gets deselected when the user clicks on that button (and not when they click on the course code link). The required changes in the `Sidebar` and `Graph` components.

In the `Sidebar` component, the `renderActiveCourses` function is responsible for rendering the side-bar panel of selected courses.
In terms of visuals, I reused the `Button` component to create the **X** ("remove") button requested.
Functionality wise, originally when you clicked on course *node*, the functionality was handled by the `nodeClick` function in the `Graph` component. This would change the `selected` state variable of the course node from true to false (and vice versa), update the FCE count, and update the list of courses shown in the panel.

What I did was essentially create a `nodeUnselect` unselect function that borrows logic from `nodeClick`, and which will be the event handler for when you click on the **X** next to a course code link. the former handling the "selection" of a course and the latter handling "unselection" of a course. The function, `nodeUnselect`, is passed down from the `Graph` component to the child `Sidebar` component as an `xClick` prop and in turn passed from the `Sidebar` component to the new `Button` component as the `mouseDown` prop.

Finally, to remove the ability to deselect a course by clicking on the couse link, I removed the usage of the `handleCourseClick` function (defined in the `Graph` component) from the `renderActiveCourses` - where it was being passed as the `onClick` prop to the `div` element corresponding to the course code link.

*****
## Round 2
Although for most courses it is true that *course-label in lowercase is the same as the course-id*, it is not true for course-groups like `MAT221/223/240 (Alg1)` or `CSC263/265`, where the (lowercase) course labels are `mat221/223/240 (alg1)` or `csc263/265` but the course-id's (the id's of the course node elements) are `mat221223240alg1` and `csc263265`. This mismatch is clearly problematic, leading to the following error:
*****
course-id-mismatch-prob
*****
It is `undefined` because the `nodeUnselect` function is trying to find a course by the course-label instead of the course id.

*Luckily*, the `Sidebar` component has a `this.props.courses` props; this is
a 2D array where each inner array is of size two and corresponds to a course: the first element is the course-id and the second element in the corresponding course label! It looks like this: `[['csc438', 'CSC438'], ..., ['csc263265', 'CSC263/265'], ...]`. This is exactly what we need, because given a course label (which is `unique`) we can determine the corresponding course code. Therefore, **I will define a helper function that given a course-label retrieves the corresponding course-id**, and THAT (the retrieved course-id) is what I will pass as input to the `xClick` event handler.

**Note**: We cannot use string manipulation to convert from course-label to course-id because composite groups have a label of `(<groupname>)`.

<details>

Screenshot 2024-05-24 at 1 10 51 PM
</details>

## Type of Change

| 🎨 _User interface change_ (change to user interface; provide screenshots)              |     X     |

## Checklist

## Questions and Comments

 - [x] In the `Sidebar` component, `this.state.results` appears to always be an empty array (no matter where I `console.log` it). Why?
 - [x] In the context of the `nodeClick` event handler defined in the     `Graph` component, `console.log(event)` and `console.log(event.currentTarget)` are one line apart but in the first log event.currentTarget is null while on the second log it's perfectly there. Here is the output:
event-target-paradox

[task-2.md](https://github.com/user-attachments/files/15519607/task-2.md)


js/components/graph/Graph.js
```diff
@@ -422,6 +422,21 @@ export class Graph extends React.Component {
     }
   }
 
+  nodeUnselect = courseId => {
+    const courseLabelArray = this.state.nodesJSON[courseId].text
+    const courseLabel = courseLabelArray[courseLabelArray.length - 1].text
+    const wasSelected = this.state.nodesStatus[courseId].selected
+    const temp = this.state.selectedNodes
+
+    wasSelected ? this.toggleSelection(courseId) : {}
```
david-yz-liu, line 431: In general, don't use ternary ifs with side effects; use if statements instead.

In this specific case, combine this with the if statement below.

js/components/graph/Graph.js
```diff
@@ -625,6 +640,13 @@ export class Graph extends React.Component {
     })
   }
 
+  courseLinkMouseClick = courseCode => {
```
david-yz-liu, line 643: No need for the word "mouse" here (this is true in general). In general when choosing names, look at existing functions and try to follow the same naming style.

In this specific case, however, I don't think the old `handleCourseClick` function is used any more, so you should actually just replace it with this function.

js/components/graph/Sidebar.js
```diff
@@ -87,10 +102,16 @@ export default class Sidebar extends React.Component {
             <div
               key={`active ${course}`}
               data-testid={`test ${course}`}
-              onClick={() => this.props.courseClick(course)}
               className="course-selection"
+              onClick={() => {
+                this.props.courseLinkClick(this.courseIdFromLabel(course))
```
david-yz-liu, line 107: see my above comment about `courseLinkClick` vs. `courseClick`

david-yz-liu (commented):
@mimischly7 good job. I left a few comments. In response to your two questions:

1. This is a good catch and can be addressed in a separate PR.
2. I'm not sure why you would be observing this and would need to see more concrete code to investigate further (not on this branch).

mimischly7:
Screenshot 2024-06-12 at 1 54 35 AM
In what reasonable way could the highlighted code evaluate to false? Isn't this prop always supposed to be a function?

js/components/graph/Graph.js
```diff
@@ -422,6 +422,20 @@ export class Graph extends React.Component {
     }
   }
 
+  nodeUnselect = courseId => {
+    const courseLabelArray = this.state.nodesJSON[courseId].text
+    const courseLabel = courseLabelArray[courseLabelArray.length - 1].text
+    const wasSelected = this.state.nodesStatus[courseId].selected
+    const temp = this.state.selectedNodes
+
+    if (typeof this.props.incrementFCECount === "function" && wasSelected) {
```
david-yz-liu, line 431: @mimischly7 in response to your question: this is a common pattern when the prop might be _omitted_ entirely, so it ends up being `undefined`.

js/components/graph/Graph.js
```diff
@@ -473,23 +487,11 @@ export class Graph extends React.Component {
...
-        selectedNodes: new Set(temp.filter(course => course !== courseLabel)),
-      })
-      this.props.incrementFCECount(-0.5)
-    } else {
-      this.setState({
-        selectedNodes: new Set([...temp, courseLabel]),
-      })
-      this.props.incrementFCECount(0.5)
-    }
+  handleCourseClick = courseCode => {
```
david-yz-liu, line 490: This is good, but make sure to update the function documentation as well

js/components/graph/Sidebar.js
```diff
@@ -30,6 +31,20 @@ export default class Sidebar extends React.Component {
       .sort()
   }
 
+  /**
+   * Given a course label return the id of the corresponding Node component.
+   * e.g. "CSC263/265" will resolve to "csc263265"
+   * @return {string} course node id
```
david-yz-liu, line 37: should also describe when `null` is returned

js/components/graph/Sidebar.js
```diff
@@ -87,10 +102,16 @@ export default class Sidebar extends React.Component {
             <div
               key={`active ${course}`}
               data-testid={`test ${course}`}
-              onClick={() => this.props.courseClick(course)}
               className="course-selection"
+              onClick={() => {
```
david-yz-liu, line 106: This is good, but don't change the relative prop position (this prop used to be above `className`)

david-yz-liu (commented):
@mimischly7 In addition to my inline comments, it appears there is a failing test.

mimischly7:
Just as a heads up, I removed a test because it no longer applies (since we no longer want clicking on a sidebar course to remove that course, that's the role of the X button now, which is tested).

david-yz-liu (approved):

PR #1421, opened 2024-06-14
Graph Generation - Handle Bad Courses Input (improvement of PR#1419)

mimischly7:
## Proposed Changes

In the Generate functionality, if no courses are found (e.g. an invalid course code is input), currently nothing happens. Instead, when no valid courses are inputted, we ant to have some kind of message is displayed to the user, and nothing in the graph to be drawn.

## How it currently works:
1. When we press the "Generate Graph" button on the `/generate` page, the `handleSubmit` function in `js/components/generate/generate.js` is triggered.
2. This handler makes a `PUT` request to the `/graph-generate` endpoint (with data including the user's selected courses):
```
fetch("/graph-generate", putData)
```
3. This (through the `Routes.hs` file) is mapped to the `Controllers.Graph.findAndSavePrereqsResponse`, which decodes the request body and passes it to a helper `findAndSavePrereqResponse` that does all the necesarry computations and returns the appropriate course information.
```
findAndSavePrereqsResponse :: ServerPart Response
findAndSavePrereqsResponse = do
    body <- getBody
    let coursesOptions :: CourseGraphOptions = fromJust $ decode body
    liftIO $ generateAndSavePrereqResponse coursesOptions
```

4. This is sent back as a response, fulfilling JS's `fetch` promise, and JS proceeds to handle the response data with a chain of promises.

Completely disregard the work from PR#1419. Here is what we are actually going to do. First, the result look like this.
Screenshot 2024-06-17 at 11 49 39 PM

In `generate.js`, in the handleSubmit, we simply need to look at the response produced by Haskell:

Here is what the handling looks like:

``` javascript
fetch("/graph-generate", putData)
      .then(res => res.json())
      .then(data => {
        const returnedCourses = data.texts.map(t => t.text)
        const missingCourses = submittedCourses.filter(
          c => !returnedCourses.includes(c)
        )
        if (missingCourses.length !== 0) {
          this.setState({ showWarning: true })
          this.setState({ invalidCourses: missingCourses })
        }
```

 - `submittedCourses` is the list of all course codes submitted by the user (valid and invalid).
 - `returnedCourses` is the list of all course codes that appear in the graph.
 - If  `submittedCourses` contains a course code that is not contained in  `returnedCourses` it means that that course code is invalid (hence we calculate `missingCourses` based upon this criterion).  Therefore, we set the (new) state variables `showWarning` and `invalidCourses` appropriately.
 - *However*, we do not break out of the Promise chain, we allow any subsequent calculations to take place. This has several benefits:
  - If some courses are valid but some are invalid, the warning box will be shown for the invalid courses, but the graph will still be generated for the valid ones.
  - Replacing a valid course with an invalid one will update the graph shown. For example if the initial input is `[CSC110Y1, CSC111H1]` (both valid course codes), and a valid graph is generated, the if the user changes the input to `[CSC110Y1, csc121h1]` and submits again that graph will be updated (removing the `CSC111H1`-related paths).

**How does the error box get shown?**

Assume the user inputted some invalid courses, so that the state variables `showWarning` and `invalidCourses` were set appropriately. In the `return` of the component, here is what happens:

``` javascript
return (
      {...}
        {this.state.showWarning && (
          <ErrorMessage
            message={`The courses [${this.state.invalidCourses}] were invalid! Please check your input.`}
            onClose={() => this.setState({ showWarning: false, invalidCourses: [] })}
          />
        )}
        {...}
      )
```
This is conditional rendering: if the `showWarning` state variable is true, then a (*new*) `ErrorMessage` component will be rendered, which will utilize the `invalidCourses` state variable to tell the user which of the inputted courses were invalid.

Importantly, when the user clicks to close the warning box, the function `() => this.setState({ showWarning: false, invalidCourses: [] })` gets executed, resetting the invalid-course-related state variables -- **otherwise the warning box will NOT leave the screen**.

Here is the `<ErrorMessage>` component for your reference.
``` javascript
function ErrorMessage({ message, onClose }) {
  return (
    <div className="error-overlay">
      <div className="error-message-box">
        <p>{message}</p>
        <button onClick={onClose} className="error-close-button">
          Close
        </button>
      </div>
    </div>
  )
}
```

I defined some styles in `app.scss` to make the box look good.

In `generate.js`, in the handleSubmit, we simply need to look at the response produced by Haskell:

Here is what the handling looks like:

``` javascript
fetch("/graph-generate", putData)
      .then(res => res.json())
      .then(data => {
        const returnedCourses = data.texts.map(t => t.text)
        const missingCourses = submittedCourses.filter(
          c => !returnedCourses.includes(c)
        )
        if (missingCourses.length !== 0) {
          this.setState({ showWarning: true })
          this.setState({ invalidCourses: missingCourses })
        }
```

 - `submittedCourses` is the list of all course codes submitted by the user (valid and invalid).
 - `returnedCourses` is the list of all course codes that appear in the graph.
 - If  `submittedCourses` contains a course code that is not contained in  `returnedCourses` it means that that course code is invalid (hence we calculate `missingCourses` based upon this criterion).  Therefore, we set the (new) state variables `showWarning` and `invalidCourses` appropriately.
 - *However*, we do not break out of the Promise chain, we allow any subsequent calculations to take place. This has several benefits:
  - If some courses are valid but some are invalid, the warning box will be shown for the invalid courses, but the graph will still be generated for the valid ones.
  - Replacing a valid course with an invalid one will update the graph shown. For example if the initial input is `[CSC110Y1, CSC111H1]` (both valid course codes), and a valid graph is generated, the if the user changes the input to `[CSC110Y1, csc121h1]` and submits again that graph will be updated (removing the `CSC111H1`-related paths).

**How does the error box get shown?**

Assume the user inputted some invalid courses, so that the state variables `showWarning` and `invalidCourses` were set appropriately. In the `return` of the component, here is what happens:

``` javascript
return (
      {...}
        {this.state.showWarning && (
          <ErrorMessage
            message={`The courses [${this.state.invalidCourses}] were invalid! Please check your input.`}
            onClose={() => this.setState({ showWarning: false, invalidCourses: [] })}
          />
        )}
        {...}
      )
```
This is conditional rendering: if the `showWarning` state variable is true, then a (*new*) `ErrorMessage` component will be rendered, which will utilize the `invalidCourses` state variable to tell the user which of the inputted courses were invalid.

Importantly, when the user clicks to close the warning box, the function `() => this.setState({ showWarning: false, invalidCourses: [] })` gets executed, resetting the invalid-course-related state variables -- **otherwise the warning box will NOT leave the screen**.

Here is the `<ErrorMessage>` component for your reference.
``` javascript
function ErrorMessage({ message, onClose }) {
  return (
    <div className="error-overlay">
      <div className="error-message-box">
        <p>{message}</p>
        <button onClick={onClose} className="error-close-button">
          Close
        </button>
      </div>
    </div>
  )
}
```

I defined some styles in `app.scss` to make the box look good.

***
### Final Stage Summary (*June 22*)
***
- Refactored the `js/components/generate.js` file into two separate files:
  1. `GenerateForm.js`, which hosts the actual component (that was originally in `generate.js`), and
  2. `generate.js` which imports the React component from `GenerateForm.js` and renders it.

  The reason I did this separation (i.e. why I believe the component definition and component rendering should be in the same file):
  - When making tests about a component, you have to import that component. If the file where the component is defined also renders that component, e.g. `ReactDOM.render(<GenerateForm />, document.getElementById("generateRoot"))` (this what the `generate.js` file had), this could lead to problems because the container element on top of which your component is not being rendered may not be defined, which is normal if you don't really care about that parent (i.e. testing the functionality of the component does not in some way depend on that parent container). This is what was happening in my case, getting the error: `React Uncaught Error: Target container is not a DOM element`. So I refactored it and in the tests I am rendering the component itself and nothing else.

- Created a test suite for the generate functionality (after setting up the testing file structure in `js/components/generate/`). It took several iterations to make it DRY and to follow RTL conventions, but here is the main logic and ideas:

  The goal was to test the warning modal for invalid course inputs. This required mocking the `/graph-generate` endpoint. A challenge was that a single mocking statement (e.g. as seen in the `TestSetup.js` file) was insufficient because I needed to check various different inputs that would have different outputs. For this reason, I create a JSON file with sample request-response key-value pairs, and `mocker` functions that produces a **double Promise**; the type of object that is returned when doing `fetch("....")`. This mock "generator" is fed inputs from the keys of the JSON. Within the test, I used that `mocker` function to reassign `global.fetch`.

  The tests cover ensuring that:
    - the warning modal shows up (with the *correct* message) for a variety of different invalid (or semi-invalid) inputs
    - given a fully valid input, the modal does not appear
    - the two close buttons ("Okay" and "X") work correctly
    - The graph is updated correctly - e.g. if you enter a valid course but then you change a letter of it (making it invalid), then everything related to that course in the graph should dissappears

- Fixed minor bug in logic of warning modal

  When the course input was an empty string, the waning message would corectly be `Cannot generate graph – no courses entered!`. But what if the user entered a bunch of spaces, like "    ". Ideally we would want the same behavior, but the code views these spaces as an attempt to specify a course, producing the warning message  `The course        was invalid! Please check your input.` The message shown in this case is determined by `this.state.courses` (the original string passed by the user), so what I do is (when the user submits) update that state variable by removing all spaces from that string. This state update is a bit subtle because it will not be visible in the function where it is being updated (due to the asynchronous nature of updates) but will be there in the next render, which is what we needed.

## Type of Change

| 🎨 _User interface change_ (change to user interface; provide screenshots)              |    X      |
| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |     X     |

## Checklist

## Questions and Comments


js/components/generate/generate.js
```diff
@@ -76,6 +81,15 @@ class GenerateForm extends React.Component {
     fetch("/graph-generate", putData)
       .then(res => res.json())
       .then(data => {
+        const returnedCourses = data.texts.map(t => t.text)
+        const missingCourses = submittedCourses.filter(
+          c => !returnedCourses.includes(c)
+        )
+        if (missingCourses.length !== 0) {
+          this.setState({ showWarning: true })
+          this.setState({ invalidCourses: missingCourses })
```
david-yz-liu, line 90: combine with preceding `setState` call

js/components/generate/generate.js
```diff
@@ -239,6 +253,12 @@ class GenerateForm extends React.Component {
   render() {
     return (
       <div style={{ display: "flex", flexDirection: "row", height: "100%" }}>
+        {this.state.showWarning && (
+          <ErrorMessage
+            message={`The courses [${this.state.invalidCourses}] were invalid! Please check your input.`}
```
david-yz-liu, line 258: Have separate messages for singular vs. plural course counts. Note that in the singular form, "was invalid" should be used.

js/components/generate/generate.js
```diff
@@ -239,6 +253,12 @@ class GenerateForm extends React.Component {
...
+            onClose={() => this.setState({ showWarning: false, invalidCourses: [] })}
```
david-yz-liu, line 259: There isn't a need to reset `invalidCourses`

style/app.scss
```diff
@@ -2250,3 +2250,43 @@ main-filter input {
 #nav-generate {
   white-space: nowrap;
 }
+
+.error-overlay {
```
david-yz-liu, line 2254: Having some custom CSS is fine, but look to match it with the existing CSS for modals (the course info modal, the export modal)

js/components/common/react_modal.js.jsx
```diff
@@ -979,4 +979,36 @@ function convertTimeToString(time) {
...
+      className="error-modal-class"
+      overlayClassName="error-overlay"
+      isOpen={isOpen}
+      ariaHideApp={false}
+      onRequestClose={onClose}
+      contentLabel="Pop-up warning for invalid courses"
+    >
+      <div className="modal-header">
+        {title}
+        <div className="button-container">
```
david-yz-liu, line 1002: I like that you're following an existing style here. However, I think buttons with icons/symbols are most appropriate for the positioning here. An "X" could be okay.

I _also_ liked the button you had at the bottom of the modal originally. Perhaps add that (in addition to the "X") with a label that says "Okay".

david-yz-liu (commented):
@mimischly7 great work. I left one inline comment. Additionally, please (1) update the changelog, and (2) add test cases for this new behaviour. I recommend mocking the server response with different "returned course" values.

js/components/common/react_modal.js.jsx
```diff
@@ -979,4 +979,44 @@ function convertTimeToString(time) {
...
+      overlayClassName="error-overlay"
+      isOpen={isOpen}
+      ariaHideApp={false}
+      contentLabel="Pop-up warning for invalid courses"
+    >
+      <div className="modal-header">
+        {title}
+        <div className="button-container">
+          <button onClick={onClose} className="error-close-button-x" type="button">
+            <b> X </b>
```
david-yz-liu, line 1003: Don't include spaces or `<b>` tags; instead, use CSS styling as needed

js/components/common/react_modal.js.jsx
```diff
@@ -979,4 +979,44 @@ function convertTimeToString(time) {
...
+          </button>
+        </div>
+      </div>
+
+      <div className="modal-body">
+        <div>{message}</div>
+
+        <div className="button-container">
+          <button onClick={onClose} className="error-close-button">
+            {" "}
```
david-yz-liu, line 1013: Similar to my above comment, don't use spaces for positioning/styling. Instead, use CSS styles.

js/components/generate/generate.js
```diff
@@ -236,9 +248,34 @@ class GenerateForm extends React.Component {
...
+  /**
+   * Produce an appropriate warning message string in the case that no courses have been entered
+   * or that one or more invalid courses have been entered.
+   * @param {string} coursesString - The string version of the inputted courses
+   * @param {string[]} invalidCourses - The array of invalid course codes
+   * @returns {string} The warning message string.
+   */
+  computeMessage(coursesString, invalidCourses) {
+    if (!coursesString.length) {
+      return "Cannot generate graph -- no courses entered!"
```
david-yz-liu, line 260: Replace the `--` with an endash, –. (*Warning*: copy that character directly, don't try to type it in yourself. It's different than the regular hyphen that appears on your keyboard.)

js/components/common/react_modal.js.jsx
```diff
@@ -979,4 +979,43 @@ function convertTimeToString(time) {
   return ((time === 12.5 ? 12 : Math.floor(time)) % 12) + ":30" + meridiem
 }
 
-export { CourseModal, MapModal, FocusModal }
+/**
+ * For use in Generate - a warning modal for when the user enters one or more invalid courses
+ *
+ * Props:
```
david-yz-liu, line 985: Add documentation for the `title` prop as well

js/components/common/react_modal.js.jsx
```diff
@@ -979,4 +979,43 @@ function convertTimeToString(time) {
   return ((time === 12.5 ? 12 : Math.floor(time)) % 12) + ":30" + meridiem
 }
 
-export { CourseModal, MapModal, FocusModal }
+/**
+ * For use in Generate - a warning modal for when the user enters one or more invalid courses
```
david-yz-liu, line 983: should be amended to include the case that the user doesn't input any courses at all; but you can say something more generic like "user's course selection is invalid" or something

js/components/generate/GenerateForm.js
```diff
@@ -0,0 +1,397 @@
...
+      [name]: value,
+    })
+  }
+
+  handleSubmit = event => {
+    event.preventDefault()
+
+    // NOTE: Due to the aynchronous nature of state update, this does not have any effect in the running function,
+    // but plays a role in the next render
+    this.setState({ courses: this.state.courses.replaceAll(" ", "") })
```
david-yz-liu, line 58: It's good to disallow inputs that are just whitespace, but the `setState` is not necessary here. Instead add something to the check below on line 60 (e.g. a `.trim()`)

js/components/generate/GenerateForm.js
```diff
@@ -0,0 +1,397 @@
...
+
+    if (!this.state.courses.length) {
+      this.setState({ showWarning: true, invalidCourses: [] })
```
david-yz-liu, line 61: include a `return` after this line, since the rest of the code shouldn't execute

js/components/generate/GenerateForm.js
```diff
@@ -0,0 +1,397 @@
...
+
+  /**
+   * Produce an appropriate warning message string in the case that no courses have been entered
+   * or that one or more invalid courses have been entered.
+   * @param {string} coursesString - The string version of the inputted courses
+   * @param {string[]} invalidCourses - The array of invalid course codes
+   * @returns {string} The warning message string.
+   */
+  computeMessage(coursesString, invalidCourses) {
+    if (!coursesString.length) {
```
david-yz-liu, line 264: You don't need to include this check (or the `coursesString` argument). Just check whether `invalidCourses.length === 0` as the "blank" case. This works in tandem with my above comment.

js/components/generate/GenerateForm.js
```diff
@@ -0,0 +1,397 @@
...
+      return `The course ${invalidCourses} was invalid! Please check your input.`
+    } else {
+      return `The courses [${invalidCourses}] were invalid! Please check your input.`
+    }
+  }
+
+  render() {
+    return (
+      <div style={{ display: "flex", flexDirection: "row", height: "100%" }}>
+        {/* {console.log(this.state.showWarning, "~"+this.state.courses+"~", this.state.invalidCourses)} */}
```
david-yz-liu, line 278: This should be removed. This is one of the things you should be looking for during your self-review of your changes.

js/components/generate/__tests__/generate.test.js
```diff
@@ -0,0 +1,129 @@
...
+  ])(".$coursesInputText", async ({ coursesInputText, expectedWarning }) => {
+    global.fetch = mocker(coursesInputText)
+    const coursesInputField = screen.getByPlaceholderText("e.g., CSC207H1, CSC324H1")
+    fireEvent.change(coursesInputField, { target: { value: coursesInputText } })
+    expect(screen.queryByText("Invalid Course Input")).toBeNull()
+    const genButton = screen.getByText("Generate Graph")
+    fireEvent.click(genButton)
+    const warningModal = (await screen.findByText("Invalid Course Input")).parentElement
+    expect(warningModal).not.toBeNull()
+    // screen.debug(warningModal)
```
david-yz-liu, line 56: Remove debugging statements (throughout)

js/components/generate/generate.js
```diff
@@ -1,354 +1,6 @@
 import React from "react"
```
david-yz-liu, line 1: Seems like this import is no longer necessary

js/components/generate/__tests__/generate.test.js
```diff
@@ -0,0 +1,129 @@
+import React from "react"
+import GenerateForm from "../GenerateForm.js"
+import { screen, render, fireEvent, waitFor, within } from "@testing-library/react"
+import courseToResp from "../__mocks__/sample_responses.json"
+
+const origFetch = global.fetch
+
+const mocker = courseInput => {
```
david-yz-liu, line 8: Having the `sample_responses.json` file is good. However, you can improve on the mocking structure by using `fetchMock` in `setupTests.js`, like the existing mocking. Note that the "response" argument to `fetchMock.mock`/`fetchMock.get`/`fetchMock.put`/etc. can be a *function* that takes information about the request and then determines what the response should be. So, you can keep the json file the way you have it, but use it in a single `fetchMock.put` call and dynamically determine what to return. See https://www.wheresrhys.co.uk/fetch-mock/#api-mockingmock_response.

js/components/generate/GenerateForm.js
mimischly7: The problem is the following scenario: 
1. A user enters a good input, e.g. `CSC110Y1, MAT237Y1`
2. Presses generate, generating a nice graph
3. makes the input blank ('')
4. Presses generate button again

**With a return, the old graph will stay, but without `return` it will be *correctly* rerendered to blank.** 

I thought this was the behavior we were after, although I do understand it is an additional server request. Let me know what is preferred.

david-yz-liu: Right, it's fine to keep the old graph displayed. No need to rerender.

js/components/generate/generate.js
mimischly7: Surprisingly JSX syntax cannot be used without importing React. Surprisingly because I read that this should be possible from React 18 (with JSX transform), so I'm not sure why it still needs it, I can look into it.

***
Screenshot 2024-06-24 at 12 21 38 AM
***
Screenshot 2024-06-24 at 12 21 30 AM
***

david-yz-liu: @mimischly7 probably needs the `.jsx` extension at the end of this filename?

mimischly7: I left you a slack on this

js/components/generate/GenerateForm.js
david-yz-liu: @mimischly7 I think you missed the point of my reply. You should add the `return` I asked you to add.

js/components/generate/GenerateForm.js
```diff
@@ -65,17 +66,35 @@ class GenerateForm extends React.Component {
...
+    let submittedCourses = data["courses"]
+
     const putData = {
       method: "PUT",
       headers: {
         "Content-Type": "application/json",
       },
       body: JSON.stringify(data), // We send data in JSON format
     }
-
```
david-yz-liu, line 75: revert this change

david-yz-liu (commented):
@mimischly7 looks good, I just left a few minor comments (one following up from the last review)

js/components/generate/GenerateForm.js
mimischly7: apologies for this, just fixed it

david-yz-liu (approved):

PR #1429, opened 2024-07-08
Upgrade courseography to React v18 -- Take 1

mimischly7:
## Proposed Changes

This PR aims to update `courseography` to React v18.

In summary, it includes:
1. Updating `package.json` with version 18 for all react-related packages, as well as with upgrades of other packages to allow compatibility with React v18.
2. Updating depracated API's (ReactDOM.render, named refs, e.t.c.) in `js/components`
3. Updating the test setup to include needed NodeJS globals (see ag-grid related problem below)
4. Updating babel configuration (both natively and in JEST context) to finetune the inclusion of  in transpilation (needed by  react-leaflet v4, see explanation below)

Now follows a more detailed explanation of the changed, grouped as per the four categories above:

### Updating depracated API's:
React v18 introduce a new API for rendering components on the DOM, that is required if we want to take advantge of the new concurrent rendering features. As such I replaced `ReactDOM.render(...)` with the new `createRoot(...).render()` API. Also, named refs (this.refs.<refname> ... <comp ref="<refname>">) has been deprecated, so I refactored to use the suggested `createRef` function.

### Manually importing NodeJS globals in `js/setupTests.js`
After upgrading to React v18 (and running `yarn cache clean` --> `yarn install`), and with the latest version of ag-grid, 
Importing (`AgGridReact`) from `"ag-grid-react"` in the `jest` tests leads raises error
 ```
 ReferenceError: TextEncoder is not defined
 ```
 **only in jest**. In the browser the application runs completely fine, but the tests throw
 ```
 ReferenceError: TextEncoder is not defined

      4 | import L from "leaflet"
      5 | import { getCourse, getPost } from "../common/utils"
    > 6 | import { AgGridReact } from "ag-grid-react"
        | ^
      7 |
      8 | const DAY_TO_INT = {
      9 |   0: "Monday",
  ```
  I think the problem has to do with the fact that jest creates a "sandbox" DOM environment to run the tests,
  which may not support all the features supported by a modern web browser. In this case, it appears that jsdom
  is missing the `TextEncoder` and `TextDecoder` libraries (from "util").

  Note I say `jdom` because we have configured `jest` to use `jsdom` for the test environment with the line
  ```
  testEnvironment: "jsdom"
  ```
  from `jest.config.js`.

  But why did this not cause a problem in React 17?

  Let's follow the stack trace:
  ```
  at .yarn/__virtual__/react-dom-virtual-4f7a30f7ec/0/cache/react-dom-npm-18.3.1-a805663f38-298954ecd8.zip/node_modules/react-dom/cjs/react-dom-server.browser.development.js:145:19
     at Object.<anonymous> (.yarn/__virtual__/react-dom-virtual-4f7a30f7ec/0/cache/react-dom-npm-18.3.1-a805663f38-298954ecd8.zip/node_modules/react-dom/cjs/react-dom-server.browser.development.js:7013:5)
     at Object.<anonymous> (.yarn/__virtual__/react-dom-virtual-4f7a30f7ec/0/cache/react-dom-npm-18.3.1-a805663f38-298954ecd8.zip/node_modules/react-dom/server.browser.js:9:7)
     at Object.<anonymous> (.yarn/__virtual__/ag-grid-react-virtual-733c540865/0/cache/ag-grid-react-npm-29.3.5-6970ed3f01-e223712504.zip/node_modules/ag-grid-react/lib/legacy/legacyReactComponent.js:25:16)
     at Object.<anonymous> (.yarn/__virtual__/ag-grid-react-virtual-733c540865/0/cache/ag-grid-react-npm-29.3.5-6970ed3f01-e223712504.zip/node_modules/ag-grid-react/lib/legacy/agGridReactLegacy.js:54:30)
     at Object.<anonymous> (.yarn/__virtual__/ag-grid-react-virtual-733c540865/0/cache/ag-grid-react-npm-29.3.5-6970ed3f01-e223712504.zip/node_modules/ag-grid-react/lib/agGridReact.js:49:27)
     at Object.<anonymous> (.yarn/__virtual__/ag-grid-react-virtual-733c540865/0/cache/ag-grid-react-npm-29.3.5-6970ed3f01-e223712504.zip/node_modules/ag-grid-react/main.js:1:125)
     at Object.require (js/components/common/react_modal.js.jsx:6:1)
     at Object.require (js/components/graph/Graph.js:3:1)
     at Object.require (js/components/graph/__tests__/findRelationship.test.js:1:1)
  ```
  It appears that the import for ag-grid eventually uses `react-dom`, which we updated to 18.3.1. So my guess is that some code from ag-grid library causes this new version react-dom to use TextEncoder; whereas it didnt need it in the previous version of react-dom.

  Anyways, this issue does not seem to be resolved, but here is a nice fix. In the test-setup file for our jest tests (in our case `js/setupTests.js`), add the following:
  ``` javascript
  // This is needed with upgrade to React 18
  global.TextEncoder = TextEncoder;
  global.TextDecoder = TextDecoder;
  ```
  , essentially giving these needed libraries to our scope.

### React Leaflet
The `react-leaflet` package had to be upgraded to v4 (from v3) to allow support for React v18. However, `react-leaflet v4` has the *breaking change* of being compiled into `ES` instead of `CommonJS`. As per the `react-leaflet` [changelog](https://github.com/PaulLeCam/react-leaflet/blob/master/CHANGELOG.md) for v4, *ES2020 is the new compilation target version*.

Since JEST uses Node (which uses `CommonJS`), this package would have to be transpiled (as per JEST jargon, "transformed") (from ES to CommonJS) using Babel. The problem is that the current jest configuration (`jest.config.js`) by default excludes all packages in `node_modules` from being transformed. Let's look at this more closely.

The `jest.config.js` file includes
```js
// A map from regular expressions to paths to transformers
transform: {
  "\\.jsx?$": "babel-jest",
},
```
, which essentially defines for each type file, how it should be transformed (transpiled). As per the [JEST docs](https://jestjs.io/docs/code-transformation#defaults), `"babel-jest"` is the default *transformer* that comes out of the box with JEST, although in this case we have explicitly (since our regex pattern of files to be transpiled is different than the default one). This transformer will load the project's babel configuration (`babel.config.js` and `.babelrc` files) and transform any files matching the regex.

So where is the problem?

The problem is that by default, **JEST will exclude `node_modules` from transpilation/transformation**. This means that if any of your packages is in `ESM`, you will get an error like
```
ERROR: Jest encountered an unexpected token
SyntaxError: Unexpected token 'export'
(full stack trace attached below)
```
,which reveals that this is an ESM module and has not been transpiled to CommonJS -- thus it is unusable by JEST which can only handle `cjs` files (support for `ESM` is experimental).

How do we solve this problem? We need to find a way to **include the react-leaflet (`ESM`) module in the set of files that will be transformed**.

It turns out that the `transformIgnorePatterns` has to be modified in order to allow `node_modules` contents to be transpiled. The `transformIgnorePatterns` must point to an array of regular expressions, each corresponding to a pattern of files that should be **excluded** from the transpilation process. The default behavior of excluding the `node_modules` from can be explicitly expressed in `jest.config.js` as
```js
transformIgnorePatterns: ["/node_modules/"]
```

So we must find a way to not be so strict (excluding everything in `node_modules`) and specify that we want some modules in there to actually be transpiled.

***
Before I proceed to explain the two approaches I took (the first a failure the second a success), I want to describe a problem that took a considerable time to resolve, related to Babel.

Starting with *v7*, `Babel` offers you [two ways of defining configuration](https://babeljs.io/docs/config-files):
- Project-wide configuration, using a `babel.config.*` files, and
- File-relative configuration, using `.babelrc.*` files

File-relative configuration (`.babelrc.*`) allows creation of independent configurations for subsections of a package, but **`.babelrc.*` files only appy to files within their own package**. This is crucial, because it means that any configuration we place inside our `.babelrc` (at the root of our project, for example) will ***NOT*** apply to packages inside `node_modules`, since **each such package will have its own `package.json` file**!

With Project-wide configuration (using `babel.config.*`), however, the **defined configuration applies broadly, including `node_modules`** (and *symlinked* packages)! So the *plugins* and *presets* you have defined can apply to modules in `node_modules`.

Therefore, I turned the `.babelrc.json` file of the project (the only source of `babel` configuration in the project, aside from `webpack.common.js`, but config in there is irrelevant for us who are currently concerned with JEST) to a `babel.config.js` file (with exactly the same content), allowing it to apply to the `node_modules` directory.

***NOTE***: What I described above is completely independent of the `jest` configuration option relating to transformation `transformIgnorePatterns`. For example, we can have `transformIgnorePatterns = []`, meaning that we want every matching file (from the `transform` option regex) to be transpiled (including `node_modules`), but if we are using file-relative configuration for babel then files in `node_modules` will still NOT be transpiled -- although the jest config wants them to, the babel config does not permit it. The vice versa is also possible.

Here are some sources that helped me understand this:
- https://github.com/gravitational/teleport/issues/33810
- **!!!** [loganfsmyth's comment](https://github.com/jestjs/jest/issues/6053#issuecomment-383632515)
- https://github.com/jestjs/jest/issues/6229#issuecomment-403539460
- [Exact issue in `react-leaflet` repo](https://github.com/PaulLeCam/react-leaflet/issues/977), and [slutske22's comment](https://github.com/PaulLeCam/react-leaflet/issues/977#issuecomment-1112225145)
- This [stack-overflow post](https://stackoverflow.com/questions/50147915/jest-transformignorepatterns-not-working)
- **!!** Error explanation in [stack-overflow post](https://stackoverflow.com/questions/49263429/jest-gives-an-error-syntaxerror-unexpected-token-export)

Secondary:
- https://stackoverflow.com/questions/71058618/to-have-some-of-your-node-modules-files-transformed-you-can-specify-a-custom
- https://stackoverflow.com/questions/35756479/does-jest-support-es6-import-export
-
***

Here are the two approaches:

1. Approach 1 (failed) -- Include all of `node_modules`
  First, I tried to add *all* of `node_modules` to the transformation, with
  ``` js
  transformIgnorePatterns = []
  ```
  in `jest.config.js`, signifying that *nothing* will be exluded from transpilation; that is, every file matching with the regex in the `transform` attribute will be transpiled.

  This solved the problem of `react-leaflet` not being transpiled (the ES code was converted to CommonJS, so JEST was able to use it successfully), but introduced another error:
  ```
  TypeError: $ is not a function
    at Object.<anonymous> (/Users/mimischly/Desktop/SDS/SUMMER/courseography/.yarn/unplugged/core-js-npm-3.31.1-7df9ac7f4f/node_modules/core-js/modules/es.global-this.js:8:1)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._compile (/Users/mimischly/Desktop/SDS/SUMMER/courseography/.yarn/cache/pirates-npm-4.0.6-a8ec571a43-46a65fefaf.zip/node_modules/pirates/lib/index.js:117:24)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at require$$0.Module._extensions..js (/Users/mimischly/Desktop/SDS/SUMMER/courseography/.pnp.cjs:23971:33)
    at Object.newLoader (/Users/mimischly/Desktop/SDS/SUMMER/courseography/.yarn/cache/pirates-npm-4.0.6-a8ec571a43
    ...
    ...
  ```
  As it turns out: **Not all modules in `node_modules` should be transpiled (from `ESM` to `CJS`)!**

  My specific problem had to do with the `core-js` package, which I use in my babel config (`babel.config.js`) as
  ``` js
  {
  "presets": [
    [
      "@babel/preset-env",
      {
        "useBuiltIns" : "usage",
        "corejs": { "version": "3.30", "proposals": true }
      }
    ],
    "@babel/preset-react"
  ]
}
  ```
  As per the extremely helpful [issue from the core-js repo](https://github.com/zloirock/core-js/issues/743), `core-js` should not be transpiled.

  (*Interestingly, this problem does not appear when that configuration is file-relative, i.e. with a `.babelrc.json` file, still investigating why*.)

  Clearly, then, we must finetune what from `node_modules` is included for transpilation. This points us to the second approach.

2. Approach 2 (success) -- Exclude everything from `node_modules` *except* `react-leaflet`.

  I want to exclude `node_modules` (since as we saw some modules must not be transpiled), *except* the problematic package, `react-leaflet`, which *does* need to be transpiled (since it is ESM).

  To achieve this, I updated the `transformIgnorePatterns` in `jest.config.js` to:
  ``` js
  transformIgnorePatterns: ["/node_modules/(?!(react-leaflet|@react-leaflet).*)"]
  ```
  The regex `/node_modules/(?!(react-leaflet|@react-leaflet).*)` translated to: "everything in a `node_modules` directory except any files starting with `react-leaflet` or `@react-leaflet`".

  As a result, everything in `node_modules` is excluded from transpilation except for the `react-leaflet`-related packages, which will be transpiled.

  Alternatively, we could also have
  ``` js
  transformIgnorePatterns: [ "/node_modules/core-js/"]
  ```
  , essentially transpiling everything except for the `core-js` module, although this would not work if there exist other modules in `node_modules` that also must not be transpiled.

I owe credit to [Adam Nathaniel Davis](https://github.com/bytebodger) for his [life-saving post](https://dev.to/bytebodger/how-i-fixed-the-unexpected-token-error-in-jest-4o1j).

<details>

</details>

## Type of Change

| 🚨 _Breaking change_ (fix or feature that would cause existing functionality to change) |         |
| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |     X     |
| 📦 _Dependency update_ (change that updates a dependency)                               |     X     |

## Checklist

## Questions and Comments


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### 🚨 Breaking changes
 
+- Upgraded to React v18
+
+- Upgraded to React v18
```
david-yz-liu, line 9: this is repeated

js/components/draw/main.js
```diff
@@ -1,10 +1,9 @@
 import React from "react"
-import ReactDOM from "react-dom"
 import { Graph } from "../graph/Graph"
+import { createRoot } from 'react-dom/client';
```
david-yz-liu, line 3: this import shouldn't have a semicolon, and should be on line 2 (replacing the old import of `react-dom`)

js/components/grid/grid.js.jsx
```diff
@@ -1,6 +1,5 @@
 import React from "react"
-import ReactDOM from "react-dom"
-
```
david-yz-liu, line 3: keep this blank line (after the `react-dom/client` import)

js/components/post/post.js.jsx
```diff
@@ -1,7 +1,8 @@
 import { SpecialistPost, MajorPost, MinorPost } from "./post_components.js.jsx"
 
 import React from "react"
-import ReactDOM from "react-dom"
+import { createRoot } from "react-dom/client"
+import { createRef } from "react"
```
david-yz-liu, line 5: Don't import this separately; you can do `React.createRef` below due to the previous react import

js/components/post/post_components.js.jsx
```diff
@@ -2,6 +2,7 @@ import React from "react"
 import { CourseCategory2 } from "./course_components.js.jsx"
 import { CourseModal } from "../common/react_modal.js.jsx"
 import * as Data from "./post_data"
+import { createRef } from "react"
```
david-yz-liu, line 5: see my previous comment about this

js/components/search/search.js.jsx
```diff
@@ -1,5 +1,6 @@
 import React from "react"
-import ReactDOM from "react-dom"
+import { createRoot } from "react-dom/client"
+import { createRef } from "react"
```
david-yz-liu, line 3: see previous comment

js/components/search/search.js.jsx
```diff
@@ -137,7 +140,10 @@ class Search extends React.Component {
           </div>
         </div>
         <div id="timetableContainer">
-          <Timetable dept={this.state.curDept} ref="timetable" />
+          <Timetable
+            dept={this.state.curDept}
+            ref=/*"timetable"*/ {this.timetableRef}
```
david-yz-liu, line 145: delete the comment

package.json
```diff
@@ -30,15 +30,16 @@
...
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
+    "react-leaflet": "^4.2.1",
     "react-modal": "^3.16.1"
   },
   "devDependencies": {
     "@babel/core": "^7.24.6",
     "@babel/eslint-parser": "^7.24.1",
     "@babel/plugin-proposal-class-properties": "^7.18.6",
+    "@babel/plugin-transform-modules-commonjs": "^7.24.7",
```
david-yz-liu, line 42: This should already be included by `@babel/preset-env` so doesn't need to be listed separately

jest.config.js
```diff
@@ -192,10 +192,7 @@ const config = {
   },
 
   // An array of regexp pattern strings that are matched against all source file paths, matched files will skip transformation
-  // transformIgnorePatterns: [
-  //   "\\\\node_modules\\\\",
-  //   "\\.pnp\\.[^\\\\]+$"
-  // ],
+  transformIgnorePatterns: ["/node_modules/(?!(react-leaflet|@react-leaflet).*)"],
```
david-yz-liu, line 195: @mimischly7 one more comment here: There should only be one package name, which is either `react-leaflet` or `@react-leaflet`. Don't include both, just include the relevant one.

mimischly7: I appears that the `react-leaflet...` package defines the public API whereas the a`@react-leaflet/core` package (installed automatically when installing the former) provides the concrete implementation of this API. Since both of these modules are in ES, I think we have to transpile both, let me know.

mimischly7: (https://react-leaflet.js.org/docs/core-introduction/)

david-yz-liu (approved):

jest.config.js
david-yz-liu: @mimischly7 Ah that makes sense, thanks 👍

PR #1431, opened 2024-07-19
Upgrade Testing Library (V12 to  V16)

mimischly7:
## Proposed Changes

Having upgraded `react` (and `react-dom`) from v17 to v18, it is necessary to also upgrade the React Testing Library to a version that works with react 18 (and takes advantage of the new features it introduces). Therefore, this PR upgrades RTL from v12 to v16, performing all necessary changes to the tests and dependees (due to the breaking changes introduced with each new major version).

**Notes and Explanation**:

1. The line
  ```js
  await waitFor(() => rtlGraph.queryByText("AAA100") !== null)
  ```
  in `TestGraph.js` is not useful (and it is misleading) due to an incorrect use of RTL's `waitFor`. The callback function passed to `waitFor` should raise an error if the desired state has not been achieved, whereas currently only a *falsey* value is returned (by the callback), which (as per the documentation) is not sufficient to have `waitFor` "retry" the callback. The reason the callback returns a falsey value is due to `queryBy` type queries, which do not raise an error when an element is not found.

  I changed this line to
  ```js
  await waitFor(() => rtlGraph.getByText("AAA100") !== null)
  ```
  , since `getBy` queries raise an error when the element is not found, causing the callback (in `waitFor`) to be retried.

  This fixed most errors, but there was one problematic test file, `js/components/graph/__tests__/Graph.test.js`, where it appeared that the `Graph` was not being rendered correctly before the assertions are made. The problem there was an incorrect jest mock, in particular the line
  ```js
  jest.mock("../Node", () => () => null)
  ```
  . Removing this line fixes the problem. My suspicion is that at some point some of the tests in this file try to render `Node` components to the DOM, but due to the mocking the `vacuous` `() => null` function is called, doing essentially nothing, and the anticipated nodes are not actually present on the DOM.

  This test file also includes the mocks
  ```js
  jest.mock("../Edge", () => () => null)
  jest.mock("../Bool", () => () => null)
  ```
  but all tests pass when these are removed. I am not sure whether we want these there.

  

  2. Incorrect waiting for asynchronous actions to take place. In the `CourseModalButtons.test.js` test file, a helper function does
  ```js
  await waitFor(() => document.querySelector(".ReactModalPortal"))
  ```
  , which is incorrect because the `waitFor` utility only reruns the callback if the callback throws an exception, and the vanilla js `querySelector` method returns `null` if the element is not found (and does not throw an exception). This could be replaced with a `findBy` which waits until the desired element appears (up to a timeout), but this helper appears useless in our case, because *we don't know what text to look for*. Instead, in each test (where we *do* know what text to expect), we will use asynchronous querying (`findBy` or `waitFor` with `getBy`) to wait for the desired element to appear -- replacing the current (incorrect) use of `querySelector`, as in
  ```js
  modalHeader = document.querySelector(".modal-header")
  ```
  , since this (1) does not wait for the element to appear (e.g. this line can run before React has had time to re-render the DOM) leading to errors, and (2) opposes the "Testing Library" ideology of testing as if you were the user.

  Examples like this have been fixed by handling asynchronous behavior using `act` and `findBy`, as in
  `await act(async () => fireEvent.click(genButton))` and
  `await screen.findByText("Invalid Course Input")`.

3. The breaking change of v16 is that @testing-library/react and @testing-library/dom now are separate (whereas before @testing-library/dom was installed automatically by installing @testing-library/react).

## Looking Forward (for another PR)
- [ ] Improve the handling of async rendering behavior (e.g. assertions after a fireEvent call) across the test files
- [ ] `react-test-renderer` used in many test files is depracated (but is used for snapshot testing). Should migrate to rtl instead for all test files
- [ ] Look for unconventional uses of RTL (or usage of deprecated methods) and fix them.
- [ ] Use React **strict mode** to detect vulnerabilities and correct them

<details>

</details>

## Type of Change

| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |    X      |
| 📦 _Dependency update_ (change that updates a dependency)                               |     X     |
| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |    X      |

## Checklist

## Questions and Comments


js/components/common/__tests__/CourseModalButtons.test.js
```diff
@@ -32,142 +25,118 @@ describe("CourseModal", () => {
...
-
     // click on the course link BBB100H1, and BBB100H1's modal opens
     const bbbCourseLink = screen.getByText("BBB100H1")
     fireEvent.click(bbbCourseLink)
 
-    await waitForModalUpdate()
-
     // the forward button is disabled, but the back button is enabled
-    const forwardButton = screen.getByText(">")
+    const forwardButton = await screen.findByText(">")
```
david-yz-liu, line 33: So in all of these tests, I think it's better to call `act` on the `fireEvent` actions above, and leave these queries to be the synchronous version (i.e., `getByText`). This seems more consistent with others uses of `act` that you've added (consistency will also make it easier to replace these with something like [`user-event`](https://testing-library.com/docs/user-event/intro) in the future)

js/components/graph/__tests__/Graph.test.js
```diff
@@ -5,7 +5,6 @@ import { Graph } from "../Graph"
 
 jest.mock("../Edge", () => () => null)
```
david-yz-liu, line 6: I'm fine if you remove all three of these lines

js/components/graph/__tests__/TestGraph.js
```diff
@@ -31,7 +31,7 @@ export default class TestGraph {
     }
 
     const rtlGraph = render(<Graph {...graphProps} />)
-    await waitFor(() => rtlGraph.queryByText("AAA100") !== null)
+    await waitFor(() => rtlGraph.getByText("AAA100") !== null)
```
david-yz-liu, line 34: This is great; you probably found in your [research](https://testing-library.com/docs/dom-testing-library/api-async#findby-queries) that "`findBy` methods are a combination of `getBy` [queries](https://testing-library.com/docs/queries/about#types-of-queries) and [`waitFor`](https://testing-library.com/docs/dom-testing-library/api-async#waitfor)". So here, and the other places you fixed similar issues, it seems like you can just use `await findByText` instead?

david-yz-liu (commented):
@mimischly7 terrific research, and very well documented. I left a few overall comments about using react-testing-library in even more idiomatic ways. 👍

david-yz-liu (approved):

PR #1436, opened 2024-08-04
User `user-event` for simulation user interactions (instead of `fireEvent`)

mimischly7:
## Proposed Changes

# Task 8 - Use `user-event` instead of `fireEvent`
*July 30, 2024*

### Digest

As per the [RTL docs](https://testing-library.com/docs/user-event/intro/),
> `fireEvent` dispatches DOM events, whereas `user-event` simulates full interactions, which may fire multiple events and do additional checks along the way.

Whereas `fireEvent` (a think wrapper around DOM's `dispatchEvent`) allows you to **trigger any event on any element**, the browser usually does much more than triggering an event as a result of a user interaction.

As showed in [this post](https://ph-fritsche.github.io/blog/post/why-userevent), `fireEvent` can lead to very misleading tests, giving the illusion that a certain event could be caused by a user interaction, whereas in reality it would never be possible for a user to cause that event (e.g. due to a `e.preventDefault` in a parent element handler that will prevent the event to propagate to children).

As an example (from the [docs](https://testing-library.com/docs/user-event/intro/)), `user-event`
> factors in that the browser e.g. wouldn't let a user click a hidden element or type in a disabled text box.

### Changes

In summary,
  - `fireEvent.click` was replaced with `user.click` (in some cases followed by `user.unhover`, see note below and code comments)
  - `fireEvent.mouseOver`/`fireEvent.mouseOut` was replaced by `user.hover`/`user.unhover`
  - `fireEvent.mouseDown` was replaced by `user.pointe('{MouseLeft>}')`
  - `fireEvent.keydown` is replaced with the appropriate `user.keyboard(...)` call
  - `fireEvent.change(...)` is replaced with the appropriate a combination of `user.click` and `user.keyboard`, in that order
  - For drag-n-drop:
    Replace
    ``` js
    fireEvent.mouseDown(svg, { clientX: 250, clientY: 10 })
    fireEvent.mouseMove(svg, { clientX: 100, clientY: 20 })
    fireEvent.mouseUp(svg, { clientX: 100, clientY: 20 })
    ```

    with

    ```js
    await user.pointer([
      // touch the screen at element1
      {keys: '[MouseLeft>]', target: svg},
      // move the touch pointer to element2
      {pointerName: 'TouchA', coords: {x: 100, y: 150}},
      // release the touch pointer at the last position (element2)
      {keys: '[/TouchA]'},
    ])
    ```
  - Unfortunately, did not find a way to replace the two `fireEvent.wheel` calls in the codebase. RTL `user-event` does not appear to offer something for this.

### Notes and Observations
- `click`+`hover` problem `user-event`

  `fireEvent.mouseOver` fully worked, but replacing with with user.hover failed -- it did not even dispatch a hover event. Why?

  It turns out it was because I was making a `user.click` call before that, which means that the mouse was already there, so really you cannot cause a `hover` event on an element if the mouse is already ion the element! -- the `fireEvent.mouseOver` artificially does that and that is why it is not preferable (and can give the illusion that an event is possible when the would never be able to fire it). In fact, even
  ```js
  // Add an event listener for each event type
           aaa201.addEventListener("mouseover", (event) => {
             console.log(`Event triggered: ${event.type}`, event);
           });
  ```
  did not trigger.

  So, instead, I had to first and unhover and then it worked.

  ``` js
  it("unselected source and (selected and hovered destination) => edge should transition to inactive 'missing'", async () => {
    const user = userEvent.setup();
    const graph = await TestGraph.build()
    const aaa201 = graph.getByTestId("aaa201")
    const path_101_201 = graph.getByTestId("h101->aaa201")

    await user.click(aaa201)
    expect(path_101_201.classList.contains("inactive")).toBe(true)

    await user.unhover(aaa201); // SOSOSOS
    await user.hover(aaa201)

    // ... assertions ...
  ```

- `click` causes `hover` in `user-event`

  `fireEvent.click` only dispatches a `click` DOM event.

  However, when a user actually clicks on an element, a `mouseOver` even will also be dispatched (since he must take the mouse to the element first). The `user-event` library's `click` function takes this into account, which explains why the following was not working:

  ``` js
  it("selected and hovered over: sets itself and all unmet prereqs to be 'missing'", async () => {
    const user = userEvent.setup();
    const graph = await TestGraph.build()
    const aaa101 = graph.getByTestId("aaa101")
    const aaa201 = graph.getByTestId("aaa201")
    const aaa102 = graph.getByTestId("aaa102")
    const aaa303 = graph.getByTestId("aaa303")

    await user.click(aaa303)

    expect(aaa101.classList.contains("takeable")).toBe(true)
    expect(aaa102.classList.contains("takeable")).toBe(true)
    expect(aaa201.classList.contains("inactive")).toBe(true)
    expect(aaa303.classList.contains("overridden")).toBe(true)

    await user.hover(aaa303)

    expect(aaa101.classList.contains("missing")).toBe(true)
    expect(aaa102.classList.contains("missing")).toBe(true)
    expect(aaa201.classList.contains("missing")).toBe(true)
    expect(aaa303.classList.contains("missing")).toBe(true)
  })
  ```

  The nodes `aaa101` and `aaa102` are `takeable` (because they have no prerequisites) except when you `hover` over a `node` that requires them (`aaa303` in this case), where they turn into `missing`. But `user.click` also dispatches a `hover` event, and hence they turn to `missing`. Therefore, after the clicking, we have to do `user.unhover(aaa303)` to essentially "move the mouse away".
  Adding
  ```js
  await user.unhover(aaa303)
  ```
  after `await user.click(aaa303)` does the job.

<details>

</details>

## Type of Change

| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |     x     |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |     x     |
| 📦 _Dependency update_ (change that updates a dependency)                               |     x     |

## Checklist

## Questions and Comments


package.json
```diff
@@ -64,7 +65,7 @@
     "sass": "^1.63.6",
     "sass-loader": "^13.3.2",
     "style-loader": "^3.3.2",
-    "stylelint": "^16.7.0",
+    "stylelint": "^15.2.0",
```
david-yz-liu, line 68: This change isn't in the scope of this PR, please revert this

david-yz-liu (commented):
@mimischly7 great job with both the changes and the write-up. The use `user.tripleClick` made me chuckle. 😄

js/components/graph/__tests__/Edge.test.js
```diff
@@ -31,244 +32,268 @@ describe("Edge", () => {
...
-        fireEvent.click(aaa303) // selected node with missing prereqs
+        user.click(aaa303) // selected node with missing prereqs
 
         expect(path_101_201.classList.contains("inactive")).toBe(true)
         expect(path_201_and.classList.contains("inactive")).toBe(true)
         expect(path_102_and.classList.contains("inactive")).toBe(true)
         expect(path_and_303.classList.contains("inactive")).toBe(true)
 
-        fireEvent.mouseOver(aaa303)
+        await user.unhover(aaa303)
```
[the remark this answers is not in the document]
mimischly7: Wow...😂

yarn.lock
```diff
@@ -4251,15 +4261,15 @@ __metadata:
   languageName: node
   linkType: hard
 
-"debug@npm:^4.3.5":
-  version: 4.3.5
-  resolution: "debug@npm:4.3.5"
+"debug@npm:^4.3.6":
```
david-yz-liu, line 4264: There are extraneous changes to this file, possibly due to how you changed the dependencies. There should also be a new file for `testing-library/user-event` under the `.yarn` folder, which is missing.

The easiest way to fix this is to do the following: first, revert *all* changes to `yarn.lock`, `package.json`, and `.pnp.cjs`, and `.yarn` (which you didn't commit, which leaves me thinking you may have uncommitted changes in your local repo). In other words, the dependencies should look *exactly* the same as on master, and running `git status` should not show any changes (including untracked changes) to any of them.

Then, run `yarn add -D @testing-library/user-event` and the commit all of the changes to the above files/folders I mentioned.

mimischly7: I see, my bad. I generally avoided to add `.pnp` and the `.yarn` folder because these are automatically generated with `yarn install`. I followed your instructions, are they ok now?

david-yz-liu (approved):

PR #1464, opened 2024-10-26
Optimize about page

ethanliu24:
## Proposed Changes

Each time we go to the about page, a markdown file is compiled to HTML and then served from the backend. This can be inefficient, so this PR aims to improve the load efficiency by letting Webpack to precompile the markdown and serve that rather than compiling everything again.

Also in the about page, when we click on links to learn about privacy and licensing, it is directed to an invalid path. This PR also fixes this issue.
...

<details>

</details>

## Type of Change

| 📦 _Dependency update_ (change that updates a dependency)                               |X         |

## Checklist

## Questions and Comments


ethanliu24:
Hey Professor Liu, in Slack I mentioned that I couldn't get `remark-html` to parse the markdown file properly: it's parsed as a serialized markdown string rather than HTML. This draft is what I have so far with this solution.

In `webpack.common.js`, the commented part is trying to use remark-html and remark-loader (compiled to markdown string, following https://webpack.js.org/loaders/remark-loader/#markdown-to-html), and the code below it is using markdown-loader (compiled to HTML string, the desired output). I have attacked screen shots of what both looks like. 

This is compiled into `./public` and I injected it in the backend. I am also not too sure about how I could retrieve the serialized string from it; I have a few ideas, though I'm not sure if they are good or not:

1. In Haskell, I can read the compiled js script and use regex to retrieve the string.

2. Instead of processing the markdown file directly in Webpack, I could write a new script and using `remark-html` directly (following https://github.com/remarkjs/remark-html?tab=readme-ov-file#use) to compile it, and insert it in the DOM. I don't think this is a good solution though, as we are recompiling the markdown over and over again, just in the front end this time.

3. Almost a combination of the previous two ideas, I could write a new script and during compile time, the serialized string could be written to a new file in `./public` (maybe a txt or html file) and the backend can read and insert it in DOM so we don't have to recompile repeatedly like idea #2?

Screenshot 2024-10-26 at 5 42 40 PM
Screenshot 2024-10-26 at 5 44 09 PM

david-yz-liu (commented):
Hi @ethanliu24, thanks for making this PR and your detailed report. I played around with `remark-html` myself, and replicated the issue you reported (no conversion to HTML), and was unable to figure out how to fix it myself! Strange. So, I'm happy to stick with the alternate library you found, and great research in finding it.

Now, for your question about how to incorporate the HTML: the generated file is a Javascript file that has access to the raw HTML. Right now you just have `README.md` as the entrypoint file, but you can instead create a separate `about.js` file with contents like

```javascript
# about.js
import aboutContents from '../path/to/README.md'

# aboutContents is the HTML string
console.log(aboutContents)
```

and then from there dynamically insert the contents into the webpage. It's not as smooth as if the page were a static HTML file being served, but it should be pretty fast.

ethanliu24:
Hey Professor Liu I've made changes to your comment. The raw HTML is now dynamically inserted into the DOM via scripts in `js/components/about` (I'm not sure where to put these scripts, since they are not react components). Is this the way of inserting HTML that you were looking for?

Also as I was fixing the `privacy` route, I noticed that it can't be triggered from anywhere. When we click on the privacy link in the About page, it directs to route `PRIVACY.md` because that's how it was directed in the markdown file. I think the easiest way to fix that is either by adding a `PRIVACY.md` route or redirect to `privacy` when we are on `PRIVACY.md`. Currently I have the former, but which one do you prefer, or if there's a better way to do it?

Last, I noticed that the `LICENSE` link is also not loaded with the same reasoning as privacy. Should I add a new license route and fix that also?

app/Routes.hs
```diff
@@ -48,8 +47,9 @@ strictRoutes aboutContents privacyContents = [
     ("post", retrievePost),
     ("post-progress", postResponse),
     ("draw", drawResponse),
-    ("about", aboutResponse aboutContents),
-    ("privacy", privacyResponse privacyContents),
+    ("about", aboutResponse),
+    ("PRIVACY.md", privacyResponse),  -- check for this route in strictMatchDir and redirect to /privacy?
```
david-yz-liu, line 51: Okay, I looked into this some more. I don't actually think it's worth generating the privacy policy as an HTML file, and instead I'd like to modify the link to refer directly to https://github.com/Courseography/courseography/blob/master/PRIVACY.md, and similarly for the LICENSE. Please make those changes in the README.md file, and then *remove* all privacy-policy-related routes/functions/files in this PR.

app/Server.hs
```diff
@@ -9,10 +9,9 @@ responses.
 module Server
     (runServer) where
 
-import Config (markdownPath, serverConf)
+import Config (serverConf)
```
david-yz-liu, line 12: This is great; you can check whether `markdownPath` is used anywhere else, and (assuming not), remove it entirely from `Config` and the configuration YAML file.

webpack.common.js
```diff
@@ -1,6 +1,7 @@
 const path = require("path")
 const MiniCssExtractPlugin = require("mini-css-extract-plugin")
 const webpack = require("webpack")
+// const RemarkHTML = import("remark-html")
```
david-yz-liu, line 4: remove this line

david-yz-liu (commented):
@ethanliu24 great work! I left some inline comments, and also I think you forgot to commit the changes in the `.yarn` folder, which is where the actual JS dependencies are stored.

ethanliu24:
It seems like when I do a `yarn install` to install the dependencies I added, quite a few existing dependencies have been deleted. Is this an expected behaviour?

david-yz-liu (commented):
@ethanliu24 it looks like you accidentally upgraded various NPM dependencies, either earlier in the semester or when adding the new dependencies. You can fix this by doing the following:

1. Revert *all* changes to `.yarn`, `package.json`, and `yarn.lock`.
2. Run `yarn install` again, and verify with a `git status` that *nothing is changed* (because at this point your `package.json` should be the same as what's on master).
3. Add the two packages to `package.json`.
4. Run `yarn install` again.

You should then see (with a `git status` that very few packages in `.yarn` have been updated, and that the `yarn.lock` file has just the new dependencies added (plus any of their transitive dependencies).

js/components/about/about.js
```diff
@@ -0,0 +1,3 @@
+import aboutContent from "../../../README.md"
+
+document.querySelector("#aboutDiv").innerHTML = aboutContent
```
david-yz-liu, line 3: Two comments:

1. It's more idiomatic to use `document.getElementById` here
2. Move this code into a [`DOMContentLoaded` event callback](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event)

david-yz-liu (commented):
@ethanliu24 great work! Everything is looking very good, I just left two quick comments.

I agree the javascript file placement isn't ideal, but I think it's fine for now (appealing to consistency with other pages).

david-yz-liu:
@ethanliu24 oh one more thing, please add `public/js/about` to the `.gitignore` (see where the other built files are being listed in that file)

david-yz-liu (approved):

PR #1475, opened 2024-11-02
Parse all transform

elescogna:
## Proposed Changes

This PR adds new parsers for all types of transformations in SVG files, specifically it adds parsers for `matrix`, `rotate`, `scale`, `skewX`, and `skewY`. After parsing the SVG transformations, they are represented as matrices.

#### How I tested my changes:

- for `rotate`, `scale`, `skewX`, and `skewY`, I added some arbitrary transformations to the existing Computer Science graph, and I checked that they were displayed accordingly on the courseography graph page after parsing them
- for `matrix`, I checked that the result remained unvaried when I substituted each of the of the other transformations for their corresponding matrix equivalent.

...

<details>

</details>

## Type of Change

| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |     X   |

## Checklist

- [X] I have updated the project documentation, if applicable.
- [ ] I have updated the project Changelog (this is required for all changes).

- [ ] I have verified that the CircleCI checks have passed.
- [ ] I have [requested a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a project maintainer.

## Questions and Comments

### DIFF
--- a/app/Svg/Generator.hs
+++ b/app/Svg/Generator.hs
@@ -177,6 +177,7 @@ rectToSVG styled courseMap rect
                          _ -> ""
 
         in S.g ! A.id_ (textValue $ sanitizeId $ shapeId_ rect)
+               ! A.transform (stringValue . show . formatTransform $ shapeTransform rect)
                ! A.class_ (textValue class_)
                ! S.customAttribute "data-group" (textValue
                                                  (getArea (shapeId_ rect)))
@@ -206,6 +207,7 @@ rectToSVG styled courseMap rect
 ellipseToSVG :: Bool -> Shape -> S.Svg
 ellipseToSVG styled ellipse =
     S.g ! A.id_ (textValue (shapeId_ ellipse))
+        ! A.transform (stringValue . show . formatTransform $ shapeTransform ellipse)
         ! A.class_ "bool" $ do
             S.ellipse ! A.cx (stringValue . show . fst $ shapePos ellipse)
                       ! A.cy (stringValue . show . snd $ shapePos ellipse)
@@ -227,6 +229,7 @@ textToSVG styled type_ xPos' text =
                       then xPos
                       else xPos')
             ! A.y (stringValue $ show yPos)
+            ! A.transform (stringValue . show . formatTransform $ textTransform text)
             ! (if styled then allStyles else baseStyles)
             $ toMarkup $ textText text
     where
@@ -263,6 +266,7 @@ edgeToSVG styled path =
     S.path ! A.id_ (textValue . T.append "path" . pathId_ $ path)
            ! A.class_ "path"
            ! A.d (textValue . T.cons 'M' . buildPathString . pathPoints $ path)
+           ! A.transform (stringValue . show . formatTransform $ pathTransform path)
            ! A.markerEnd "url(#arrow)"
            ! S.customAttribute "data-source-node" (textValue $ sanitizeId
                                                           $ pathSource path)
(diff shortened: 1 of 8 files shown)


app/Svg/Parser.hs
```diff
@@ -428,10 +428,31 @@ parseTransform "" = [[1, 0, 0],
...
+                  P.try rotate <|> 
+                  P.try translate <|> 
+                  P.try matrix <|> 
+                  P.try skewX <|> 
+                  P.try skewY
+        scale = do
+            _ <- P.string "scale("
+            xScale <- double
+            _ <- P.char ',' <|> P.char ' '
+            yScale <- P.option xScale double
```
elescogna, line 441: Here, as well as in lines 449 and 451 I used `P.option` because of the optional argument, even if they aren't yet supported.

david-yz-liu: Good investigation; if you'd like to handle optional arguments in this PR, instead of using `P.option`, a better approach is to parse all arguments as a list using [`P.sepBy`](https://hackage.haskell.org/package/parsec-3.1.17.0/docs/Text-Parsec.html). This parses a list, and then you can extract the values based on the length of the list.

david-yz-liu (commented):
@elescogna the code looks good. I left one comment about handling optional arguments, since you seem to have made good progress on that front. Make sure to update the changelog as well.

Keep me posted on your debugging the issues you found through manual testing 👍

app/Svg/Generator.hs
```diff
@@ -177,6 +177,7 @@ rectToSVG styled courseMap rect
                          _ -> ""
 
         in S.g ! A.id_ (textValue $ sanitizeId $ shapeId_ rect)
+               ! A.transform (stringValue . show . formatTransform $ shapeTransform rect)
```
elescogna, line 180: This is the line I added to add the transformation to the new SVG file. It made sense to me looking at how the other attributes were added, though it does not appear to be working as it is now.

app/Svg/Generator.hs
```diff
@@ -263,6 +266,7 @@ edgeToSVG styled path =
     S.path ! A.id_ (textValue . T.append "path" . pathId_ $ path)
            ! A.class_ "path"
            ! A.d (textValue . T.cons 'M' . buildPathString . pathPoints $ path)
+           ! A.transform (stringValue . show . formatTransform $ pathTransform path)
```
elescogna, line 269: I am unsure this is the right place to add this line since it is not included in a <g> tag.

app/Svg/Generator.hs
david-yz-liu: Ah sorry I didn't clue into this earlier! This module is used to generate the SVG for the static images (produced when "exporting"), but not the front-end graph rendering. So you should keep these changes, but also I added one commit which adds the changes for the front-end to get the `transform` attribute properly applied.

david-yz-liu (commented):
@elescogna I did some digging here, but it's a tough problem! First, I added one commit to this branch to incorporate some front-end changes to get the `transform` attribute to be properly applied. (So you'll need to pull from this branch to get my latest commit.)

As for the actual intersection issues: I wasn't able to pinpoint the source. There's a problem even for just the raw graph (just translations!). I believe the issue is in how the text and shape intersection code is working (e.g. `buildRect`); when I reverted those changes specifically, the original graph looked better, but still not perfect.

One thing I noticed in `Svg.Parser`: in `parseGraph`, a `globalTransform` is being applied to each point in the associated shape/text/path. This is logically incorrect, and this transform should instead be applied to each entitie's individual `transform` attribute. I'm not sure if this will itself solve the problem, but you should make this change!

david-yz-liu:
Work paused and moved to https://github.com/Courseography/courseography/tree/matrix-transforms.

PR #1502, opened 2025-01-15
Controllers.Course retrieveCourse tests

Jai0212:
## Proposed Changes

Testing is a very important part of any software project.  The goal of this PR is to add tests for all possible cases for the `retrieveCourse` function from `app/Controllers/Course`. Tests are written in `backend-test/Controllers/CourseControllerTests.hs`.

Also added a new file `backend-test/Controllers/ControllerTests.hs` to match the similar format of `backend-test/RequirementTests/RequirementTests.hs` for scalability if tests need to be added for the other Controllers as well.

## Type of Change

| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |    X    |
| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |    X    |

## Checklist

- [X] I have added tests for my changes, if applicable.
- [X] If this is my first contribution, I have added myself to the list of contributors.

## Questions and Comments

Jan 14, 2025:
The addition of `backend-test/Controllers/ControllerTests.hs` was not required but I found it easier to implement it that way and to make the tests similar to the `RequirementTests`. 

Additionally, the current `insertCourse` function only inserts the course name and no other details. If required I can add a new function that adds all other details as well, if it seems necessary for the testing of `retrieveCourse`. 

I created a new mockRequest and runServer for this test in `backend-test/TestHelpers.hs`. For adding future tests, this might not be optimal (because we would not ideally want as many mockRequests  and runServers as the number of tests). There can be a dynamic way of implementing this by passing appropriate parameters to the appropriate mock functions.

I also had to slightly change the original `retrieveCourse` function because without that I could not find a way to extract the course name from the mockRequest.


app/Controllers/Course.hs
```diff
@@ -3,21 +3,31 @@ module Controllers.Course
...
+        (_, Input (Right value) _ _) <- find (\(key, _) -> key == "name") inputs
+        return (BSL.unpack value)
 
 -- | Takes a course code (e.g. \"CSC108H1\") and sends a JSON representation
 -- of the course as a response.
 retrieveCourse :: ServerPart Response
 retrieveCourse = do
-    name <- lookText' "name"
-    courseJSON <- liftIO $ CourseHelpers.queryCourse name
+    req <- askRq
```
david-yz-liu, line 28: Overall, you shouldn't make changes to this file, as you're only adding tests. Revert the changes in this file, and if you're still encountering failing tests, I'll take a closer look and give you some pointers.

app/Controllers/Course.hs
```diff
@@ -43,4 +53,4 @@ depts = do
         return $ sort . nub $ map g coursesList :: SqlPersistM [String]
     return $ createJSONResponse deptList
     where
-        g = take 3 . T.unpack . coursesCode . entityVal
+        g = take 3 . T.unpack . coursesCode . entityVal
```
david-yz-liu, line 56: This symbol on GitHub means you have not ended the file with a new line. You should make sure all files you added/changed end with a new line, and configure your text editor to ensure that all files end with a new line, to make sure this doesn't happen.

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -44,12 +53,25 @@ runIndexTest label courses expected =
...
+runRetrieveCourseTests :: [Test]
+runRetrieveCourseTests = map (\(label, courses, name, expected) -> runRetrieveCourseTest label courses name expected) retrieveCourseTestCases
+
 -- | Test suite for Course Controller Module
-courseControllerTestSuite :: IO Test
-courseControllerTestSuite = do
-    let tests = TestLabel "Course Controller tests" $ TestList runIndexTests
-    return tests
+courseControllerTestSuite :: Test
+courseControllerTestSuite = TestLabel "Course Controller tests" $ TestList (runIndexTests ++ runRetrieveCourseTests)
```
david-yz-liu, line 77: instead of using `(++)`, use `concat:: [[a]] -> [a]`, as this will be more extensible in the future if we add additional test cases

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -33,6 +34,14 @@ indexTestCases =
        "CSC101\nCSC102\nCSC103\nCSC104\nCSC105\n")
     ]
 
+-- | List of test cases as (input courses, course name, expected JSON output)
```
david-yz-liu, line 37: In terms of code organization, keep each controller function's test code separate. That is, make sure everything to do with the `index` function is together, and everything to do with the `retrieveCourse` function is also together.

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -44,12 +53,25 @@ runIndexTest label courses expected =
         let actual = BL.unpack $ rsBody response
         assertEqual ("Unexpected response body for " ++ label) expected actual
 
+-- | Run a test case (case, input, expected output) on the retrieveCourse function.
+runRetrieveCourseTest :: String -> [T.Text] -> String -> String -> Test
+runRetrieveCourseTest label courses name expected =
+    TestLabel label $ TestCase $ do
+        runDb $ do
+            clearDatabase
+            insertCourses courses
```
david-yz-liu, line 62: It doesn't really make sense to use the `insertCourses` helper here. Instead, I would recommend just inserting a course manually (adapting code from the body of `insertCourses`), and pass in authentic values for the `Course` fields rather than `null`

backend-test/TestHelpers.hs
```diff
@@ -40,6 +41,44 @@ runServerPart sp = do
...
+mockRequestWithQuery :: String -> IO Request
+mockRequestWithQuery courseName = do
+    inputsBody <- newMVar []
+    requestBody <- newEmptyMVar
+    return Request
+        { rqSecure          = False
+        , rqMethod          = GET
+        , rqPaths           = ["course"]
+        , rqUri             = "/course"
+        , rqQuery           = "name=" ++ courseName
```
david-yz-liu, line 54: It looks here like you're trying to pass the `name` key-value pair in two different ways, likely the result of your experimentation. Building on my comment above, you should tidy up your work by reverting changes to the controller code, and then verifying whether just one of these approaches work, without requiring the other. If you are stuck, you can then push what you've tried and then I'll take a closer look.

david-yz-liu (commented):
@Jai0212 nice work! I left a few inline comments for you to work on 👍

backend-test/Controllers/CourseControllerTests.hs
Jai0212: Hello Professor, I made this change along with the other mentioned changes but while committing  the pre-commit hook gave this prettier error.

Screenshot 2025-01-16 at 6 58 47 PM

Please advice what I should do in this case.

backend-test/Controllers/CourseControllerTests.hs
Jai0212: Hello Professor, just wanted to check if I can use the function `insertCourse` in `Database.CourseInsertion` for this because details like distribution and breadth require a specific type which this function handles and I am unable to figure out a way to add them using `insert_` directly.

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -6,17 +6,88 @@ Module that contains the tests for the functions in the Course Controller module
...
+                    , coursesDistribution = Nothing
+                    , coursesPrereqString = fmap T.pack $ Map.lookup "prereqString" courseData <|> Nothing
+                    , coursesCoreqs = fmap T.pack $ Map.lookup "coreqs" courseData <|> Nothing
+                    , coursesVideoUrls = maybe [] (\urls -> [T.pack urls]) (Map.lookup "videoUrls" courseData)
+                    }
+                ]
+
+        runDb $ do
+            clearDatabase
+            mapM_ (unless (T.null currCourseName) . insert_) courseToInsert
```
david-yz-liu, line 82: You don't need a `mapM_` here, as you're just trying to insert one course. Make the type of `courseToInsert` be just `Courses` rather than `[Courses]`, and call `insert_ courseToInsert`.

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -6,17 +6,88 @@ Module that contains the tests for the functions in the Course Controller module
...
 import qualified Data.Text as T
 import Database.Persist.Sqlite (SqlPersistM, insert_)
 import Database.Tables (Courses(..))
 import Happstack.Server (rsBody)
 import Test.HUnit (Test(..), assertEqual)
-import TestHelpers (clearDatabase, runServerPart)
+import TestHelpers (clearDatabase, runServerPart, runServerPartWithQuery)
+
+-- | List of test cases as (input course name, course data, expected JSON output)
+retrieveCourseTestCases :: [(String, T.Text, Map.Map String String, String)]
```
david-yz-liu, line 27: Make the type of this map `Map.Map T.Text T.Text`. If we use the [`OverloadedStrings` compiler extension](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/overloaded_strings.html) you can keep the string literals as written, and the compiler will automatically interpret these as `T.Text` values rather than `String` values. I believe the `courseography.cabal` file has enabled `OverloadedStrings` globally, but may not have done so for the tests, in which case you should add it to the configuration in that file.

backend-test/Controllers/CourseControllerTests.hs
```diff
@@ -6,17 +6,88 @@ Module that contains the tests for the functions in the Course Controller module
...
+-- | Run a test case (case, input, expected output) on the retrieveCourse function.
+runRetrieveCourseTest :: String -> T.Text -> Map.Map String String -> String -> Test
+runRetrieveCourseTest label courseName courseData expected =
+    TestLabel label $ TestCase $ do
+        let currCourseName = T.pack $ fromMaybe "" $ Map.lookup "name" courseData
+
+        let courseToInsert = 
+                [ Courses
+                    { coursesCode = currCourseName
+                    , coursesTitle = fmap T.pack $ Map.lookup "title" courseData <|> Nothing
```
david-yz-liu, line 68: The `<|> Nothing` isn't necessary, `Map.lookup` will already return `Nothing` if the key isn't found

backend-test/TestHelpers.hs
david-yz-liu: I don't think you've resolved this comment

PR #1516, opened 2025-02-20
Added front-end validations for GenerateForm

Jai0212:
## Proposed Changes

Instead of validations in the `handleSubmit` function, used Formik `validateForm` to add validations for `GenerateForm`. Removed the existing modal and replaced it with text based specific error messages for each field. 

These are the implemented checks:
- No course entered
- Invalid course/department (based on regex)
- Course/department not separated by commas

image

## Type of Change

| ✨ _New feature_ (non-breaking change that adds functionality)                          |    X    |
| 🎨 _User interface change_ (change to user interface; provide screenshots)              |    X    |
| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |         |
| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |         |

## Checklist

- [X] I have updated the project documentation, if applicable.

## Questions and Comments

23rd Feb, 2025
1) The value of the `errors` variable in `validateForm` isn't really being used anywhere apart from the fact that it has something that indicates that the form should not be allowed to submit. With the way the current `computeMessage` and `ErrorMessage` are implemented, I am not too sure how the value of `errors` can be made of use. Also currently the `errors` is early return so it only stores the first error found.
2) I also added a new title for the error message. It was only "Invalid Course Input" before, now there is an "Invalid Department Input" also

28th Feb, 2025
1) I deleted the old CSS for the error modal as well
2) Added a few more tests for different cases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
 - ⁠Fixed issue with "Info" popup disappearing when hovering over it and reduced the delay before it disappears
 - ⁠Fixed issue where scrolling in the course modal triggered graph zoom
 - Added dependency `formik` and replaced standard React forms with Formik in `GenerateForm`
+- Added front-end validations for Generate form
 
 ## [0.6.0] - 2024-06-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### ✨ New features/enhancements
 
 - Added text warning when there's no timetable information available to display
+- Added text warnings and validations for each field in `GenerateForm`
 
 ### 🐛 Bug fixes
 
@@ -44,7 +45,6 @@
 - ⁠Fixed issue with "Info" popup disappearing when hovering over it and reduced the delay before it disappears
 - ⁠Fixed issue where scrolling in the course modal triggered graph zoom
 - Added dependency `formik` and replaced standard React forms with Formik in `GenerateForm`
-- Added front-end validations for Generate form
 
 ## [0.6.0] - 2024-06-24
 
--- a/js/components/generate/GenerateForm.js
+++ b/js/components/generate/GenerateForm.js
@@ -11,7 +11,7 @@ export default class GenerateForm extends React.Component {
       fceCount: 0,
       showWarning: false,
       invalidCourses: [],
-      errorTitle: "Error",
+      errorTitle: "",
     }
 
     this.graph = React.createRef()
@@ -309,10 +309,12 @@ export default class GenerateForm extends React.Component {
    * @returns {string} The warning message string.
    */
   computeMessage(invalidCourses) {
-    if (invalidCourses.length === 1) {
+    if (this.state.errorTitle) {
       return invalidCourses[0]
+    } else if (invalidCourses.length === 1) {
+      return `The course ${invalidCourses} was invalid! Please check your input.`
     } else {
-      return `The inputs [${invalidCourses}] were invalid! Please check your input.`
+      return `The courses [${invalidCourses}] were invalid! Please check your input.`
     }
   }
 
@@ -320,7 +322,7 @@ export default class GenerateForm extends React.Component {
     return (
       <div style={{ display: "flex", flexDirection: "row", height: "100%" }}>
         <ErrorMessage
-          title={this.state.errorTitle}
+          title={this.state.errorTitle || "Invalid Course Input"}
           message={this.computeMessage(
             this.state.invalidCourses.filter(str => !!/\S/.test(str))
           )}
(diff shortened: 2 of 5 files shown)


david-yz-liu (commented):
@Jai0212 nice work! This is looking good but I think we can go even further with our use of Formik.

Currently we display our own error message modal, but I'd like to instead use Formik's [`ErrorMessage` component](https://formik.org/docs/api/errormessage). This is field-specific (so you'll need to add one to each field) and uses whatever messages are stored in the `errors` variable.

This should allow you to actually get rid of the existing `showWarning`, `errorTitle`, and `invalidCourses` state, and instead just rely on the setting `errors` messages to be displayed. Also note that errors can be displayed in separate `ErrorMessage` components for the different fields, and so in the validation you should accumulate all errors, not stopping at the first one.

Jai0212:
> @Jai0212 nice work! This is looking good but I think we can go even further with our use of Formik.
> 
> Currently we display our own error message modal, but I'd like to instead use Formik's [`ErrorMessage` component](https://formik.org/docs/api/errormessage). This is field-specific (so you'll need to add one to each field) and uses whatever messages are stored in the `errors` variable.
> 
> This should allow you to actually get rid of the existing `showWarning`, `errorTitle`, and `invalidCourses` state, and instead just rely on the setting `errors` messages to be displayed. Also note that errors can be displayed in separate `ErrorMessage` components for the different fields, and so in the validation you should accumulate all errors, not stopping at the first one.

Hi, if we use the `ErrorMessage` component from Formik with each `Field`, it would look something like this with all the errors. So is this fine or should the error message be a pop-up (like it was before)? In the latter case, a modal would be needed.

image

david-yz-liu:
@Jai0212 oh yes I expected that, and yes that's what I'd like to see (I don't think the modal is necessary, and from a UI perspective I think it's better to see the error messages right next to the inputs to give the user an easier time in fixing them.)

That said I'd approve of you tweaking the styling so that the error messages appear more close to the input that they are referring to (right now there seems to be a margin-bottom or something that's causing a separation).

Finally, I forgot to mention that this might be the only place where our custom `ErrorMessage` component is used; if that's the case, you can remove it entirely.

js/components/generate/GenerateForm.js
```diff
@@ -1,16 +1,21 @@
 import React from "react"
-import { Field, Form, Formik } from "formik"
+import { ErrorMessage, Field, Form, Formik } from "formik"
 import { Graph, populateHybridRelatives } from "../graph/Graph"
 import Disclaimer from "../common/Disclaimer"
-import { ErrorMessage } from "../common/react_modal.js.jsx"
+
+const errorMessageStyle = {
```
david-yz-liu, line 6: Add this to the CSS file (targeting an error message class; if Formik doesn't provide one by default, you can add your own)

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
 - ⁠Fixed issue with "Info" popup disappearing when hovering over it and reduced the delay before it disappears
 - ⁠Fixed issue where scrolling in the course modal triggered graph zoom
 - Added dependency `formik` and replaced standard React forms with Formik in `GenerateForm`
+- Added front-end validations for Generate form
```
david-yz-liu, line 47: This is a new feature (not purely internal change)

js/components/generate/GenerateForm.js
```diff
@@ -230,35 +235,56 @@ export default class GenerateForm extends React.Component {
...
     } else {
-      return `The courses [${invalidCourses}] were invalid! Please check your input.`
+      const courses = values.courses.split(",").map(course => course.trim())
+      const invalidCourses = courses.filter(course => !coursePattern.test(course))
+
+      if (invalidCourses.length > 0) {
+        errors.courses =
+          invalidCourses.length === 1
+            ? `The course ${invalidCourses} was invalid! Please check your input.`
+            : `The courses [${invalidCourses}] were invalid! Please check your input.`
```
david-yz-liu, line 254: Please use `Array.prototype.join` here, so that there's a space after each comma. Same for all of the error messages involving lists.

david-yz-liu (commented):
@Jai0212 nice work! I left a few minor comments; please also add a final screenshot to your PR description to show the validation errors.

js/components/generate/GenerateForm.js
```diff
@@ -280,84 +298,92 @@ export default class GenerateForm extends React.Component {
...
-                <option value="utsg">St. George</option>
-                <option value="utm">Mississauga</option>
-                <option value="utsc">Scarborough</option>
-              </Field>
-
-              <p>
-                <label htmlFor="includeRaws">Include non-course prerequisites</label>
-                <Field id="includeRaws" name="includeRaws" type="checkbox"
-                  style={{ marginLeft: '1em', verticalAlign: 'middle' }}
+                <ErrorMessage id="error-message" name="courses" component="div" />
```
david-yz-liu, line 314: In general, HTML `id`s should not be duplicated. You should instead use classes to apply common styling. In JSX syntax you need to use `className` to set the class.

david-yz-liu (approved):
Nice work, @Jai0212!

PR #1517, opened 2025-02-25
Test intersection checking logic

ethanliu24:
## Proposed Changes

This PR adds test cases and fix any errors to ensure that the intersection checking logics supports different types of transformations, that is, translation, scaling, rotation and skewing.

...

<details>

</details>

## Type of Change

| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  |X         |
| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            |X         |

## Checklist

- [ ] I have performed a self-review of my changes.
- [ ] I have updated the project Changelog (this is required for all changes).

- [ ] I have verified that the CircleCI checks have passed.
- [ ] I have [requested a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a project maintainer.

## Questions and Comments

### DIFF
--- a/app/Svg/Builder.hs
+++ b/app/Svg/Builder.hs
@@ -64,13 +64,20 @@ buildRect :: [Text]  -- ^ A list of shapes that may intersect with the given nod
           -> Integer -- ^ An integer to uniquely identify the shape
           -> Shape
 buildRect texts entity elementId =
-    let rectTexts = filter
+    let shapeInverseTransformation = invertMatrix3x3 $ listToMatrix $ shapeTransform entity
+        rectTexts = filter
                 (\text -> intersects
                     (shapeWidth entity)
                     (shapeHeight entity)
-                    (matrixPointMultiply (listToMatrix $ shapeTransform entity) (shapePos entity))
+                    (matrixPointMultiply
+                        shapeInverseTransformation $
+                        matrixPointMultiply (listToMatrix $ shapeTransform entity) (shapePos entity)
+                    )
                     0  -- no tolerance for text intersection
-                    (matrixPointMultiply (listToMatrix $ textTransform text) (textPos text))
+                    (matrixPointMultiply
+                        shapeInverseTransformation $
+                        matrixPointMultiply (listToMatrix $ textTransform text) (textPos text)
+                    )
                 ) texts
         textString = T.concat $ map textText rectTexts
         id_ = case shapeType_ entity of
@@ -92,16 +99,22 @@ buildEllipses :: [Text]  -- ^ A list of Text elements that may or may not inters
               -> Integer -- ^ A number to use in the ID of the ellipse.
               -> Shape
 buildEllipses texts entity elementId =
-    let ellipseText = filter
+    let shapeInverseTransformation = invertMatrix3x3 $ listToMatrix $ shapeTransform entity
+        ellipseText = filter
                     (\text -> intersectsEllipse
                         (shapeWidth entity / 2)
                         (shapeHeight entity / 2)
                         (matrixPointMultiply
-                            (listToMatrix $ shapeTransform entity)
-                            (fst (shapePos entity) - shapeWidth entity / 2,
-                            snd (shapePos entity) - shapeHeight entity / 2)
+                            shapeInverseTransformation $
+                            matrixPointMultiply
+                                (listToMatrix $ shapeTransform entity)
+                                (fst (shapePos entity) - shapeWidth entity / 2,
+                                snd (shapePos entity) - shapeHeight entity / 2)
                         )
-                        (matrixPointMultiply (listToMatrix $ textTransform text) (textPos text))
+                        (matrixPointMultiply
+                            shapeInverseTransformation $
+                            matrixPointMultiply (listToMatrix $ textTransform text) (textPos text)
+                        )
                     ) texts
     in
         entity {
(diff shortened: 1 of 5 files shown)


ethanliu24:
Hey @david-yz-liu could I get some feedback on my test cases to see if I'm on the right track? So far the test cases for `buildRect` and `buildEllipses` are pretty much completed, and I'm imagining the other functions would be similar to what I have now.

david-yz-liu (commented):
@ethanliu24 exciting work, this is great progress!

A few notes:

1. Overall yes the tests are certainly what I'm looking for.
2. When setting up the individual test objects, there's currently a lot of code duplication. One way to simplify this is to create a default object (e.g., `defaultText`) and then use [record field update syntax](https://en.wikibooks.org/wiki/Haskell/More_on_datatypes) to just modify the fields you care about.
3. You've chosen test cases well, but right now the comments contain really valuable information. Instead of making them comments, you should add them to the test case labels (e.g., an extra bit of information that's concatenated to the overall test label in `runBuildRectTests`, `runBuildEllipsesTests`, etc.)

courseography.cabal
```diff
@@ -78,6 +78,8 @@ library
 test-suite Tests
   main-is: Main.hs
   other-modules:
+    SvgTests.SvgTests,
```
david-yz-liu, line 81: Please preserve alphabetical order of this list

app/Svg/Builder.hs
```diff
@@ -192,7 +209,7 @@ shapeTolerance s =
 -- Invert a 3x3 matrix. Assumes that the matrix is invertible
 invertMatrix3x3 :: Matrix -> Matrix
 invertMatrix3x3 m =
-    map (map (* (1 / determinantMatrix3x3 m))) (transposeMatrix3x3 (cofactorMatrix3x3 m))
+    transposeMatrix3x3 $ map (map (* (1 / determinantMatrix3x3 m))) (transposeMatrix3x3 (cofactorMatrix3x3 m))
```
david-yz-liu, line 212: You've correctly identified the bug (fantastic!) but in terms of fixing it, there is an underlying issue.

First, since the main operation on this line is multiplying a matrix by a scalar, the two transposes cancel each other out, and so it is more elegant to remove the existing `transmoseMatrix3x3` call rather than add another one.

Second, the underlying issue is that `cofactorMatrix3x3` isn't actually returning the cofactor matrix, but rather the [adjoint matrix](https://en.wikipedia.org/wiki/Adjugate_matrix#3_%C3%97_3_generic_matrix) (which is the transpose of the cofactor matrix, which explains the "extra" transpose in the existing code). This is the underlying error in the formula. Please rename the function to `adjointMatrix3x3`.

app/Svg/Builder.hs
```diff
@@ -216,5 +233,5 @@ transposeMatrix3x3 _ = error "Matrix must be 3x3"
 
 -- Parse transform back from the format stored in the database
 listToMatrix :: [Double] -> Matrix
-listToMatrix [a, b, c, d, e, f] = [[a, c, e], [b, d, f], [0, 0, 1]]
+listToMatrix [a, b, c, d, e, f] = [[a, b, e], [c, d, f], [0, 0, 1]]
```
david-yz-liu, line 236: On the other hand, this change seems suspicious. The SVG `transform` `matrix` version stores the six numbers in column-major order (https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#matrix) so assuming these six numbers in the database are in the original order of the `matrix` attribute, the original code is correct.

app/Svg/Builder.hs
```diff
@@ -64,13 +68,20 @@ buildRect :: [Text]  -- ^ A list of shapes that may intersect with the given nod
...
           -> Shape
 buildRect texts entity elementId =
-    let rectTexts = filter
+    let shapeInverseTransformation = invertMatrix3x3 $ listToMatrix $ shapeTransform entity
+        rectTexts = filter
                 (\text -> intersects
                     (shapeWidth entity)
                     (shapeHeight entity)
-                    (matrixPointMultiply (listToMatrix $ shapeTransform entity) (shapePos entity))
+                    (matrixPointMultiply
```
david-yz-liu, line 76: Okay, so given this code is working correctly (i.e. passing all test cases), since mathematically we're multiplying a matrix and its inverse, you should be able to remove both multiplications and just pass in `shapePos entity` and preserve the same behaviour.

david-yz-liu (commented):
@ethanliu24 great debugging work, you helped me identify the mathematical error in the previous code. I left some comments to help refine the changes further.

app/Svg/Builder.hs
ethanliu24: I noticed that there may be some floating point/rounding errors. For example, say a text and a shape is both at (50, 100) and they have the same matrix. When we apply the matrix and the inverse to the text only, the resulting coordinate could be something like (50, 99.9999997) while the shape is still at (50, 100). Since there's no tolerance for text checking, the function would detect no intersections. Would this case be considered to not have an intersection?

david-yz-liu: Good question, but that's right we shouldn't consider that a point of intersection. In general, we shouldn't rely on floating point errors for the intersection checking behaviour.

david-yz-liu (approved):
@ethanliu24 great work! 🎉

PR #1616, opened 2025-10-17
Initialize Schema Version Table

akarki2005:
## Proposed Changes

This PR introduces a schema version table within `Tables.hs`, and updates the `setupDatabase` function in `Database.hs` to initialize the schema version table on database creation.

## Type of Change

| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |      X    |

## Checklist

- [N/A] I have added tests for my changes, if applicable.
- [N/A] I have updated the project documentation, if applicable.
- [N/A] If this is my first contribution, I have added myself to the list of contributors.

- [X] I have verified that the CircleCI checks have passed.
- [X] I have [requested a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a project maintainer.

### DIFF
--- a/app/Database/Database.hs
+++ b/app/Database/Database.hs
@@ -37,7 +37,7 @@ setupDatabase quiet = do
         db = T.unpack $ T.take ind dbPath
     createDirectoryIfMissing True db
 
-    -- Ensure SQL schema matches ORM, and initialize schema version table
+    -- Match SQL database with ORM, then initialize schema version table
     let migrateFunction = if quiet then void . runMigrationQuiet else runMigration
     runDb $ void $ migrateFunction migrateAll >> getDatabaseVersion
 


app/Database/Database.hs
```diff
@@ -35,12 +36,29 @@ setupDatabase quiet = do
...
+
+    -- 1. Ensure SQL schema matches ORM (in Tables.hs)
     runDb (
         if quiet
             then void $ runMigrationQuiet migrateAll
             else runMigration migrateAll
         )
 
+    -- 2. Initialize the SchemaVersion table.
+    void $ runDb getDatabaseVersion
```
david-yz-liu: Please use the same logic as above (only call `void` if `quiet`). You can first define the action `runMigrationQuiet migrateAll >> runDb getDatabaseVersion` and then conditionally call `void` on it or just invoke it directly

app/Database/Database.hs
```diff
@@ -35,12 +36,29 @@ setupDatabase quiet = do
...
+-- | Gets the current version of the database.
+-- If no version is defined, initialize the
+-- version to 1 and return that.
+getDatabaseVersion :: MonadIO m => SqlPersistT m Int
+getDatabaseVersion = do
+    result <- selectFirst [] []
+    case result of
+        Just entity -> pure $ schemaVersionVersion $ entityVal entity
+        Nothing -> do
+            insert_ $ SchemaVersion 1
```
david-yz-liu: let's define a variable for `1` so that there's no ambiguity that the `1` plays the same role in both this line and the next.

app/Database/Database.hs
```diff
@@ -35,12 +36,27 @@ setupDatabase quiet = do
...
         db = T.unpack $ T.take ind dbPath
     createDirectoryIfMissing True db
+
+    -- Ensure SQL schema matches ORM and initialize schema version table
     runDb (
         if quiet
-            then void $ runMigrationQuiet migrateAll
-            else runMigration migrateAll
+            then void (runMigrationQuiet migrateAll >> getDatabaseVersion)
+            else runMigration migrateAll >> getDatabaseVersion >> return ()
```
akarki2005: Hi David, just wanted to get a second set of eyes on this logic here; I had to add in the `return ()` so that the return type would match the `setupDatabase` signature, but my intention is for this to be temporary and be replaced in my next PR with a `else do` block that does `runMigration migrateAll`, `version <- getDatabaseVersion` and `applyMigrations version migrations` sequentially. 

I would appreciate any guidance if I am on the wrong track with this; thanks in advance.

app/Database/Database.hs
akarki2005: Hi David, I have changed this section of the function to incorporate your feedback. However, I'm a bit unsure as to whether or not my solution was the intended one. 

Currently, in the else branch, I execute `runMigration migrateAll >> getDatabaseVersion >> return ()`. I initially did not have the `return ()`, but got a compiler error due to a type mismatch between the return value of `getDatabaseVersion` and the `setupDatabase` function. With the `return ()` snippet in place, the type mismatch is resolved, however, as per my understanding of how actions work, this simply discards the `getDatabaseVersion` return value; nearly equivalent to just voiding the return value in the first place. 

I would appreciate any feedback you may have regarding my solution; thanks in advance!

app/Database/Database.hs
```diff
@@ -35,11 +36,24 @@ setupDatabase quiet = do
...
-    runDb (
-        if quiet
-            then void $ runMigrationQuiet migrateAll
-            else runMigration migrateAll
-        )
+
+    -- Ensure SQL schema matches ORM, and initialize schema version table
+    runDb (if quiet
+            then void $ runMigrationQuiet migrateAll >> getDatabaseVersion
+            else runMigration migrateAll >> getDatabaseVersion >> return ())
```
david-yz-liu: @akarki2005 So you can use the [`void`](https://hackage.haskell.org/package/base-4.21.0.0/docs/Control-Monad.html#v:void) function to throw away the return value of a monadic action (instead of `>> return ()`). But, in my previous comment I missed that the key difference between the branches is the use of `runMigrationQuiet` vs. `runMigration`.

So you can simplify the code further by defining the migration action separately:

```
let migrateAction = if quiet then void . runMigrationQuiet else runMigration
runDb $ void $ ...
```

This is a more functional way about thinking of this problem, where you use a conditional to select a _function_ rather than to actually control program flow.

akarki2005: Makes sense, thanks a lot for the feedback David! I'll be sure to point this out at next week's meeting.

akarki2005:
Hi David, I've made the requested changes, but my CI checks are now failing in the `Install Package Dependencies` step with this error message:

```
--2025-10-19 18:48:58--  https://imagemagick.org/archive/binaries/magick
Resolving imagemagick.org (imagemagick.org)... 50.251.58.13
Connecting to imagemagick.org (imagemagick.org)|50.251.58.13|:443... connected.
Unable to establish SSL connection.
```

From this, I've deduced the corresponding step in our CI pipeline is `sudo wget https://imagemagick.org/archive/binaries/magick`. Based on the way other package dependencies are installed in that step, I think we can modify this line to `sudo apt-get install -y imagemagick` to avoid connection issues.

PR #1659, opened 2026-01-30
Implement 404 status code to retrieveCourse

a1-su:
## Proposed Changes

This PR adds a 404 status code return to `retrieveCourse` function in `Controllers/Courses.hs` when the course is not found (in addition to the corresponding 200 status code return when it is). 

It additionally modifies the tests to include the added status codes, and updates the front-end functions that use the route to the change.

## Type of Change

| 🚨 _Breaking change_ (fix or feature that would cause existing functionality to change) | X         |
| 🔧 _Internal_ (change that _only_ affects developers or continuous integration)         |X          |

## Checklist

## Questions and Comments

1. When initially trying to commit, I ran into the issue of an unused variable error, for `TIMEOUT_NAMES_ENUM`, from eslint in the `Graph.js` file on the front-end. When I checked the [original commit](https://github.com/Courseography/courseography/commit/f5e8a3a9464b202b484d34321c807dbc749fb339#diff-d55b268d8dd410dcc45789c61abb69ec32768cd7cb37f0157f8d8dd35e9c24ad) I wasn't sure as to how the change had originally passed eslint.

    https://github.com/Courseography/courseography/blob/7d974d29302ff8fa2d16d2819802d8baba33f600/js/components/graph/Graph.js#L25-L27

    From checking the previous commits on master, the only changes between the commit that added this change and now were from dependabot, which to my knowledge didn't affect the eslint configuration. I unfortunately wasn't able to separate the unused variable removal and 404 status code changes into separate commits since committing my main changes without removing the unused variable meant I ran into the same eslint error, so if this change should have been made in a separate pull request I would be happy to fix it.

2. In addition to modifying the front-end function, `getCourse` in `components/common/utils.js`, I modified the functions that called this utility since from what I gathered they did not handle the not found case correctly, although since this was not in the original description let me know if I should revert it. 

    https://github.com/Courseography/courseography/blob/5fe4f9bbc6b7dc1a11a607994acf88b8f0e650f8/js/components/common/react_modal.js.jsx#L103-L112

    https://github.com/Courseography/courseography/blob/5fe4f9bbc6b7dc1a11a607994acf88b8f0e650f8/js/components/grid/course_panel.js.jsx#L153-L165

    Originally, if the course was not found, it would also return `null`, which ostensibly causes an error should the old functions try to access a key in the response body. It instead now returns a default value and logs the error on the console; I'm not completely sure if this standard for this project, since from what I saw similar functions on the front-end either did not check error conditions or have a need to in the first place. 

3. I've added the status code itself to the tests for the `retrieveCourse` function, although no other test cases in the application follow a similar pattern, either because they do not return a response with a status code, or because the only status code that they return is 200 so it would be redundant to test it. I was uncertain if this was part of the requested changes, so feel free to let me know should it need to be reverted.

4. To my understanding, this change is a breaking change since it changes the API contract for the `retrieveCourse` function used by the `/course` endpoint in the application, but if this is mislabelled I am happy to fix it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### 🚨 Breaking changes
 
-- Added 404 status code return to `retrieveCourse` in `Controllers/Course`
-
 ### ✨ New features/enhancements
+
+- Added 404 status code return to `retrieveCourse` in `Controllers/Course` and added front-end tests for affected components
 
 ### 🐛 Bug fixes
 
--- a/js/components/common/__tests__/CourseModalUpdate.test.js
+++ b/js/components/common/__tests__/CourseModalUpdate.test.js
@@ -1,0 +1,64 @@
+import { render, screen, cleanup } from "@testing-library/react"
+import { CourseModal } from "../react_modal.js.jsx"
+import fetchMock from "fetch-mock"
+
+describe("componentDidUpdate", () => {
+  beforeEach(() => {
+    cleanup()
+    fetchMock.restore()
+  })
+
+  afterEach(() => {
+    fetchMock.restore()
+  })
+
+  it("fetches course data and updates title when a new course is selected", async () => {
+    const mockCourseData = {
+      name: "CSC110Y1",
+      title: "Foundations of Computer Science I",
+      description: null,
+      prereqs: null,
+      exclusions: null,
+      breadth: null,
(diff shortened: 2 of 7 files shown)


david-yz-liu (approved):
Nice work, @a1-su!

PR #1660, opened 2026-01-30
Add tests for getGraphJSON

Purplegaze:
## Proposed Changes

Draft PR to add `getGraphJSON` to the `Controllers.Graph` test suite.

There are currently no test cases other than the empty case. I'll make this PR ready for review once those are added (see Questions section)

...

<details>

</details>

## Type of Change

| 🚦 _Test update_ (change that _only_ adds or modifies tests)                            | X        |

## Checklist

- [X] If this is my first contribution, I have added myself to the list of contributors.

## Questions and Comments

~~This is a draft PR until I populate `getGraphJSONTestCases` beyond the empty case.~~

~~I'm not sure whether I should do this via building the shapes directly (which I'll need to figure out, with `Svg.Builder` I assume), or directly via JSON import like in the `saveGraphJSON` tests. I think the former would be better since getGraphJSON and saveGraphJSON tests sound like they should be separate from each other, but I'd like confirmation on what direction to go here.~~

~~The assertEqual statement is just a string check at the moment, but it would be cleaner to change it to parse the relevant JSON fields directly (with one assert statement for each relevant part of the response body, if I do the former option of building the shapes directly)~~ **Resolved.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     ]
   },
   "dependencies": {
-    "@babel/runtime": "^7.28.4",
+    "@babel/runtime": "^7.28.6",
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.1",
     "@fortawesome/fontawesome-svg-core": "^7.1.0",
@@ -49,10 +49,10 @@
     "react-tooltip": "^5.28.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.28.5",
-    "@babel/eslint-parser": "^7.28.5",
+    "@babel/core": "^7.29.0",
+    "@babel/eslint-parser": "^7.28.6",
     "@babel/plugin-proposal-class-properties": "^7.18.6",
-    "@babel/preset-env": "^7.28.5",
+    "@babel/preset-env": "^7.29.0",
     "@babel/preset-react": "^7.28.5",
     "@eslint/compat": "^1.2.9",
     "@testing-library/dom": "^10.4.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -29,14 +29,32 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/compat-data@npm:^7.27.2, @babel/compat-data@npm:^7.27.7, @babel/compat-data@npm:^7.28.5":
+"@babel/code-frame@npm:^7.28.6, @babel/code-frame@npm:^7.29.0":
+  version: 7.29.0
+  resolution: "@babel/code-frame@npm:7.29.0"
+  dependencies:
+    "@babel/helper-validator-identifier": ^7.28.5
+    js-tokens: ^4.0.0
+    picocolors: ^1.1.1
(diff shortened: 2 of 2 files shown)


david-yz-liu (commented):
Hi @Purplegaze, you're definitely on the right track!

Regarding creating instances of `Text`, `Path`, and `Shape`, you can see the original definitions of these data types in `app/Database/Table.hs`. The syntax is a bit different from the normal type definitions because it's using Persistent's schema syntax, but the principle is the same: each type can be treated as a [record](https://learnyouahaskell.github.io/making-our-own-types-and-typeclasses.html#record-syntax) and can be created just by passing in values of the correct types for each field. The only tricky one is `GraphId` - you should be able to use `toSqlKey 1` as the value for this (don't worry about the exact key value, as it'll be overridden in `insertGraph`).

In terms of the assertion, you can try decoding the response string into an `SvgJSON` instance, like what `saveGraphJSON` does.

backend-test/Controllers/GraphControllerTests.hs
```diff
@@ -102,6 +103,104 @@ runSaveGraphJSONTest label payload =
...
+        ),
+
+        ("Two-shape connected graph",
+            ([testWords1, testWords2], [testShape1, testShape2], [testPath])
+        )
+    ]
+
+-- | Run a test case (case, (texts, shapes, paths)) on getGraphJSON.
+runGetGraphJSONTest :: String -> ([Text], [Shape], [Path]) -> TestTree
+runGetGraphJSONTest label (textlist, shapelist, pathlist) =
```
david-yz-liu: in general, for lists prefer a naming convention which just uses a noun plural form (e.g., `texts` over `textlist`)

david-yz-liu: Good question, you can use `texts'` etc. instead

backend-test/Controllers/GraphControllerTests.hs
```diff
@@ -102,6 +103,104 @@ runSaveGraphJSONTest label payload =
...
+        runDb $ do
+            clearDatabase
+            insertGraph graphName (SvgJSON textlist shapelist pathlist)
+        response <- runServerPartWith Controllers.Graph.getGraphJSON $ mockGetRequest "/get-json-data" [("graphName", T.unpack graphName)] ""
+        let body = rsBody response
+        let jsonObj = decode body :: Maybe SvgJSON
+        case jsonObj of
+            Nothing -> assertFailure ("Maybe SvgJSON returned as Nothing for " ++ label)
+            Just svg -> do
+                assertEqual ("Texts differ for " ++ label) (show (map (\text -> text {textGraph = toSqlKey 1}) textlist)) (show (map (\text -> text {textGraph = toSqlKey 1}) (texts svg)))
```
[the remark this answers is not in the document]
david-yz-liu: I see, in that case it's fine for you to add `deriving Eq` to these tables.

Purplegaze:
Changes applied!

PR #1774, opened 2026-07-26
Refactor hybrid text parsing to account for arbitrarily-nested ANDs and ORs

Purplegaze:
## Proposed Changes

Currently, hybrid text nodes are only equipped to handle a single AND of ORs, which makes the 2nd year calculus hybrid node (`(MAT235,MAT236)/MAT237/MAT257`) not work properly. This PR fixes this bug by making hybrid text parsing recursive and changing the code that updates the state and relatives of hybrid nodes to work separately from regular nodes and account for arbitarily-nested course prerequisite strings.

...

<details>

</details>

## Type of Change

| 🐛 _Bug fix_ (non-breaking change that fixes an issue)                                  | X        |

## Checklist

## Questions and Comments

See below.


Purplegaze:
Hello, here's my draft PR for this issue.

My current solution is to split `updateNode` to handle regular and hybrid nodes separately, by having a new function `arePrereqsSatisfiedHybrid` that determines the state of a hybrid node between `active` and `inactive` with the recursive parsing framework you described. Then, every course contained within a hybrid node is added (flattened into one list) to the parent-child relationships so that states are properly checked when they need to be.

The existing code that updates nodes did not ever set hybrid nodes to anything that isn't `active` or `inactive`, so I believe the `updateNode` logic is unchanged here. However, I noticed `takeable` can be a state that was applied to hybrid nodes when they are first initialized (line 191 of `Graph.js`), and the Stats graph also has takeable hybrid nodes right now, though they appear not to function properly. I'm wondering if I need to update this or pursue a different approach to get the `takeable` state to function in hybrid nodes.

Another approach would be to change the functionality of parent-child relationships as well, and add the full nested layers of nodes to them, since currently parent-child relationships only work on "ANDs of ORs" as well, though that might be an overly complicated thing to implement if the only thing that's necessary is to test whether a hybrid node is active.


Also, with recursive parsing, it becomes more difficult to have parentheses be able to open up in the middle of a course name, such as in the string `MAT(135,136)/137/157` as was in one of the tests, which I think would throw this function in an infinite loop (at least until the maximum recursion depth is hit). It is probably possible by somehow combining recursive logic with the iterative parsing that existed before, but since no hybrid nodes appear to use this level of shorthand in the graphs, I've left it out right now. Basic shorthand that doesn't go in and out of parenthesis layers (like `MAT137/157`) still works. Let me know if you'd like me to still implement the more complex type of shorthand.

david-yz-liu:
Hi @Purplegaze, thanks for the detailed update. Overall I think this is the right approach.

- The split to a new `arePrereqsSatisfiedHybrid` function is good
- It's fine for hybrid nodes to only have active/inactive states, so let's not worry about "takeable" for these nodes.
- No need to worry about parentheses within a course name, e.g. `"MAT(135,136)"`.

js/components/graph/Graph.js
```diff
@@ -945,17 +945,30 @@ export class Graph extends React.Component {
    */
   updateNode = (nodeId, recursive) => {
     let newState
-    if (this.arePrereqsSatisfiedNode(nodeId)) {
-      if (this.isSelected(nodeId) || this.state.hybridsJSON[nodeId]) {
+    // For a hybrid node, set the state to 'active' or 'inactive' depending on if its text is satisfied
```
david-yz-liu: Put this comment inside the if branch (and same with the else branch). This removes the need to change the style of `} else {` below

js/components/graph/Graph.js
```diff
@@ -1278,6 +1291,46 @@ export class Graph extends React.Component {
...
 
+  /**
+   * Checks whether a hybrid node's prereq string is satisfied
+   * @return {boolean}
+   */
+  arePrereqsSatisfiedHybrid = nodeId => {
+    // Concatenate prereq string
+    let hybridNode = this.state.hybridsJSON[nodeId]
+    let hybridText = ""
+    hybridNode.text.forEach(textTag => (hybridText += textTag.text))
```
david-yz-liu: You can use [`Array.prototype.join`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) here

js/components/graph/Graph.js
```diff
@@ -1764,39 +1817,30 @@ export function populateHybridRelatives(hybridNode, nodesJSON, parents, children
   let hybridText = ""
   hybridNode.text.forEach(textTag => (hybridText += textTag.text))
   const nodeParents = []
-  // First search for entire string (see Stats graph)
+  // First search for a node that matches the entire string
```
david-yz-liu: Same comment as above. Also, keep the reference to the Stats graph.

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
+  return splitList
 }
 
 /**
- * Helper function for parsing hybrid node's text
- *
- * @param {string} s a combination of courses
- * @param {string} prefix prefix of the current course
- * @returns {Array} an array containing parsed courses
+ * Helper function to strip a string entirely contained within a pair of parentheses.
```
david-yz-liu, line 85: Include the `@param` and `@return` definitions for this function

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
  */
-export function parseCourse(s, prefix) {
-  "use strict"
+export function removeOuterParens(s) {
+  if (s.length < 2 || s.charAt(0) !== "(" || s.charAt(s.length - 1) !== ")") {
```
david-yz-liu, line 90: Overall I think this implementation is more complex than it needs to be.

You can keep track of start/end pointers and iterate while they match `'('` and `')'`, respectively, and then return when they don't. This will allow this function to naturally generalize to cases with more than one outer set of parentheses like `"((CSC148H1))"`.

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
+  let splitList = []
+  let currIndex = 0
+  let curr = ""
+  let parenLayer = 0 // Depth of nested parentheses
+  while (currIndex < s.length) {
+    // If a parenthesis is encountered, update parenLayer
+    if (s.charAt(currIndex) === "(") {
+      parenLayer += 1
+    }
+    if (s.charAt(currIndex) === ")") {
```
david-yz-liu: use `else if` here and below, as these cases are mutually exclusive

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
+      parenLayer -= 1
+    } 
+
+    // If the separator is encountered and we aren't inside parentheses, split on it and reset curr
```
david-yz-liu: As above, put the comments inside the branches

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
   }
 
-  if (s.length === 3) {
-    return [prefix + s, ""]
+  return s.substr(1, s.length - 2)
+}
+
+/**
+ * Helper function to expand shorthand course codes (e.g. "MAT237/257") in-place from a list of course
+ * strings, and remove any grade requirement strings (e.g. "MAT137 (73%)")
```
david-yz-liu, line 123: Include `@param` and `@return` annotations here as well

js/util/util.js
```diff
@@ -1,99 +1,139 @@
...
+ */
+export function parseSplitList(splitList) {
+  let currPrefix = ""
+  for (let i = 0; i < splitList.length; i++) {
+    if (typeof splitList[i] === "object") {
+      currPrefix = ""
+    }
+    else if (typeof splitList[i] === "string") {
```
david-yz-liu: write this on the previous line

david-yz-liu (commented):
Nice work, @Purplegaze! I just left a few style comments, but the overall approach is great 👍

js/util/util.js
Purplegaze: Wouldn't this break for a string like `(CSC110),(CSC111)`?

david-yz-liu: Ah that's right, great point. You can keep the iteration structure below but should be able to extend it to handle multiple pairs of opening and closing parentheses by precomputing the extent of consecutive `")"` at the end of the string; iterating up to that point can be used to determine how many pairs of outer parentheses there are.

david-yz-liu (approved):
Nice work, @Purplegaze!

PR #1789, opened 2026-08-07
Refactor JSON parsing to replace data types with helper functions

AngelsandDevsLOL:
## Proposed Changes

This PR refactors `app/WebParsing/UtsgJsonParser.hs` to remove the `DB` and `DBList` datatypes in favour of two helper functions, `parseMeetingInfo` and `parseCourse`. In addition, test cases are added for the `insertMeeting` function. 

The `DB` and `DBList` datatypes exist to carry `FromJSON` instances for parsing the timetable API response. They aren't used anywhere else in the codebase and are constructed at a single call site only to be unwrapped again immediately by a helper function. Replacing them with helper functions gives the same behaviour and makes the parsing logic more readable. 

## Type of Change

| ♻️ _Refactoring_ (internal change to codebase, without changing functionality)          |     X    |

## Checklist

- [X] I have added tests for my changes, if applicable.

- [X] I have verified that the CircleCI checks have passed.
- [X] I have [requested a review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a project maintainer.

### DIFF
--- a/app/WebParsing/UtsgJsonParser.hs
+++ b/app/WebParsing/UtsgJsonParser.hs
@@ -1,7 +1,6 @@
-module WebParsing.UtsgJsonParser (parseTimetable, insertAllMeetings) where
+module WebParsing.UtsgJsonParser (parseTimetable, insertAllMeetings, insertCourses) where
 
 import Config (createReqBody, reqHeaders, runDb, timetableApiUrl)
-import Control.Monad (forM_)
 import Control.Monad.IO.Class (liftIO)
 import Data.Aeson (Value, decode, encode, withObject, (.!=), (.:), (.:?))
 import Data.Aeson.Types (Parser, parseMaybe)
@@ -72,8 +71,10 @@ getPageInfo respBody = do
 
 -- Helper function to insert courses for a page of a HTTP response
 insertCourses :: ByteString -> SqlPersistM ()
-insertCourses respBody = do
-    forM_ (parseMeetingInfo respBody) (mapM_ insertMeeting)
+insertCourses respBody =
+    case parseMeetingInfo respBody of
+        Nothing -> liftIO $ print ("Failed to parse meeting information." :: String)
+        Just meetTimes -> mapM_ insertMeeting meetTimes
 
 -- | insert/update all the data into the Meeting and Times schema by creating and sending
 --   the http request to Artsci Timetable and then parsing the JSON response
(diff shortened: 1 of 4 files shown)


app/WebParsing/UtsgJsonParser.hs
```diff
@@ -72,14 +73,7 @@ getPageInfo respBody = do
...
 insertCourses respBody = do
-    let meetings :: Maybe DBList = decode respBody
-    case meetings of
-        Nothing -> return ()
-        Just dblist -> mapM_ insertMeeting $ flattenDBList dblist
-
--- | Helper function to flatten the list of DB Objects
-flattenDBList :: DBList -> [MeetTime]
-flattenDBList (DBList meetings) = concatMap (\(DB meetTimes) -> meetTimes) meetings
+    forM_ (parseMeetingInfo respBody) (mapM_ insertMeeting)
```
david-yz-liu: Since `parseMeetingInfo` returns a `Maybe [MeetTime]`, you can simplify to use just a bit of pattern-matching here and display an error message when the parsing fails. (You also don't need the `do`, which is for multi-step monadic computations.)

```haskell
case parseMeetingInfo respBody of
  Nothing -> -- print an error message
  Just meetTimes -> ...
```

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -136,28 +130,24 @@ meetingUpdates m =
...
-    parseJSON = withObject "Expected an Object for DBList" $ \o -> do
-        maybeOrg <- o .:? "payload"
-        case maybeOrg of
+-- | Parse all of the meetings, along with their associated times
+parseMeetingInfo :: ByteString -> Maybe [MeetTime]
+parseMeetingInfo respBody = do
+    json <- decode respBody
+    flip parseMaybe json $ \obj -> do
+        maybePayload <- obj .:? "payload"
+        case maybePayload of
```
david-yz-liu: I don't think you need a `case` here. Instead use `.:` above, and if it fails then `Nothing` will be returned, which I think is okay.

app/WebParsing/UtsgJsonParser.hs
```diff
@@ -136,28 +130,24 @@ meetingUpdates m =
...
+            Nothing -> return []
             Just payload -> do
                 pageableCourse <- payload .: "pageableCourse"
-                courses <- pageableCourse .: "courses"
-                dbList <- mapM parseJSON courses
-                return $ DBList dbList
-            Nothing -> return $ DBList []
+                courses :: [Value] <- pageableCourse .: "courses"
```
david-yz-liu: Is this type annotation necessary? I would think that GHC can infer the type `[Value]` based on how `courses` is being used below.

david-yz-liu (commented):
@AngelsandDevsLOL nice work! I left a few minor comments, but please also add new tests for the `insertCourses` function (this file currently isn't being tested!).

app/WebParsing/UtsgJsonParser.hs
AngelsandDevsLOL: It is necessary, because `concat` is defined as `concat :: Foldable f => f [a] -> [a]`, where courses should be `[Value]`, but the `[]` is not inferred. The same thing happens with `mapM`, where it takes `mapM :: (Traversable t, Monad m) => (a -> m b) -> t a -> m (t b)`, where `courses` should match `t a` but doesn't know what `t` (in this case `[]`) is. 

If it isn't included, then `courses` gets type `Any Value`, and GHC tells us that `courses` should be `courses :: t0 Value`, but `t0` is ambiguous.

david-yz-liu: Ah okay that makes sense, thanks!

david-yz-liu (approved):
Great, thanks @AngelsandDevsLOL!

